PR #17, opened 2017-07-20
Escape plist values in pbxproj

### LINKED ISSUE #16: Unescaped strings in pbxproj
Content that is written to a pbxproj is invalid if certain properties contains spaces or are empty:

For example
```
name = My cool name;
```
should be escaped to
```
name = "My cool name";
```

yonaskolb:
Fixes #16

### Short description
This makes sure string content in a plist file is escaped by surrounded quotes when writing if it is:
- empty
- contains a space
- contains invalid characters

### Solution
This simply surrounds a given string with quotes given the above rule. 
The first commit fixes the use of spaces and empty strings in places where they weren't manually escaped.
As a follow up, it removes the manual use of .quoted, as they can be applied automatically now, which also removes a lot of un-required quotes

### Implementation
- made sure strings were escaped
- removed manual use of .quoted


yonaskolb:
Hi @pepibumur, just so I know for the future, should I mention you to generate notifications for you, or just leave it, as I noticed you aren't a watcher on this repo

Sources/xcodeproj/PBXProjWriter.swift
```diff
@@ -84,7 +84,11 @@ class PBXProjWriter {
     }
     
     private func write(commentedString: CommentedString) {
-        write(string: commentedString.string)
+        var string = commentedString.string
+        if (string == "" || string.contains(" ") || string.contains("<") || string.contains(">")) && !(string.hasPrefix("\"") && string.hasSuffix("\"")) {
```
pepibumur: 👏 great job @yonaskolb. I wonder if we are missing any case here.

pepibumur:
Hey @yonaskolb. It'd be great if you could mention me or assign me as a reviewer. I've added you as collaborator so I think you should be able to do it. 
Just only a small detail on this PR. Since we don't export the strings commented anymore, there are a few unit tests failing. Could you update them to check that we are not exporting the quoted strings anymore?
🎉 great first contribution!

Sources/xcodeproj/PBXProjWriter.swift
yonaskolb: I added a few more cases in a regex 👍

PR #20, opened 2017-07-22
Make XcodeProj Writable

### LINKED ISSUE #18: Make XcodeProj writable
This would write the workspace, pbxproj, and shared data

yonaskolb:
resolves #18

This makes `XcodeProj` writable

There are also some other changes bundled up here:
### Removal of path property from top level objects. 
Where something came from is not really an important part of its identity. The path where things are written has now been moved to the Writable protocol. This also allows one to change the top level path of something, without having to change all the child objects. (For example when writing XcodeProj to a new path)

### Use PathKit for writing
Changed to use the simple API of PathKit to write files. The FileManager instance used for writing before was hardcoded anyway. There is still a dependancy injected FileManager that is used in the reading of files. This could be replaced with PathKit as well, removing quite a bit of code. What do you think @pepibumur?

### DIFF
--- a/Sources/xcodeproj/XcodeProj.swift
+++ b/Sources/xcodeproj/XcodeProj.swift
@@ -25,8 +25,7 @@ public struct XcodeProj {
         if pbxprojPaths.count == 0 {
             throw XCodeProjError.pbxprojNotFound(path: path)
         }
-        pbxproj = try PBXProj(path: pbxprojPaths.first!,
-                              name: path.lastComponentWithoutExtension)
+        pbxproj = try PBXProj(path: pbxprojPaths.first!)
         let xcworkspacePaths = path.glob("*.xcworkspace")
         if xcworkspacePaths.count == 0 {
             throw XCodeProjError.xcworkspaceNotFound(path: path)
--- a/Sources/xcodeproj/XcodeProj.swift
+++ b/Sources/xcodeproj/XcodeProj.swift
@@ -7,9 +7,6 @@ public struct XcodeProj {
     
     // MARK: - Properties
     
-    /// Project path
-    public let path: Path
-    
     // Project workspace
     public let workspace: XCWorkspace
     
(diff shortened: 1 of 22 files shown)


Sources/xcodeproj/XcodeProj.swift
```diff
@@ -27,33 +25,68 @@ public struct XcodeProj {
...
 }
 
+// MARK: - <Writable>
+
+extension XcodeProj: Writable {
+
+    public func write(path: Path, override: Bool = true) throws {
+        try path.mkpath()
+
+        // write workspace
```
pepibumur: I'd create `fileprivate` methods instead of using comments.

yonaskolb:
Oh, there's a failing test here, let me fix it. (And also rebase)

Sources/xcodeproj/XcodeProj.swift
```diff
@@ -27,33 +25,68 @@ public struct XcodeProj {
...
+}
+
+// MARK: - XcodeProj Extension (Equatable)
+
+extension XcodeProj: Equatable {
+
+    public static func == (lhs: XcodeProj, rhs: XcodeProj) -> Bool {
+        return lhs.workspace == rhs.workspace &&
+            lhs.pbxproj == rhs.pbxproj
+            //TODO: make SharedData equatable: lhs.sharedData == rhs.sharedData
```
pepibumur: I'd do this on a different PR, also conforming the protocol from `XCSharedData`.

pepibumur:
Great job @yonaskolb, I left a couple of comments but in general looks ok. Answering your comments in the description:

> Removal of path property from top level objects.
I agree with this point, it gives you the flexibility to decide where to write the model.

> Use PathKit for writing
Sounds good! I passed the `FileManager` dependency to inject the behaviour for unit testing. However, I ended up testing the integration directly with the filesystem. If in the future we plan to unit-test it *(besides the integration test)* we can consider injecting a dependency.

pepibumur:
There seems to be a conflict in the `XCWorkspace` 😛

yonaskolb:
Should be good to go!

pepibumur:
Merging @yonaskolb

PR #23, opened 2017-07-23
Convert booleans to YES and NO

yonaskolb:
This fixes cases where build settings contain booleans instead of YES and NO, and writes them properly so Xcode can read them


pepibumur:
Good catch @yonaskolb 👏 . It'd be great to separate the `PBXProjWriter` in smaller components that are easier to test, so that we could unit test these little things.

PR #32, opened 2017-07-31
Reference Generator

### LINKED ISSUE #29: Linux and Crypto
It seems that CommonCrypto is the only holdout for getting this working on Linux. I know that usually anything to do with Xcode probably won't run in a Linux machine but it's still nice to have.

As far as I can see CommonCrypto is only used for getting an md5 of a string that is then used in `generateUUID()`. A hash function should work fine here as the counter is used to resolve any duplicates

### LINKED ISSUE #31: Generate UUIDs
It would also be nice to:
- generate reference UUIDs automatically
- generate them so they are deterministic

yonaskolb:
Fixes #31, Fixes #29 

This adds a new `ReferenceGenerator` class which automatically creates deterministic references for ProjectElement's

There are a few things in this PR:

- removal of Common Crypto
- addition of ReferenceGenerator
- changed `reference: String` to an optional parameter in Project Element initializers. If it is not specified, then reference is automatically generated via a default ReferenceGenerator
- tests

This enabled the following features:

- linux support with the removal of Common Crypto
- fixes old generator not generating unique references
- deterministic references
- automatic references when initializing objects

Things to consider:
- should `ProjectElement.reference` be renamed `ProjectElement.id`?
- the automatic generation in the inits is happening via a shared `ReferenceGenerator`. When should this be `reset()` and cleared of cached references, for both memory (negligible) and for determinism. If it's a manual process left up to the user, how is this documented?
- should references be initialized with a totally random and guaranteed string and then only on write make them deterministic. This might solve the above issue

### DIFF
--- a/Sources/xcodeproj/ReferenceGenerator.swift
+++ b/Sources/xcodeproj/ReferenceGenerator.swift
@@ -1,0 +1,45 @@
+//
+//  ReferenceGenerator.swift
+//  xcodeproj
+//
+//  Created by Yonas Kolb on 31/7/17.
+//
+//
+
+import Foundation
+
+
+/// Generates unique reference
+public class ReferenceGenerator {
+
+    private var references: Set<String> = []
+
+    public init() {
+
+    }
+
+    /// This resets all the cached references
+    public func reset() {
+        references = []
+    }
+
+
+    /// This generates a unique 20 character reference id to be used in ProjectElement.reference
+    ///
+    /// - Parameters:
+    ///   - element: The type of Project Element to generate for
+    ///   - id: some string value that uniquely defines the object. It doesn't have to be completely unique as duplicates will be handled, but it provides deterministic references
+    /// - Returns: The generated unique reference
+    public func generateReference<T: ProjectElement>(_ element: T.Type, _ id: String) -> String {
+        var reference = ""
+        var counter = 0
+        let className = String(describing: T.self).replacingOccurrences(of: "PBX", with: "")
+        let classAcronym = String(className.characters.filter { String($0).lowercased() != String($0) })
+        let stringID = String(abs(id.hashValue).description.characters.prefix(18 - classAcronym.characters.count))
+        repeat {
+            counter += 1
+            reference = "\(classAcronym)\(stringID)\(String(format: "%02d", counter))"
+        } while (references.contains(reference))
+        references.insert(reference)
+        return reference
+    }}
--- a/Sources/xcodeproj/ReferenceGenerator.swift
+++ b/Sources/xcodeproj/ReferenceGenerator.swift
@@ -1,16 +1,9 @@
-//
-//  ReferenceGenerator.swift
-//  xcodeproj
-//
-//  Created by Yonas Kolb on 31/7/17.
-//
-//
-
 import Foundation
 
-
-/// Generates unique reference
+/// Generates unique deterministic reference. Must be reset if generating for a new project if totally deterministc references are wanted
 public class ReferenceGenerator {
+
+    static let shared = ReferenceGenerator()
 
     private var references: Set<String> = []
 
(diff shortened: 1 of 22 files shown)


yonaskolb:
Thoughts @pepibumur?

Sources/xcodeproj/ReferenceGenerator.swift
```diff
@@ -0,0 +1,38 @@
...
+    ///
+    /// - Parameters:
+    ///   - element: The type of Project Element to generate for
+    ///   - uniqueId: some string value that uniquely defines the object.
+    ///         It doesn't have to be completely unique as duplicates will be handled, but it provides deterministic references
+    /// - Returns: The generated unique reference
+    public func generateReference<T: ProjectElement>(_ element: T.Type, _ uniqueId: String) -> String {
+        var reference = ""
+        var counter = 0
+        let className = String(describing: T.self).replacingOccurrences(of: "PBX", with: "")
```
pepibumur: There are some other files that don't start with `PBX`, for example `XCConfigurationList`. Should we also consider them here?

yonaskolb: Good catch. I'll add that prefix check 👍

pepibumur:
Hey @yonaskolb, great job! Here are my thoughts:
- Great idea removing common crypto. Although we don't have Xcode for Linux I think it's a great idea:
  - You don't need macOS on your CI if the only purpose of your project is just generating Xcode projects.
  - If eventually we have Xcode or another IDE for Linux, we'll have a project/API that allows interacting with Xcode projects.
- Regarding the generation of ids: I think the deterministic approach of generating ids might eventually lead to duplicated ids. If I understood it properly, I think that can happen if you follow the following steps:
  - You open `ProjectX.pbxproj` that has been generated with xcodeproj. When this project was created the `ReferenceGenerator` ensured that there was no duplicate by having the generated references in memory.
  - The project contains a `PBXGroup`, with name `Public`.
  - We add another group to the project with the same name but in a different path. Since the `ReferenceGenerator` instance is fresh, the list of references is empty, and the reference that we'll get will be the same as the one we already have in the project.
  - We'll have a collision with two elements with the same reference.

The way CocoaPods xcodeproj handles this is by auto-solving those collisions when the project is about to be save. We could do the same, or find a way to prevent it when new elements get created. The problem that I see with the second approach is that it'd require to somehow know the `PBXProj` where the element will be inserted. What you suggest at the end sounds a better approach, starting with something random and making it deterministic on write. If even with that there are collision we could solve them when the project gets written.

What do you think @yonaskolb ?

yonaskolb:
I've opened a seperate pull request just to remove common crypto so that #35 can be closed without this. I'll rebase this off master once that is merged.

That's a good point about loading a project and then adding new files. I guess the generator could be given the existing objects, so it would know what already exists, but that seems a bit manual.
In your specific example the PBXGroup passes both name and path to the unique string id, so it might catch that issue. An object should really pass in all it's information to be unique (like children), but I thought that would bring the determinism down. It should be done however. 
But thinking about  it now, the mutable changes break this as you can change properties but then the reference wouldn't change as it's set in the initializer. This is another case for creating deterministic references on write instead of creation. That means the object inits don't need a reference init param at all as they can be completely random. I only left them in so people could create a reference with better information to make it more deterministic.

yonaskolb:
I'll have to have a think about the best way to do references on write while keeping references intact. 
If these were all classes it would be easier as you can just pass in the class that something is supposed to reference, and then write the reference ids on write. But I guess we want to keep these all as structs.

Then we need to pass in something in the pbx writer that keeps track of everything and has a lookup table for old and new references.

What do you think about renaming reference to id?

yonaskolb:
rebased off master

pepibumur:
[Review Reference Generator](https://trello.com/c/Qsqi0Vg3/15-review-reference-generator)

yonaskolb:
I don't have access to that Trello card

pepibumur:
Lol, forget about it @yonaskolb. I integrated it with my personal Trello account and it posts the link on the PR automatically.

pepibumur:
Still pending to review this one @yonaskolb and get it merged :stuck

pepibumur:
@yonaskolb replying to your last two comments.
- I think renaming `reference` to `id` is a good idea.
- Regarding the generation of UUIDs, I think we can go with your approach but we'd need some validation when the project is written. In case we detect duplicates we can just update the reference of those files to solve the conflict between the identifiers.

I found the following resources that might be useful
- [uuid_generator](https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project/uuid_generator.rb)
- [predictabilize_uuids](https://github.com/CocoaPods/Xcodeproj/blob/c27b3fe8b464ecd365b89fb5752700fa33d65104/lib/xcodeproj/project.rb#L371)

rahul-malik:
@pepibumur @yonaskolb - Is UUID generation unstable currently? Does this need to be merged?

yonaskolb:
UIID generation in xcproj via PBXProj.generateUUID(..) works if you add every object to the project before generating a new UUID.
XcodeGen uses it's own UUID generation however to be deterministic

PR #33, opened 2017-07-31
Mutable project

### LINKED ISSUE #21: Mutable properties
Since everything is a Struct, should we make all the properties mutable? 
- mutability of a struct can be controlled by using let by whatever contains it, doesn't need to be enforced my it's members
- makes it easier to change values of properties to edit projects (if allowed by above)
- removes the need for functions that return a copy (like `adding(object:)`, or `removing(object:)` functions in PBXProj)
- no danger of things changing from under you like with Classes, where immutable properties pattern would be useful

Thoughts @pepibumur?

yonaskolb:
This makes all the properties in a project mutable, which resolves #21 

This PR has 2 commits
- making properties mutable
- removing all the previously - but no longer - necessary `adding` and `removing` functions

There is an argument to be made for leaving the `adding` and `removing` functions, but in my opinion removing them now before a 1.0 release might be the better option. It removes a ton of code and tests, and means we won't have to support and maintain those going forward. Also if they are there people will expect other methods.
Thoughts @pepibumur?


pepibumur:
image
I have no concern. Only that you removed more than 1000 lines of code :trollface:

PR #35, opened 2017-08-01
Support CocoaPods

pepibumur:
**Issue:** [Link](https://github.com/carambalabs/xcodeproj/issues/34)

### Short description
Support CocoaPods

### Solution
I flatted the project hierarchy and included a `.podspec` with the specification of the library.

### Implementation
- [x] Flat the library folders hierarchy.
- [x] Include a `.podspec` file in the repository
- [x] Update the RELEASE process.
- [x] Release a new version of xcodeproj.

### GIF
gif

PR #36, opened 2017-08-01
Remove CommonCrypto

### LINKED ISSUE #29: Linux and Crypto

yonaskolb:
Fixes #29, and resolves blocker for #35


pepibumur:
🎉  :shipit:

PR #39, opened 2017-08-08
Fixes #38

### LINKED ISSUE #38: Crash when opening Framework
## Context

Hey !

First, thanks for your work! 

I'm building a project that uses your library to get all the files within a given Xcode project. It works well for getting the information on iOS/macOS projects but it crashed when I try it on Frameworks (e.g. Alamofire). 

I'm on macOS 10.12.6 and I'm using the version 0.0.9.

## What

When I try to open a Framework, it crashed at this [line](https://github.com/carambalabs/xcodeproj/blob/master/Sources/XCScheme.swift#L113) because the attribute `runnableDebuggingMode` does not exist.

The function is called on this [line](https://github.com/carambalabs/xcodeproj/blob/master/Sources/XCScheme.swift#L289) and the attribute `BuildableProductRunnable` passed in parameter does not exist either.

## Proposal

I saw on this [line](https://github.com/carambalabs/xcodeproj/blob/master/Sources/XCScheme.swift#L108) that if the value is not present, you set it to `"0"` by default. 

Maybe you could do something like this:
```
self.runnableDebuggingMode = element.attributes["runnableDebuggingMode"] ?? "0"
```
Where if the value does not exist, it sets the field to `"0"`. I've tried it on my project and it works with no issue. If you're ok with it, I can do a pull request.

Thanks !

Shakarang:
**Issue:** [Link](https://github.com/carambalabs/xcodeproj/issues/38)

### Short description
I added a default value instead of force unwrap. It resolves the crash when opening frameworks.

### GIF
Hey

### DIFF
--- a/Sources/XCScheme.swift
+++ b/Sources/XCScheme.swift
@@ -110,7 +110,7 @@ public struct XCScheme {
             self.runnableDebuggingMode = runnableDebuggingMode
         }
         public init(element: AEXMLElement) throws {
-            self.runnableDebuggingMode = element.attributes["runnableDebuggingMode"]!
+            self.runnableDebuggingMode = element.attributes["runnableDebuggingMode"] ?? "0"
             self.buildableReference = try BuildableReference(element:  element["BuildableReference"])
         }
         public func xmlElement() -> AEXMLElement {


yonaskolb:
Thanks @Shakarang!

PR #48, opened 2017-08-24
String escaping

### LINKED ISSUE #37: Escape newlines and quotes in PBXProj

yonaskolb:
This changes whether a string is quoted when written from a blacklist to a whitelist. This makes it much more foolproof.
This fixes https://github.com/yonaskolb/XcodeGen/issues/24 by now properly quoting string with `@` in them.

This also fixes #37 by escaping quotes and newlines within a string with `/`

### DIFF
--- a/Sources/CommentedString.swift
+++ b/Sources/CommentedString.swift
@@ -18,6 +18,35 @@ struct CommentedString {
         self.string = string
         self.comment = comment
     }
+
+    private static let invalidCharacters = CharacterSet.alphanumerics.inverted
+        .subtracting(CharacterSet(charactersIn: "_./"))
+
+    var validString: String {
+        switch string {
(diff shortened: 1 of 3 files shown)


yonaskolb:
There's an issue here where if `"` or `\n` is already escaped for some reason, it will incorrectly double escape it. Let me fix that first

yonaskolb:
Updates:
- added tests
- moved escaping into CommentedString for tests
- don't escape already escaped newlines and quotes

Should be good to merge now @pepibumur. Would be good to get a hotfix out

pepibumur:
Awesome 😛, great job @yonaskolb. I'll merge it and release a new version when I merge this https://github.com/carambalabs/xcodeproj/pull/50

PR #50, opened 2017-08-24
[WIP] Move structs to classes

pepibumur:
**Issue:** [Link](https://github.com/carambalabs/xcodeproj/issues/45)

### Short description
Move structs to classes.

### Solution
Just change structs to be classes and do all the necessary changes to get it working. It shouldn't have any impact on the interfaces of these entities.

### GIF
gif

### DIFF
--- a/Package.pins
+++ b/Package.pins
@@ -8,12 +8,6 @@
       "version": "4.1.0"
     },
     {
-      "package": "CCommonCrypto",
-      "reason": null,
-      "repositoryURL": "https://github.com/Bouke/CCommonCrypto.git",
-      "version": "1.0.0"
-    },
-    {
       "package": "PathKit",
       "reason": null,
       "repositoryURL": "https://github.com/kylef/PathKit.git",
--- a/Sources/BuildSettings.swift
+++ b/Sources/BuildSettings.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// Struct that represents the Xcode build settings.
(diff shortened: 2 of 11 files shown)


yonaskolb:
Looks good. A couple of things:
- I guess you haven't finished yet cause there's still a lot of top level structs (a search and rename should make the job easier 😄 )
- I would vote for making ProjectElement, PBXTarget and a PBXBuildPhase abstract classes. That would remove a whole bunch of duplicated code with the property declarations and the init contents. I'm not usually into large inheritance hierarchies, but it makes sense here, and we may as well match it to how Apple probably designed these classes in the beginning.
- is there a specific reason to use the typed init from the dictionary init as opposed to setting the values like before? If it's because of the compiler complaining then you could move that dictionary init from an extension into the class

pepibumur:
Replying @yonaskolb 
- Yeah, I'm more into doing it step by step and then fix the issues along the way. But in this case, search and replace would have been faster.
- That sounds a great idea. I'm also against having very deep hierarchies but since in this case, it's only a two-level one, I think it's fine and easy to understand.
- Not really. When I moved the structs into classes it forced me to define the `init` in the extension as a convenience initializer, and you have to then use another initializer. By moving the `init` to the main definition we avoid this problem 😛

pepibumur:
Closing this one. We can continue the discussion here https://github.com/carambalabs/xcodeproj/pull/51/files

PR #81, opened 2017-09-19
Fix build phase script error under Xcode 9 GM

kixswift:
It seems under Xcode 9 build will fail if shell path is `bin/sh`.

Make it `/bin/sh`.

- [x] Generate a basic project with inline build shell script with `xcodegen`.
- [x] Copy the project and recreate that build shell script phase manually with Xcode and did diff with original.
- [x] Turned out that the project cannot be built because of shell path.

### DIFF
--- a/Sources/PBXShellScriptBuildPhase.swift
+++ b/Sources/PBXShellScriptBuildPhase.swift
@@ -20,7 +20,7 @@ public class PBXShellScriptBuildPhase: PBXBuildPhase, Hashable {
 
     /// Shell script.
     public var shellScript: String?
-    
+
     /// Show environment variables in the logs.
     public var showEnvVarsInLog: UInt?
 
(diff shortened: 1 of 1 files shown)


pepibumur:
Nice job @kixswift

kixswift:
Thanks 😄
It was the first ever pull request I made in my life.

pepibumur:
Really? Congrats! Did you get an invite to join the contributors group? You can also join the Slack channel using the [following link](https://camo.githubusercontent.com/c9b1aff84e08381ef57b7b33949acdc655d0ad10/68747470733a2f2f73776966742d78636f64652e6865726f6b756170702e636f6d2f62616467652e737667)

kixswift:
@yonaskolb, that’s true. I heavily rely on xcodegen, which is why made me to find the cause. You can pass it by specifying the shell path manually in spec, though.

PR #91, opened 2017-09-30
Fix search/replace issue in README

Mazyod:
**Issue:** None

### Short description
We all love xcproj's shiny new name, but perhaps we shouldn't update swift subcommand so it actually runs.

### Solution
Revert to the correct subcommand.

### Implementation

- [x] Cloned, tried to contribute, failed.
- [x] After some research, found the issue, fixed the command with SublimeText (Latest version!)


pepibumur:
Good catch @Mazyod. Merging :)

pepibumur:
I've added you as contributors @Mazyod so that you can write to the repository. Moreover, I'd like to invite you to the organization Slack http://xcodeswift.herokuapp.com/ where other members are 😛

Mazyod:
Thanks! Btw, somehow the git repo url changed to a seemingly old org name

pepibumur:
Solved @Mazyod 😛

PR #94, opened 2017-10-01
Fix classes type: array => dictionary

toshi0383:
Fixed wrong type of field `classes`.
It turns out that `classes` is dictionary, not array.

```diff
 // !$*UTF8*$!
 {
        archiveVersion = 1;
-       classes = (
-       );
+       classes = {
+       };
```


pepibumur:
Good catch @toshi. Merging 😛

PR #97, opened 2017-10-02
Make targetProxy property optional in PBXTargetDependency

pepibumur:
### Short description 📝
I've got some issues opening a spm generated project because `PBXTargetDependency` elements didn't have the targetProxy property.

### Solution 📦
Make the property optional.

### Implementation 👩‍💻👨‍💻
- [x] Make the property optional.
- [x] Update the tests.

### GIF
gif

### DIFF
--- a/Sources/xcproj/PBXTargetDependency.swift
+++ b/Sources/xcproj/PBXTargetDependency.swift
@@ -10,7 +10,7 @@ public class PBXTargetDependency: PBXObject, Hashable {
     public var target: String
     
     /// Target proxy
-    public var targetProxy: String
+    public var targetProxy: String?
     
     // MARK: - Init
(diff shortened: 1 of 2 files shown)


pepibumur:
Interesting @toshi0383. I've checked the project again and now they include the `targetProxy` property. I think that property must be non-optional. It makes no sense having a target dependency that proxies to nothing. If we find a target dependency without that property we should throw and let the developer know about a wrong format. What do you think @toshi0383?

pepibumur:
I'll close this because we'll approach this problem on a different way:
- Improving the error handling and detecting this issues before reading the project.
- Maybe filling those attributes with default data so that the user can at least open it with xcproj and fix them from there.

PR #99, opened 2017-10-02
Parse objects conforming the Codable protocol

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/83

### Short description 📝
Remove Unbox dependency by using the native Codable protocol.

### Solution 📦
Update all the elements to conform the Codable protocol and remove the Unboxable conformance.

### Implementation 👩‍💻👨‍💻

- [x] PBXFileElement
- [x] PBXBuildFile
- [x] PBXAggregateTarget
- [x] PBXBuildPhase
- [x] PBXContainerItemProxy
- [x] PBXCopyFilesBuildPhase
- [x] PBXFileReference
- [x] PBXFrameworksBuildPhase
- [x] PBXGroup
- [x] PBXHeadersBuildPhase
- [x] PBXNativeTarget
- [x] PBXProductType
- [x] PBXProj
- [x] PBXProject
- [x] PBXReferenceProxy
- [x] PBXResourcesBuildPhase
- [x] PBXShellScriptBuildPhase
- [x] PBXSourceTree
- [x] PBXSourcesBuildPhase
- [x] PBXTarget
- [x] PBXTargetDependency
- [x] PBXVariantGroup
- [x] XCBuildConfiguration
- [x] XCVersionGroup
- [x] XCWorkspace
- [x] XCWorkspaceData

### GIF
gif


Tests/xcprojTests/PBXFileElementSpec.swift
```diff
@@ -36,26 +36,32 @@ final class PBXFileElementSpec: XCTestCase {
     func test_init_failsIfSourceTreeIsMissing() {
         var dictionary = testDictionary()
         dictionary.removeValue(forKey: "sourceTree")
+        let data = try! JSONSerialization.data(withJSONObject: dictionary, options: [])
+        let decoder = JSONDecoder()
         do {
-            _ = try PBXFileElement(reference: "ref", dictionary: dictionary)
```
yonaskolb: It's probably worth making an extension on PBXObject or Decodable which is `init(jsonDictionary: [String: Any]` that handles all this boilerplate

artemnovichkov:
Can I help you with it? `PBXBuildFile` is next

pepibumur:
Sure @artemnovichkov, you are welcome to help. It'll take us time to migrate all of them, but with help, we'll be faster 😛

Tests/xcprojTests/PBXFileElementSpec.swift
pepibumur: Yeah good idea, I'll do it @yonaskolb

Sources/xcproj/PBXTarget.swift
```diff
@@ -47,18 +46,31 @@ public class PBXTarget: PBXObject, Hashable {
...
+    }
+    
+    public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.buildConfigurationList = try container.decode(String.self, forKey: .buildConfigurationList)
+        self.buildPhases = try container.decode([String].self, forKey: .buildPhases)
+        self.buildRules = try container.decode([String].self, forKey: .buildRules)
+        self.dependencies = try container.decode([String].self, forKey: .dependencies)
+        self.name = try container.decode(String.self, forKey: .name)
+        self.productName = try container.decode(String?.self, forKey: .productName)
```
yonaskolb: Optional properties should use ‘decodeIfPresent’

Tests/xcprojTests/PBXFileElementSpec.swift
pepibumur: Done @yonaskolb

Sources/xcproj/XCVersionGroup.swift
```diff
@@ -64,6 +52,31 @@ public class XCVersionGroup: PBXObject, Hashable {
...
+    }
+
+    public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.currentVersion = try container.decode(String.self, forKey: .currentVersion)
+        self.path = try container.decode(String.self, forKey: .path)
+        self.name = try container.decodeIfPresent(String.self, forKey: .name)
+        self.sourceTree = try container.decode(PBXSourceTree.self, forKey: .sourceTree)
+        self.versionGroupType = try container.decode(String.self, forKey: .versionGroupType)
+        let children = try? container.decode([String].self, forKey: .children)
```
yonaskolb: I think lines like this can be:
```
self.children = try container.decodeIfPresent([String].self, forKey: .children) ?? []
```

Tests/xcprojTests/PBXFileElementSpec.swift
yonaskolb: Nice! I think you could also put an extension on PBXObject as well that is an init with reference
```
init(reference: String, jsonDictionary: [String: Any]
```
That would remove a bit of the boilerplate in tests and other places

yonaskolb:
I find this extension really useful when using Decodable

```swift
extension KeyedDecodingContainer {

    public func decode<T>(_ key: KeyedDecodingContainer.Key) throws -> T where T: Decodable {
        return try decode(T.self, forKey: key)
    }

    public func decodeIfPresent<T>(_ key: KeyedDecodingContainer.Key) throws -> T? where T: Decodable {
        return try decodeIfPresent(T.self, forKey: key)
    }
}
```

It makes the code easier to write as you don't need to specify the type again:

```diff
- self.name = try container.decode(String.self, forKey: .name)
+ self.name = try container.decode(.name)
```

It also means that support can be added by doing search and replace on the unbox functions

pepibumur:
It's very useful! Great idea @yonaskolb. I'll update the code to use it.

yonaskolb:
I just noticed that this removed the BuildSettings typealias. Was there any particular reason for that?

pepibumur:
I came across some issues trying to parse the settings, and I removed it thinking it was related. It turned out it wasn't. Do you think we should keep it?

yonaskolb:
It’s not strictly needed, but I think it was nice how it unified the types and purpose wherever build settings were used. The factthat it was public means people may have been using it too (as was the case of XcodeGen)

PR #106, opened 2017-10-07
Prevent errors parsing misformated project using default values

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/101

### Short description 📝
Assign default values to those values that are missing. By doing that, we allow developers open their projects and update them. To verify that the format is correct, they could use xclint and run it before saving their projects.

### Solution 📦
Update the objects init method and their respective tests.

### Implementation 👩‍💻👨‍💻
- [x] Update constructors.
- [x] Update tests.

### GIF
gif

### DIFF
--- a/Sources/xcproj/PBXBuildFile.swift
+++ b/Sources/xcproj/PBXBuildFile.swift
@@ -37,10 +37,9 @@ public class PBXBuildFile: PBXObject, Hashable {
     
     public required init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-        let reference: String = try container.decode(.reference)
-        self.fileRef = try container.decode(.fileRef)        
+        self.fileRef = try container.decodeIfPresent(.fileRef) ?? ""
         self.settings = try container.decodeIfPresent([String: Any].self, forKey: .settings) ?? [:]
-        super.init(reference: reference)
+        try super.init(from: decoder)
     }
     
     // MARK: - Hashable
--- a/Sources/xcproj/PBXBuildPhase.swift
+++ b/Sources/xcproj/PBXBuildPhase.swift
@@ -33,13 +33,12 @@ public class PBXBuildPhase: PBXObject {
     
     public required init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-        let buildActionMaskString: String = try container.decode(.buildActionMask)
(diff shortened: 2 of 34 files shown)


toshi0383:
Oh I don’t think it’s a good idea to assign default value for missing value.❌

pepibumur:
@toshi0383 there's an open discussion on this [issue](https://github.com/xcodeswift/xcproj/issues/101). The problem by throwing errors if the properties are missing is that developer is not able to fix those issues because the project doesn't get parsed. We could solve it by using optionals, but we think that working with optionals will make all the code more verbose, full of if guard statements. 
This change would be complemented with the xclint validator that the developer can use to make sure all the properties are valid.

toshi0383:
Yea I saw that conversation. This is my opinion.
1. By default xcproj should fail parsing on missing required field. (throw error)
2. To make xclint to work with invalid pbxproj, we should be able to inject different Decoder from outside of xcproj framework.

pepibumur:
Why do you think xcproj should fail the parsing? Projects generated by the Swift Package Manager cannot be read with xcproj, however, they can be opened with Xcode. They don't seem to be invalid for Xcode but they are xcproj. I see this very confusing for a developer using the tool. Let's say that I want to read my project to do some changes, and I get an error.

toshi0383:
1. In general, a Parser should fail on unexpected structure. Setting default value for missing value is unexpected behavior for a Parser. Do we agree on this?
2. I think it's okay to make some fields Optional for SPM compatibility. If Xcode succeeds to open, xcproj should too.

So the solution should be "Make some fields optional for SPM compatibility", not "Assign default values to those values that are missing".

pepibumur:
The rule to decide the optionality of an attribute could be, is Xcode able to handle a nil value for a given property? What do you think @esttorhe / @yonaskolb

esttorhe:
I agree with @toshi0383; maybe we shouldn't assign a default value if missing except `nil` as long as `Xcode` can handle the missing value.

If `Xcode` can't handle the missing value we should fail the parsing.

Or if we want to go the extra mile provide an `optimistic` parsing that will add a default value to the missing fields (kind of like `Xcode` does)

pepibumur:
Closing the PR. I opened [another one](https://github.com/xcodeswift/xcproj/pull/107) where I'll tackle this differently based on your suggestions. I'll review the optionality of the values using Xcode.

PR #107, opened 2017-10-10
Review optionality of attributes

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/101

### Short description 📝
Review which attributes should be optional or not. 

### Solution 📦
For each attribute, I remove it from the object and try to open the project with Xcode. If Xcode can handle not having the value, then the attribute is optional, if the project is considered invalid by Xcode, then the attribute is required.

### Implementation
- [x] PBXAggregateTarget
- [x] PBXBuildFile
- [x] PBXBuildPhase
- [x] PBXContainerItemProxy
- [x] PBXCopyFilesBuildPhase
- [x] PBXFileElement
- [x] PBXFileReference
- [x] PBXFrameworksBuildPhase
- [x] PBXGroup
- [x] PBXHeadersBuildPhase
- [x] PBXNativeTarget
- [x] PBXObject
- [x] PBXProj
- [x] PBXReferenceProxy
- [x] PBXResourcesBuildPhase
- [x] PBXShellScriptBuildPhase
- [x] PBXSourcesBuildPhase
- [x] PBXTarget
- [x] PBXTargetDependency
- [x] PBXVariantGroup
- [x] XCBuildConfiguration
- [x] XCConfigurationList
- [x] XCScheme
- [x] XCSharedData
- [x] XCVersionGroup
- [x] XCWorkspaceData

### GIF
gif

### DIFF
--- a/Sources/xcproj/PBXBuildPhase.swift
+++ b/Sources/xcproj/PBXBuildPhase.swift
@@ -4,7 +4,7 @@ import Foundation
 public class PBXBuildPhase: PBXObject {
 
     /// Element build action mask.
-    public var buildActionMask: UInt
+    public var buildActionMask: UInt?
 
     /// Element files.
     public var files: [String]
@@ -14,7 +14,7 @@ public class PBXBuildPhase: PBXObject {
 
     public init(reference: String,
                 files: [String] = [],
-                buildActionMask: UInt = 2147483647,
+                buildActionMask: UInt? = nil,
                 runOnlyForDeploymentPostprocessing: UInt = 0) {
         self.files = files
         self.buildActionMask = buildActionMask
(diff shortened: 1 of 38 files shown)


Sources/xcproj/PBXBuildPhase.swift
```diff
@@ -14,7 +14,7 @@ public class PBXBuildPhase: PBXObject {
 
     public init(reference: String,
                 files: [String] = [],
-                buildActionMask: UInt = 2147483647,
```
esttorhe: Yes!! No more magic ✨  numbers!

Sources/xcproj/PBXContainerItemProxy.swift
```diff
@@ -61,9 +61,9 @@ public class PBXContainerItemProxy: PBXObject, Hashable {
     public required init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.containerPortal = try container.decode(.containerPortal)
-        let proxyTypeString: String = try container.decode(.proxyType)
-        self.proxyType = UInt(proxyTypeString).flatMap({ProxyType(rawValue: $0)}) ?? .other
-        self.remoteGlobalIDString = try container.decode(.remoteGlobalIDString)
+        let proxyTypeString: String? = try container.decodeIfPresent(.proxyType)
```
yonaskolb: I noticed when the decoding changes happened there are a whole bunch of these cases where something is decoded into a string and then flatMaped into a UInt. Is there a specific reason for that, as opposed to decoding directly to a UInt?

yonaskolb: Does this not work?
```
self.proxyType = try container.decodeIfPresent(UInt.self, forKey: . proxyType).flatMap(ProxyType.init) ?? .other
```

pepibumur: Yeah the decoder throws because doesn't know how to convert those values into integers @yonaskolb

yonaskolb: Ah ok, makes sense

PR #116, opened 2017-10-13
Add more examples to the README

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/44

### Short description 📝
Add some examples to the README that show some real use cases.

### Solution 📦
Update the README adding those examples.

### Implementation 👩‍💻👨‍💻
- [x] Creating a group inside another group
- [x] Adding a file inside a group
- [x] Adding a existing file as a source file

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/116)

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,13 +121,12 @@ A `XcodeProj` has the following properties:
 - `XCWorkspace` that defines the structure of the project workspace.
 - `PBXProj` that defines the strcuture of the project.
 
-Amongt other properties, the most important one in the `PBXProj` object is `Objects`. Projects are defined by a list of those objects that can be classified in the following groups:
+Among other properties, the most important one in the `PBXProj` object is `Objects`. Projects are defined by a list of those objects that can be classified in the following groups:
 
 - **Build phases objects**: Define the available build phases.
-- **Targets objects**: Define your project targets and dependencies between them.
+- **Target objects**: Define your project targets and dependencies between them.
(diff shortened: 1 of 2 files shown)


pepibumur:
@Daltron is there any other example that you'd like us to include?

README.md
```diff
@@ -129,6 +128,28 @@ You can read more about what each of these objects is for on the [following link
...
 - New versions of Xcode might introduce new models or property that are not supported by `xcproj`. If you find any, don't hesitate to [open an issue](https://github.com/xcodeswift/xcproj/issues/new) on the repository.
 
+## Examples
+
+### Creating a group inside another group
+
+```swift
+let project = try! XcodeProj(path: "myproject.xcodeproj")
+let myGroup = PBXGroup(reference: "xxx", children: [], sourceTree: .group, path: "MyGroup")
+project.pbxproj.append(myGroup)
```
yonaskolb: This wouldn't actually make the group visible. You should probably use the reference generator as well.
Also is `pbxproj.append` a valid function?

pepibumur: It should be `project.pbxproj.objects.append(myGroup)`. What does it need to be visible? Adding it to another group?
Regarding the reference generator, since it's not merged yet, I'd say leave it like that until we have something in place to handle them. what do you think?

pepibumur: I've updated the examples. What do you think @yonaskolb ?

PR #123, opened 2017-10-19
Add new init to initialize the workspace and the project with a path string

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/121

### Short description 📝
The only init method available took a `PathKit.Path` value to initialize the model. That forced developers to import `PathKit` and pass `Path` values instead of just simply a `String` that represents the path.

### Solution 📦
Add a new convenience init that accepts a path `String`.

### GIF
gif

### DIFF
--- a/Sources/xcproj/XCWorkspace.swift
+++ b/Sources/xcproj/XCWorkspace.swift
@@ -25,6 +25,14 @@ public class XCWorkspace {
         }
         data = try XCWorkspace.Data(path: xcworkspaceDataPaths.first!)
     }
+    
+    /// Initializes the workspace with the path string.
+    ///
+    /// - Parameter pathString: path string.
+    /// - Throws: throws an error if the initialization fails.
+    public convenience init(pathString: String) throws {
+        try self.init(path: Path(pathString))
+    }
 
     /// Initializes the workspace with its properties.
     ///
--- a/Sources/xcproj/XcodeProj.swift
+++ b/Sources/xcproj/XcodeProj.swift
@@ -34,6 +34,10 @@ public class XcodeProj {
         let sharedDataPath = path + Path("xcshareddata")
         self.sharedData = try? XCSharedData(path: sharedDataPath)
     }
+    
+    public convenience init(pathString: String) throws {
+        try self.init(path: Path(pathString))
+    }
 
     /// Initializes the XCodeProj
     ///


Sources/xcproj/XCWorkspace.swift
```diff
@@ -25,6 +25,14 @@ public class XCWorkspace {
         }
         data = try XCWorkspace.Data(path: xcworkspaceDataPaths.first!)
     }
+    
+    /// Initializes the workspace with the path string.
+    ///
+    /// - Parameter pathString: path string.
+    /// - Throws: throws an error if the initialization fails.
+    public convenience init(pathString: String) throws {
```
esttorhe: Looks good; 2 things though.

1. The fact that `init(path: Path)` is still part of the `API` and public wouldn't still require users to import `Path` in order to properly compile/use the tool?
1. I'm assuming that `Path` checks if a `String` is a valid path or not but perhaps we could check in the `convenience init`s if the passed `String` is a valid path and if not `throw` with a proper/more meaningful error?

yonaskolb: 1. A user wouldn't need to import `Path` if they are not making reference to the symbol, so this is fine.

esttorhe: Thanks @yonaskolb wasn't 💯  sure it would work or not.

pepibumur:
Does it look good @yonaskolb  / @esttorhe ?

esttorhe:
@pepibumur looks good to me; I just have this one comment
>I'm assuming that Path checks if a String is a valid path or not but perhaps we could check in the convenience inits if the passed String is a valid path and if not throw with a proper/more meaningful error?

pepibumur:
Good point @esttorhe. It'd be great if they checked that but the Path constructor doesn't. What we do internally is to fail if the path is invalid or there's no file in that path. What do you think about adding some deprecation notes on the public stuff that is using `Path`. We benefit from using path internally because it has some convenient methods, but from the public interface, there isn't much value on exposing it.

yonaskolb:
I think many people using xcproj from their codebase would appreciate the path initializers. I know I use them in XcodeGen. I see no reason to remove them

esttorhe:
Yeah, I wouldn't get rid of it; I think it provides a handy way of initializing and more so `XcodeGen` relies on it.

pepibumur:
Let's keep it then 😛

PR #124, opened 2017-10-19
Support SPM generated projects

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/119

### Short description 📝
As @josefdolezal described in the issue above, projects generated by the Swift Package Manager cannot be opened with xcproj because the project.workspace property is non-optional. The purpose of this PR is to provide support for those.

### Solution 📦
Since property is public, any change in the property would introduce a breaking change in the API. What I did instead is initializing a workspace with one file reference, `self:` in the same way Xcode does.

### Implementation 👩‍💻👨‍💻
- [x] Add changes to the `XcodeProject` model.
- [x] Add some unit and integration tests.

> Note: I added a project that we use as a fixture for the test that I added.

### GIF
gif


Sources/xcproj/XCWorkspace.swift
```diff
@@ -34,6 +34,12 @@ public class XCWorkspace {
         self.data = data
     }
 
+    /// Returns an empty workspace.
+    public static var `empty`: XCWorkspace {
+        let data = XCWorkspace.Data(references: [])
```
artemnovichkov: Is it possible to omit `XCWorkspace.` and use only `Data(references: [])`?

Tests/xcprojTests/XcodeProjIntegrationSpec.swift
```diff
@@ -13,28 +13,36 @@ final class XcodeProjIntegrationSpec: XCTestCase {
...
+        testWrite(from: fixtureiOSProjectPath(),
                   initModel: { try? XcodeProj(path: $0) },
                   modify: { $0 })
     }
-
-    private func fixturePath() -> Path {
-        return fixturesPath() + "iOS/Project.xcodeproj"
+    
+    func test_init_usesAnEmptyWorkspace_whenItsMissing() {
+        let got = projectWithoutWorkspace()
```
artemnovichkov: Why did you use this name?

Sources/xcproj/XCWorkspace.swift
```diff
@@ -34,6 +34,12 @@ public class XCWorkspace {
         self.data = data
     }
 
+    /// Returns an empty workspace.
+    public static var `default`: XCWorkspace {
```
yonaskolb: This makes it seem as if it's a singleton. You could remove this and just make a new initializer that doesn't take `data`.

Tests/xcprojTests/XcodeProjIntegrationSpec.swift
pepibumur: It's a project that doesn't have a workspace internally. Would you use a different name @artemnovichkov?

Sources/xcproj/XCWorkspace.swift
pepibumur: Yeah, makes more sense. Let me update it.

pepibumur: Done @yonaskolb

PR #125, opened 2017-10-19
Carthage support

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/41

### Short description 📝
`xcproj` cannot be linked from another project using Carthage. To support it, the repository needs to expose a project with targets that compile the framework for all the supported platforms.

### Solution 📦
Use [xcodegen](https://github.com/yonaskolb/xcodegen) to generate the Carthage project as part of the release process.

### Implementation 👩‍💻👨‍💻
- [x] Define `project.yml`.
- [x] Add Carthage support to PathKit. **blocker**
- [x] Update the release process.
- [x] Update ci task to validates that the project generated for Carthage is valid.

### GIF
gif

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -146,7 +146,7 @@ playground.xcworkspace
 # Carthage
 #
 # Add this line if you want to avoid checking in source code from Carthage dependencies.
-# Carthage/Checkouts
+Carthage/Checkouts
 
 Carthage/Build
 
--- a/.gitignore
+++ b/.gitignore
@@ -146,9 +146,7 @@ playground.xcworkspace
 # Carthage
 #
 # Add this line if you want to avoid checking in source code from Carthage dependencies.
-Carthage/Checkouts
-
-Carthage/Build
+Carthage/
 
 # fastlane
 #
--- a/Rakefile
+++ b/Rakefile
@@ -5,6 +5,7 @@ require 'colorize'
 require 'fileutils'
 
 DESTINATION = "platform=iOS Simulator,name=iPhone 6,OS=11.0"
+XCODEGEN_VERSION = "1.3.0"
 
 def generate_docs
   print "Executing tests"
@@ -68,7 +69,7 @@ end
 
 def generate_carthage_project
   throw "Mint is necessary. Make sure it's installed in your system" unless command?("mint")
-  sh "mint run yonaskolb/xcodegen@1.3.0"
+  sh "mint run yonaskolb/xcodegen@#{XCODEGEN_VERSION}"
 end
 
 def print(message)
--- a/Rakefile
+++ b/Rakefile
@@ -4,10 +4,10 @@ require 'semantic'
 require 'colorize'
 require 'fileutils'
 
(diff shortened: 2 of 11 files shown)


pepibumur:
This is ready for review @xcodeswift/contributors 😬

.gitignore
```diff
@@ -146,7 +146,7 @@ playground.xcworkspace
 # Carthage
 #
 # Add this line if you want to avoid checking in source code from Carthage dependencies.
-# Carthage/Checkouts
+Carthage/Checkouts
 
 Carthage/Build
```
toshi0383: Why not ignore whole `Carthage/` ?

toshi0383:
Why do you need iOS xcscheme?

toshi0383:
The name `Carthage.xcodeproj/` is not a good idea. It should be something like `carthage-support/xcproj.xcodeproj/`
See https://github.com/RxSwiftCommunity/RxNimble/pull/15#issuecomment-294087549

.gitignore
pepibumur: Good idea. I've added the whole folder.

pepibumur:
@toshi0383 someone asked support for iOS in the past because they use the tool from the simulator to do some checks. I just kept the support in Carthage as well.

pepibumur:
Why do you think `Carthage.xcodeproj` is not a good idea?

Rakefile
```diff
@@ -33,54 +51,67 @@ def bump_to_version(from, to)
...
-  `git tag #{to}`
+  if tag then `git tag #{tag}` end
   `git push origin --tags`
 end
 
-def print(message)
-  puts message.colorize(:yellow)
+def generate_carthage_project
+  throw "Mint is necessary. Make sure it's installed in your system" unless command?("mint")
+  sh "mint run yonaskolb/xcodegen@1.3.0"
```
esttorhe: 😍 👍

pepibumur: That's a good idea @esttorhe. However, I tried to do it but it's not supported by Mint yet. When you pass a parameter to specify the project using mint, mint tries to parse that parameter. I've opened an [issue](https://github.com/yonaskolb/Mint/issues/13) to track it. Do you know if we could do this somehow or we need to add support for that in Mint @yonaskolb ?

esttorhe:
ah ok; no biggie though. Just a suggestion :)

–
Esteban Torres
iOS Developer

Phone: (+49)0162-4190-543
Skype: esteban.a.torres.hernandez
Twitter: @esttorhe

toshi0383:
> Why do you think Carthage.xcodeproj is not a good idea?

@pepibumur 
Sorry for delay.
It's because you can't tell if `Carthage.xcodeproj` is from `xcproj` or other library on Xcode.
That's why [I've renamed `CarthageSupport.xcodeproj` of RxSwiftCommunity/NSObject-Rx](https://github.com/RxSwiftCommunity/NSObject-Rx/pull/35).

Issue #121, opened 2017-10-18
Project path as String


artemnovichkov:
## Context 🕵️‍♀️
I have a project path as `String`, but `XcodeProj` object has no initializers for that.

## What 🌱
This code:
```swift
do {
    let projectPath = "Test.xcodeproj"
    let project = try XcodeProj(path: projectPath)
}
catch {
    print(error)
}
```
Related to compilation error:
`Cannot convert value of type 'String' to expected argument type 'Path'`

## Proposal 🎉
Add `init(path: String)` method.

PR #130, opened 2017-10-23
Add method for writing with path string

artemnovichkov:
### Short description 📝
Extend `Writable` protocol with `String` as path.

### Solution 📦
I already told about `Path`s here: #121. I added a new method for convenience writing w/o Path framework imports and hardcoded paths.

### Implementation 👩‍💻👨‍💻
I added a new method with default implementation to prevent code duplication. What do you think about it?

### DIFF
--- a/Sources/xcproj/Writable.swift
+++ b/Sources/xcproj/Writable.swift
@@ -11,4 +11,18 @@ public protocol Writable {
     /// - Throws: writing error if something goes wrong.
     func write(path: Path, override: Bool) throws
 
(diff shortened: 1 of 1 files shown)


artemnovichkov:
Should I write of update any tests to merge it?

pepibumur:
It's not necessary @artemnovichkov since it's a convenience initializer that is using an existing one.

artemnovichkov:
Cool, in this case I've finished it.

pepibumur:
Great job @artemnovichkov. Merged! 🎉

PR #133, opened 2017-10-29
No default lastKnownFileType

toshi0383:
### Short description 📝
Reverts auto calculation of `lastKnownFileType` introduced in #58.

### Solution 📦
It turns out that Xcode sets `explicitFileType` instead of `lastKnownFileType` for macOS app target automatically. So we shouldn't set default value for that.
```
        FR4799450001 /* LGTM.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = LGTM.app; sourceTree = BUILT_PRODUCTS_DIR; };
```

### DIFF
--- a/Sources/xcproj/PBXFileReference.swift
+++ b/Sources/xcproj/PBXFileReference.swift
@@ -52,7 +52,7 @@ public class PBXFileReference: PBXObject, Hashable {
                 xcLanguageSpecificationIdentifier: String? = nil) {
         self.fileEncoding = fileEncoding
         self.explicitFileType = explicitFileType
-        self.lastKnownFileType = lastKnownFileType ?? path.flatMap { PBXFileReference.fileType(path: Path($0)) }
+        self.lastKnownFileType = lastKnownFileType
         self.name = name
         self.path = path
         self.sourceTree = sourceTree


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 🚀 Check out the guidelines [here](https://github.com/xcodeswift/contributors/blob/master/CHANGELOG_GUIDELINES.md)
 
+## master
```
pepibumur: These changes are not necessary. We'll tag this PR with the proper milestone and we'll update the Changelog accordingly when we make the next release.

Sources/xcproj/PBXFileReference.swift
```diff
@@ -52,7 +52,7 @@ public class PBXFileReference: PBXObject, Hashable {
                 xcLanguageSpecificationIdentifier: String? = nil) {
         self.fileEncoding = fileEncoding
         self.explicitFileType = explicitFileType
-        self.lastKnownFileType = lastKnownFileType ?? path.flatMap { PBXFileReference.fileType(path: Path($0)) }
+        self.lastKnownFileType = lastKnownFileType
```
pepibumur: What are your thoughts on this @yonaskolb?

CHANGELOG.md
toshi0383: Reverted👌

Sources/xcproj/PBXFileReference.swift
yonaskolb: I guess if that' really the case. I can see why removing this might be good for other reasons as well, where we might be calculating the wrong `lastKnownFileType` as `PBXFileReference.fileType(` would never be perfect

PR #134, opened 2017-11-03
added support for legacy octest

solgar:
### Short description 📝
`productType = "com.apple.product-type.bundle.ocunit-test";` in PBXNativeTarget caused crash when loading project

### Solution 📦
`case ocUnitTestBundle = "com.apple.product-type.bundle.ocunit-test"` in PBXProductType has been added to address this issue


pepibumur:
Nice catch @solgar. Would you mind adding a test that validates the string value for that enum case?

solgar:
@pepibumur Will do.

pepibumur:
👏 🎉  Nice job @solgar. I've merged it!

PR #135, opened 2017-11-03
fix for projectReferences type

solgar:
### Short description 📝
Loading project with `projectReferences` was causing crash in init from decodable.

### Solution 📦
`projectReferences` type has been changed from `[Any]` to `[[String : String]]` to reflect the document.


pepibumur:
Nice @solgar. Tests seem to be failing because of your changes. Could you update them? Then it should be ready to merge

solgar:
@pepibumur Sure. Code submit eta +12.

pepibumur:
Same here. Merging :shipit:

PR #136, opened 2017-11-08
Change PBXObject collections for performance

rahul-malik:
- [x] Add Dictionary extensions for file references
- [x] Update PBXProj to use dictionaries over arrays for various properties. These dictionaries use the reference string of PBXObject as their key.
- [x] Add `ProjEncoderBuffer` to `PBXProjEncoder` which allows for alternate data types besides `String` to be utilized for appending the final project output. I left a String implementation for reference but the default is now `Data`


Sources/xcproj/PBXObject.swift
```diff
@@ -113,6 +113,20 @@ extension Array where Element: Referenceable {
...
+extension Dictionary where Value: Referenceable {
+    public var references: [String] {
+        return self.values.map { $0.reference }
+    }
+
+    public func contains(reference: Key) -> Bool {
+        return self[reference] != nil
+    }
+
+    public func getReference(_ reference: Key) -> Value? {
```
pepibumur: We don't need this, do we? If I'm not wrong, we could just access the dictionary directly, couldn't we?

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -98,16 +98,17 @@ extension PBXProj {
 extension PBXProj: Writable {
     
     public func write(path: Path, override: Bool) throws {
-        let encoder = PBXProjEncoder()
-        let output = encoder.encode(proj: self)
         if override && path.exists {
             try path.delete()
         }
-        try path.write(output)
+        let encoder = PBXProjEncoder(with: Data())
```
pepibumur: 👏

Sources/xcproj/PBXProj.swift
```diff
@@ -22,20 +22,20 @@ public class PBXProj: Decodable {
     public var aggregateTargets: [PBXAggregateTarget] = []
     public var containerItemProxies: [PBXContainerItemProxy] = []
     public var copyFilesBuildPhases: [PBXCopyFilesBuildPhase] = []
-    public var groups: [PBXGroup] = []
+    public var groups: [String: PBXGroup] = [:]
```
pepibumur: This will be a breaking change. We'll have to do a major version bump. What do you think @yonaskolb ?

Sources/xcproj/PBXProjEncoder.swift
```diff
@@ -10,14 +10,76 @@ extension PlistSerializable {
     var multiline: Bool { return true }
 }
 
+protocol ProjEncoderBuffer {
```
pepibumur: Should we make the protocol and the extensions below `fileprivate` since they are only used here?

Sources/xcproj/PBXProjEncoder.swift
```diff
@@ -10,14 +10,76 @@ extension PlistSerializable {
...
+
+        let data = value.data(using: .utf8)!
+        _ = data.withUnsafeBytes{
+            self.write($0, maxLength: data.count)
+        }
+    }
+
+    func write(to fileURL: URL) throws {
+        // No-Op since we'll used
+        switch self.streamStatus {
```
pepibumur: It's confusing having a method in the protocol that takes `URL` and that is not used since the `OutputStream` needs to be initialized with an URL. What about having two protocols that *"inherit"* from a base one that contains the `append` method?

pepibumur:
Nice job @rahul-malik! Looking forward to those improvements. It's great that you can push the library to its limits and find areas that need some improvement. 👏

Sources/xcproj/PBXObject.swift
yonaskolb: I assume this was added for backwards compatibility 👍

Sources/xcproj/PBXProj.swift
yonaskolb: Yeah, it will be a breaking change. There is the option of creating these dictionary caches privately and on demand when we need them, which is at the beginning of writing. But perhaps have them as dictionaries is a better public API as well? 
Another side effect of this is that all duplicate references get removed automatically. This is mostly good, but means tools like `xclint` can't show you which references are duplicated. Is that ok?

In either case, I think all the properties should match, and if the public api is changing they should then all be changed to dictionaries (including getters like `buildPhases` and `objects`).

Sources/xcproj/PBXObject.swift
rahul-malik: Yep this was added to minimize updating all callers

Sources/xcproj/PBXProj.swift
rahul-malik: What does this specifically break? I tested with XCodeGen so I'm assuming it's other projects in the xcodeswift org?

Sources/xcproj/PBXProjEncoder.swift
rahul-malik: I can try that. I believe they will need to be available for the pbxproj+helpers changes.

Sources/xcproj/PBXProjEncoder.swift
rahul-malik: Sorry I meant to remove this, the output stream implementation wasnt better than a simple in-memory data buffer so this complexity is unnecessary.

rahul-malik:
Thanks for the review! I also captured instruments traces if you'd like to see the performance changes or bottlenecks

Sources/xcproj/PBXProj.swift
yonaskolb: It breaks any project that reads or writes these properties and expects them to be arrays.
eg:
```
pbxProj.fileReferences.append(myFile)
```
XcodeGen is ok as it uses `func addObject(_ object: PBXObject)`

yonaskolb:
Yeah, I'd love to know which contributes more to the speedup, the dictionary accessors or the encoding. I guess in a perfect world these would be 2 different PR's...

rahul-malik:
@pepibumur @yonaskolb - When you get a chance, I've updated to improve the performance of most fields in `PBXProj` so it's a larger change but the performance wins are substantial for large projects

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,94 @@
...
+    public func contains(reference: Key) -> Bool {
+        return self[reference] != nil
+    }
+
+    public func getReference(_ reference: Key) -> Value? {
+        return self[reference]
+    }
+}
+
+public struct ReferenceableCollection<T: Referenceable & Equatable>: Equatable, ExpressibleByArrayLiteral {
```
yonaskolb: @rahul-malik What benefits do we get out of using this struct as a opposed to a Dictionary?

rahul-malik: @yonaskolb - The benefits are largely that you have an abstraction around the internal data structure and the mutations are only additive (through `add(_)`). `PBXProj` consumes many arrays and this simplifies the conversion to the internal dictionary representation.

For me this is more about encapsulation and constraining the data type to match it's usage. 

Thoughts?

yonaskolb: I'm not quite convinced it's worth it. It seems to just be mirroring a Dictionary API with some extra helpers (that could still be added as a constrained extension on a dictionary). We also lose out on any of the new Swift 4 dictionary apis, and  - in the current form - any way to remove objects. 
Personally I would say these could all just be `[String: PBXFileReference]` like you had before.

rahul-malik: @yonaskolb - I could certainly look into changing it back. For use case I'm never removing objects from the projects but if that is a use-case we support then that makes sense.

@pepibumur - Thoughts?

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,94 @@
+//
```
pepibumur: I'd remove the header comments.

Sources/xcproj/Referenceable.swift
pepibumur: I get your point about abstraction @rahul-malik but comparing the benefits of having such abstraction, and the downside of having to maintain the interface and keep up with `Foundation` changes I think it's not that worth having such abstraction. I would see a benefit in being able to optimize stuff keeping the interface intact but I'm not sure how further we can go with optimization at the collection level once we move from arrays to dictionaries. In case we need to change the collections types in the future, we can always extend them to match the interface that we are currently exposing.

Is there any other benefit that I might be missing @rahul-malik?

Sources/xcproj/PBXProj.swift
```diff
@@ -18,34 +17,38 @@ public class PBXProj: Decodable {
...
-    public var shellScriptBuildPhases: [PBXShellScriptBuildPhase] = []
-    public var resourcesBuildPhases: [PBXResourcesBuildPhase] = []
-    public var frameworksBuildPhases: [PBXFrameworksBuildPhase] = []
-    public var headersBuildPhases: [PBXHeadersBuildPhase] = []
-    public var nativeTargets: [PBXNativeTarget] = []
-    public var fileReferences: [String: PBXFileReference] = [:]
-    public var projects: [PBXProject] = []
-    public var referenceProxies: [PBXReferenceProxy] = []
+
+    public var aggregateTargets: ReferenceableCollection<PBXAggregateTarget> = ReferenceableCollection<PBXAggregateTarget>()
```
pepibumur: Nice @rahul-malik! Looking forward to having this in the next release 😛

Sources/xcproj/PBXProj.swift
```diff
@@ -18,34 +17,38 @@ public class PBXProj: Decodable {
...
-    public var targetDependencies: [PBXTargetDependency] = []
-    public var sourcesBuildPhases: [PBXSourcesBuildPhase] = []
-    public var shellScriptBuildPhases: [PBXShellScriptBuildPhase] = []
-    public var resourcesBuildPhases: [PBXResourcesBuildPhase] = []
-    public var frameworksBuildPhases: [PBXFrameworksBuildPhase] = []
-    public var headersBuildPhases: [PBXHeadersBuildPhase] = []
-    public var nativeTargets: [PBXNativeTarget] = []
-    public var fileReferences: [PBXFileReference] = []
-    public var projects: [PBXProject] = []
-    public var referenceProxies: [PBXReferenceProxy] = []
```
pepibumur: With those changes, we'd need to bump the major version again because those are breaking changes. What do you think about having something like this:

```swift
@available(*, deprecated, message: "Use referenceProxiesCollection instead") public var referenceProxies: [PBXReferenceProxy] {
  return referenceProxiesCollection.values
}
public var referenceProxiesCollection: ReferenceableCollection<PBXReferenceProxy> = ReferenceableCollection<PBXReferenceProxy>()
```
I'm not quite happy about the name though. I think we could avoid having to append `Collection` by using a class a a namespace:

```swift
class PBXProj {
  class PBXProjObjects {
    public var referenceProxies: ReferenceableCollection<PBXReferenceProxy> = ReferenceableCollection<PBXReferenceProxy>()
  }
  var objects: PBXProjObjects = PBXProjObjects()
}
```
You could access them as `proj.objects. referenceProxies` instead of `proj. referenceProxies` that would be deprecated.

What are your thoughts on this @rahul-malik  / @yonaskolb ?

yonaskolb: Yeah, I quite like the idea of encapsulating all the collections into one objects class. Then it keeps all those huge switch statements in one place. 
As above though I still think those actual collections should just be dictionaries instead of a `ReferenceableCollection`

Sources/xcproj/Referenceable.swift
yonaskolb: @rahul-malik yeah we need to support removing as well, as that comes as part of modifying a project. I understand now you wanted to restrict that, hence the encapsulation

yonaskolb:
I just thought of something else. If elements are stored by reference, then we need to make `PBXObject.reference` a let otherwise it could change out from under us. This would restrict people from modifying a reference, but I don't think anyone would want to do that anyway?

yonaskolb:
Also if we do change references to be generated automatically on writing as discussed in #32, the public api would probably change yet again. I've opened PR #153 that handles the caching internally while we decide on a final public API

pepibumur:
I didn't think about that @yonaskolb. I'd make it let and prevent people from updating the value. Developers might not know the implications of doing such modification and get unexpected results using xcproj.  By making it `let` we'd be forcing them to create a new copy of the objects if they want to change the reference, removing the old one from the `PBXProj` and adding the new one. It's a bit cumbersome but safe.

I had another idea, what do you think about not having a reference in the objects? The reference would be the key that references the object in the objects' dictionary. 
- We create the dictionary when we parse the project, so all the objects have their respective reference.
- If we add a new object, it's up to us to specify its reference when adding it to the dictionary.
- Objects could return a deterministic hash that is calculated based on its attributes and that is used with the list of existing references to calculate the reference that it'd have in the project:

```swift
protocol ObjectHashable {
  var hash: String
}
class PBXObject: ObjectHashable {
  // Each object would implement the hash
}
class PBXProj {
  var objects: [String: PBXObject]
  func add(object: PBXObject) -> String // Returns the generated reference
}
```
What do you think @rahul-malik / @yonaskolb ?

Sources/xcproj/PBXProj.swift
```diff
@@ -20,36 +145,28 @@ public class PBXProj: Decodable {
...
-        var phases: [PBXBuildPhase] = []
-        phases.append(contentsOf: self.copyFilesBuildPhases as [PBXBuildPhase])
-        phases.append(contentsOf: self.sourcesBuildPhases as [PBXBuildPhase])
-        phases.append(contentsOf: self.shellScriptBuildPhases as [PBXBuildPhase])
-        phases.append(contentsOf: self.resourcesBuildPhases as [PBXBuildPhase])
-        phases.append(contentsOf: self.frameworksBuildPhases as [PBXBuildPhase])
-        phases.append(contentsOf: self.headersBuildPhases as [PBXBuildPhase])
-        return phases
-    }
+    @available(*, deprecated, message: "Use objects.buildFiles instead") public var buildFiles: [PBXBuildFile] { get { return objects.buildFiles.referenceValues } set(v) { objects.buildFiles = Dictionary.from(references: v) } }
```
yonaskolb: We could move these into a `Deprecations.swift` file that has an extension on PBXProj

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -41,15 +41,15 @@ extension PBXProj {
     /// - Parameter reference: reference of the file whose type will be returned.
     /// - Returns: String with the type of file.
     func buildPhaseType(buildFileReference: String) -> BuildPhase? {
-        if sourcesBuildPhases.contains(where: {$0.files.contains(buildFileReference)}) {
+        if objects.sourcesBuildPhases.contains(where: { k,v in v.files.contains(buildFileReference)}) {
```
yonaskolb: Nitpick: there's probably going to be a warning about not using `k`. It could either be `_, v` or what about just
```
if objects.sourcesBuildPhases.values.contains(where: { $0.files.contains(buildFileReference)}) {
```

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,133 @@ import PathKit
 
 /// Represents a .pbxproj file
 public class PBXProj: Decodable {
+    public class PBXProjObjects: Equatable {
```
yonaskolb: `Objects`?

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,133 @@ import PathKit
...
+        public var nativeTargets: ReferenceableCache<PBXNativeTarget> = ReferenceableCache<PBXNativeTarget>()
+        public var fileReferences: ReferenceableCache<PBXFileReference> = ReferenceableCache<PBXFileReference>()
+        public var projects: ReferenceableCache<PBXProject> = ReferenceableCache<PBXProject>()
+        public var referenceProxies: ReferenceableCache<PBXReferenceProxy> = ReferenceableCache<PBXReferenceProxy>()
+
+        // Open Questions about buildPhases
+        // - Unsure if there is value in making this a dictionary
+        // - Will the ordering of the phases be unstable given we are retrieving values from a dictionary
+        // and this might require a sort of some kind (compare references strings?)
+        // - Should we convert the build phases objects back to arrays
```
yonaskolb: For what it's worth, I don't use this property from XcodeGen. I can't see how the order of these values could matter. It's not the order they are defined in the project. That order is defined on each target. 
I think this could be typed as [String: PBXBuildPhase]

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,133 @@ import PathKit
...
+            case let object as PBXNativeTarget: nativeTargets.append(object)
+            case let object as PBXFileReference: fileReferences.append(object)
+            case let object as PBXProject: projects.append(object)
+            case let object as XCVersionGroup: versionGroups.append(object)
+            case let object as PBXReferenceProxy: referenceProxies.append(object)
+            default: fatalError("Unhandled PBXObject type for \(object), this is likely a bug / todo")
+            }
+        }
+
+        public func contains(reference: String) -> Bool {
```
yonaskolb: It would be useful, at least for XcodeGen to have a `getObject(reference: String) -> PBXObject` as well. This function could use that and just do a `!= nil`

Sources/xcproj/PBXProjEncoder.swift
```diff
@@ -19,8 +19,8 @@ class PBXProjEncoder {
     
     func encode(proj: PBXProj) -> String {
         // populate reference cache
-        proj.referenceCache.removeAll()
-        proj.objects.forEach { proj.referenceCache[$0.reference] = $0 }
+//        proj.referenceCache.removeAll()
```
yonaskolb: You can just remove all this `referenceCache` code

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,112 @@
...
+//  xcprojPackageDescription
+//
+//  Created by rmalik on 11/8/17.
+//
+
+import Foundation
+
+// Typealias that represents dictionary from string -> T where T is Referenceable (i.e. PBXObjects)
+
+public typealias ReferenceableCache<T: Referenceable & Equatable> = [String: T]
```
yonaskolb: It's not just a cache it's our internal data store now. So maybe another name? `ReferenceCollection`?

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,112 @@
...
+
+public protocol Referenceable {
+    var reference: String { get }
+}
+
+extension Array where Element: Referenceable {
```
yonaskolb: Are these still used anywhere? I guess for backwards compatibility 👍

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public func getReference(_ reference: String) -> Element? {
+        return first { $0.reference == reference }
+    }
+}
+
+
+extension Dictionary where Key == String, Value: Referenceable {
+    static func from(references: [Value]) -> Dictionary<String, Value> {
```
yonaskolb: init instead?

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,112 @@
...
+        return Dictionary<String, Value>(uniqueKeysWithValues: references.flatMap { ($0.reference, $0) })
+    }
+
+    public var references: [String] {
+        return self.values.map { $0.reference }
```
yonaskolb: Isn't this just self.keys? This again brings up the possible conflict between the key of the dictionary and the reference property. The reference property should be made a let.

The recent idea of removing the reference property entirely is an interesting one too. The use cases would be to be thought about, but maybe that's a good move

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public var referenceValues: [Value] {
```
yonaskolb: Is there much use in this?

Sources/xcproj/Referenceable.swift
```diff
@@ -0,0 +1,112 @@
...
+    public func getReference(_ reference: Key) -> Value? {
+        return self[reference]
+    }
+
+    mutating public func append(_ value: Value) {
+        self[value.reference] = value
+    }
+}
+
+//public struct ReferenceableCollection<T: Referenceable & Equatable>: Equatable, ExpressibleByArrayLiteral {
```
yonaskolb: This can be deleted

Sources/xcproj/XCConfigurationList.swift
```diff
@@ -79,8 +79,8 @@ extension XCConfigurationList: PlistSerializable {
     }
     
     private func plistComment(proj: PBXProj) -> String? {
-        let project = proj.projects.filter { $0.buildConfigurationList == self.reference }.first
-        let target = proj.nativeTargets.filter { $0.buildConfigurationList == self.reference }.first
+        let project = proj.objects.projects.filter { k,v in v.buildConfigurationList == self.reference }.first?.value
```
yonaskolb: This remind me this should just be using `first` instead of filter

Sources/xcproj/PBXProj+Helpers.swift
yonaskolb: Same for all below this in different functions

Sources/xcproj/Referenceable.swift
rahul-malik: It's convenient instead of writing the return statement everywhere we need the list of values

Sources/xcproj/Referenceable.swift
rahul-malik: Yep, backwards compatibility

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.4.0 - (next version)
 
+- Improve efficiency of looking up `PBXObject`'s from `PBXProj` by @rmalik
```
pepibumur: Nice @rahul-malik

Sources/xcproj/Deprecations.swift
```diff
@@ -0,0 +1,35 @@
+//
```
pepibumur: Don't forget to remove the headers @rahul-malik! I'll add a Danger task to chek that automatically.

Sources/xcproj/Deprecations.swift
```diff
@@ -0,0 +1,35 @@
...
+//  xcprojPackageDescription
+//
+//  Created by rmalik on 11/15/17.
+//
+
+import Foundation
+
+extension PBXProj {
+
+    @available(*, deprecated, message: "Use objects.buildFiles instead") public var buildFiles: [PBXBuildFile] { get { return objects.buildFiles.referenceValues } set(v) { objects.buildFiles = Dictionary(references: v) } }
```
pepibumur: This is great @rahul-malik 😛

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,135 @@ import PathKit
...
+        public var variantGroups: ReferenceableCollection<PBXVariantGroup> = ReferenceableCollection<PBXVariantGroup>()
+        public var targetDependencies: ReferenceableCollection<PBXTargetDependency> = ReferenceableCollection<PBXTargetDependency>()
+
+        public var nativeTargets: ReferenceableCollection<PBXNativeTarget> = ReferenceableCollection<PBXNativeTarget>()
+        public var fileReferences: ReferenceableCollection<PBXFileReference> = ReferenceableCollection<PBXFileReference>()
+        public var projects: ReferenceableCollection<PBXProject> = ReferenceableCollection<PBXProject>()
+        public var referenceProxies: ReferenceableCollection<PBXReferenceProxy> = ReferenceableCollection<PBXReferenceProxy>()
+
+
+        // Ordering matters with build phases so leave these as arrays
```
pepibumur: @rahul-malik I think I'm missing something here. Where does the order matter?

Sources/xcproj/PBXProj.swift
pepibumur: Yeah, unless I miss something I don't think the order matters since the order is specified in the target that references them. I think we can keep them as dictionaries.
I use the property from `xclint` but it seems that it's very concrete to the library use case. Feel free to deprecate it since I can extend the `PBXProj` from there.

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,135 @@ import PathKit
...
+                buildConfigurations,
+                variantGroups,
+                targetDependencies,
+                nativeTargets,
+                fileReferences,
+                projects,
+                versionGroups,
+                referenceProxies
+            ]
+            return caches.first { cache in cache[reference] != nil }?[reference] as? T ?? self.buildPhases.first { $0.reference == reference } as? T ?? nil
```
pepibumur: After moving the buildPhases to dictionaries you'll be able to simplify this line

Sources/xcproj/PBXGroup.swift
```diff
@@ -106,11 +106,11 @@ extension PBXGroup: PlistSerializable {
     }
 
     fileprivate func name(reference: String, proj: PBXProj) -> String? {
-        if let group: PBXGroup = proj.getCachedReference(reference) {
+        if let group: PBXGroup = proj.objects.getReference(reference) {
```
yonaskolb: These can just use the normal `objects.groups`. It will be much more efficient

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -81,17 +81,17 @@ extension PBXProj {
...
+        if objects.sourcesBuildPhases.contains(reference: buildPhaseReference) {
             return "Sources"
-        } else if frameworksBuildPhases.contains(reference: buildPhaseReference) {
+        } else if objects.frameworksBuildPhases.contains(reference: buildPhaseReference) {
             return "Frameworks"
-        } else if resourcesBuildPhases.contains(reference: buildPhaseReference) {
+        } else if objects.resourcesBuildPhases.contains(reference: buildPhaseReference) {
             return "Resources"
-        } else if let copyFilesBuildPhase = copyFilesBuildPhases.first(where: {$0.reference == buildPhaseReference}) {
+        } else if let copyFilesBuildPhase = objects.copyFilesBuildPhases.first(where: { $0.reference == buildPhaseReference}) {
```
yonaskolb: These can use the more performant `getReference`

Sources/xcproj/PBXProj.swift
yonaskolb: Yeah these can also be ReferenceableCollection

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,135 @@ import PathKit
...
+            case let object as PBXNativeTarget: nativeTargets.append(object)
+            case let object as PBXFileReference: fileReferences.append(object)
+            case let object as PBXProject: projects.append(object)
+            case let object as XCVersionGroup: versionGroups.append(object)
+            case let object as PBXReferenceProxy: referenceProxies.append(object)
+            default: fatalError("Unhandled PBXObject type for \(object), this is likely a bug / todo")
+            }
+        }
+
+        public func getReference<T: PBXObject>(_ reference: String) -> T? {
```
yonaskolb: This doesn't need to be generic. The other previous one was as the old temporary cache didn't have other explicit dictionaries to access. If someone knew which type they wanted they can now just use one of the explicit dictionaries. (which also doesn't have to loop through every collection)
This can just return a `PBXObject?`

Sources/xcproj/PBXProject.swift
```diff
@@ -184,7 +184,7 @@ extension PBXProject: PlistSerializable {
         dictionary["projectRoot"] = .string(CommentedString(projectRoot))
         dictionary["targets"] = PlistValue.array(targets
             .map { target in
-                let targetName = (proj.getCachedReference(target) as? PBXTarget)?.name
+                let targetName = (proj.objects.getReference(target) as? PBXTarget)?.name
```
yonaskolb: `.nativeTargets`

Sources/xcproj/PBXTargetDependency.swift
```diff
@@ -60,7 +60,7 @@ extension PBXTargetDependency: PlistSerializable {
         var dictionary: [CommentedString: PlistValue] = [:]
         dictionary["isa"] = .string(CommentedString(PBXTargetDependency.isa))
         if let target = target {
-            let targetName = (proj.getCachedReference(target) as? PBXTarget)?.name
+            let targetName = (proj.objects.getReference(target) as? PBXTarget)?.name
```
yonaskolb: `.nativeTargets`

Sources/xcproj/PBXProj.swift
rahul-malik: Ok I'll change these to dictionaries. I had a test failure in XCodeGen that I assumed was due to ordering but turned out the test just needed to be updated.

Sources/xcproj/PBXGroup.swift
rahul-malik: Good catch! I was updating references to get everything building but didn't reevaluate callers of the general getReference func

Sources/xcproj/PBXProject.swift
rahul-malik: @yonaskolb - What about aggregate targets?

Sources/xcproj/PBXTargetDependency.swift
rahul-malik: @yonaskolb - What about aggregate targets?

Sources/xcproj/PBXProject.swift
yonaskolb: True. That means the old implementation was also missing this. Maybe we need a getTarget(reference: String) -> PBXTarget

Sources/xcproj/PBXProj.swift
```diff
@@ -3,8 +3,127 @@ import PathKit
...
+                fileElements,
+                configurationLists,
+                buildConfigurations,
+                variantGroups,
+                targetDependencies,
+                nativeTargets,
+                fileReferences,
+                projects,
+                versionGroups,
+                referenceProxies
```
yonaskolb: Add the buildPhases

Sources/xcproj/PBXProject.swift
yonaskolb: I think it's fine as it is though. This isn't a heavily trafficked line of code, so the performance tradeoff isn't that bad

rahul-malik: Adding `getTarget(reference: String)` since I thought the same thing :)

yonaskolb:
If the tests pass, this is good to merge 👍

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -36,7 +36,7 @@ extension PBXProj {
     /// - Parameter configReference: reference of the XCBuildConfiguration.
     /// - Returns: config name.
     func configName(configReference: String) -> String? {
-        return (objects.getReference(configReference) as? XCBuildConfiguration)?.name
+        return objects.buildConfigurations[configReference]?.name
```
yonaskolb: 👍  
You've probably done it already, but a search for any of these that are left would be good.

yonaskolb:
@rahul-malik One more thing. Can we make `PBXObject.reference` a let so that it won't get out of sync with the dictionaries

yonaskolb:
This is failing on CI because the Carthage project needs to be regenerated to add the new Referenceable.swift file
```
bundle exec rake generate_carthage_project
```

PR #150, opened 2017-11-09
Fix comments for build phases

allu22:
Resolves https://github.com/xcodeswift/xcproj/issues/138 and https://github.com/xcodeswift/xcproj/issues/139

### Short description 📝
Currently phase type is used as comment for PBXBuildFile and PBXBuildPhase files.

### Solution 📦
Use build phase name for comments, if present, instead of phase type.

### GIF
gif

### DIFF
--- a/Sources/xcproj/PBXProj+Helpers.swift
+++ b/Sources/xcproj/PBXProj+Helpers.swift
@@ -90,7 +90,26 @@ extension PBXProj {
         }
         return nil
     }
-    
+
+    /// Returns the build phase name a file is in (mostly used for comments).
+    ///
+    /// - Parameter reference: reference of the file whose type name will be returned.
+    /// - Returns: the build phase name.
+    func buildPhaseName(buildFileReference: String) -> String? {
+        if sourcesBuildPhases.filter({$0.files.contains(buildFileReference)}).count != 0 {
+            return BuildPhase.sources.rawValue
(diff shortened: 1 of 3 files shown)


Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -90,7 +90,26 @@ extension PBXProj {
...
         return nil
     }
-    
+
+    /// Returns the build phase name a file is in (mostly used for comments).
+    ///
+    /// - Parameter reference: reference of the file whose type name will be returned.
+    /// - Returns: the build phase name.
+    func buildPhaseName(buildFileReference: String) -> String? {
+        if sourcesBuildPhases.filter({$0.files.contains(buildFileReference)}).count != 0 {
```
yonaskolb: I think this whole function could just return `buildPhaseType(buildFileReference: buildFileReference)?.rawValue`
Also both this and buildPhaseType can make use of `contains(..)` instead of filtering and then counting. It's more performant as if a match is found early it will return straight away, and we're currently trying to resolve performance issues

pepibumur: Adding the improvements. Good catch!

pepibumur: @yonaskolb https://github.com/xcodeswift/xcproj/commit/ae5aacc98564ea8e35423d632485b6a80801061d

PR #151, opened 2017-11-09
Fix PBXTarget product reference comment

allu22:
Resolves https://github.com/xcodeswift/xcproj/issues/143

### Short description 📝
PBXNativeTarget productReference is missing comment.

### Solution 📦
The comment was missing because the file name was searched from build files but should have been searched from file references.


pepibumur:
Nice @allu22

PR #152, opened 2017-11-09
Add PBXProject name to comments

allu22:
Resolves https://github.com/xcodeswift/xcproj/issues/144

### Short description 📝
Adds PBXProject name to comments.

### Solution 📦
As PBXProject name is determined by its filename:
- When decoding from path, use it to find project name and assign it to root project.
- When encoding to path, use the path to overwrite the project name.

### Alternatives considered
We could use regex to determine current project name:
```` swift
let prefix = "Build configuration list for PBXProject \""
let suffix = "\""

let projectName = String(data: data, encoding: .utf8).flatMap { project in
    project.range(of: "\(prefix).*\(suffix)", options: .regularExpression)
        .map({ range in
            String(project[range]
                .dropFirst(prefix.count)
                .dropLast(suffix.count)
            )
        })
}
````
This would help in situations where file name is not available, for example: when parsing from `Data`. But as `XcodeProj` does not at the moment directly support initialising with `Data`, I believe using file name approach is better.

### GIF
gif

### DIFF
--- a/Sources/xcproj/PBXProj+Helpers.swift
+++ b/Sources/xcproj/PBXProj+Helpers.swift
@@ -111,6 +111,21 @@ extension PBXProj {
         }
     }
 
+    /// Infers project name from Path and sets it as project name
+    ///
+    /// Project name is needed for certain comments when serialising PBXProj
+    ///
+    /// - Parameters:
+    ///   - path: path to .xcodeproj directory.
+    func updateProjectName(path: Path) {
+        guard path.parent().extension == "xcodeproj" else {
+            return
+        }
+        let projectName = path.parent().lastComponentWithoutExtension
+        let rootProject = projects.first(where: { $0.reference == rootObject })
+        rootProject?.name = projectName
+    }
+
 }
 
 // MARK: - PBXProj extension (Writable)
--- a/Sources/xcproj/XcodeProj.swift
+++ b/Sources/xcproj/XcodeProj.swift
@@ -26,6 +26,7 @@ public class XcodeProj {
         let pbxProjData = try Data(contentsOf: pbxprojPaths.first!.url)
         let plistDecoder = PropertyListDecoder()
         pbxproj = try plistDecoder.decode(PBXProj.self, from: pbxProjData)
+        pbxproj.updateProjectName(path: pbxprojPaths.first!)
         let xcworkspacePaths = path.glob("*.xcworkspace")
         if xcworkspacePaths.count == 0 {
             workspace = XCWorkspace()
--- a/Tests/xcprojTests/XcodeProjIntegrationSpec.swift
+++ b/Tests/xcprojTests/XcodeProjIntegrationSpec.swift
@@ -28,6 +28,13 @@ final class XcodeProjIntegrationSpec: XCTestCase {
         XCTAssertNotNil(got)
     }
 
+    func test_init_setsCorrectProjectName() {
+        let proj = projectiOS()!.pbxproj
+        let rootObject = proj.rootObject
+        let rootProject = proj.projects.first(where: { $0.reference == rootObject })
+        XCTAssertEqual(rootProject?.name, "Project")
+    }
+
     // MARK: - Private
 
     private func fixtureWithoutWorkspaceProjectPath() -> Path {

### CODE REVIEW

File: Sources/xcproj/Path+Helpers.swift [hunk 1]
```diff
@@ -0,0 +1,18 @@
+import PathKit
+
+// MARK: - Path Extension (Extras)
+
+extension Path {
+
+    /// - Returns: Base name for xcodeproj directory
+    func projectName() -> String? {
```
pepibumur: I'd add a unit test here.

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -98,6 +98,10 @@ extension PBXProj {
 extension PBXProj: Writable {
     
     public func write(path: Path, override: Bool) throws {
+        if let projectName = path.projectName() {
```
pepibumur: I don't think we need this here. If we already set the name when we read the project, that should be enough. Moreover, I don't think this method would work because the path that we are passing here is the `.pbxproj` path, not the project path, so the name that you would get here is `project` which is the default `pbxproj` name.

Sources/xcproj/XcodeProj.swift
```diff
@@ -26,6 +26,11 @@ public class XcodeProj {
         let pbxProjData = try Data(contentsOf: pbxprojPaths.first!.url)
         let plistDecoder = PropertyListDecoder()
         pbxproj = try plistDecoder.decode(PBXProj.self, from: pbxProjData)
+        if let projectName = pbxprojPaths.first!.projectName() {
```
pepibumur: It might be confusing for someone looking at this code to know why this is necessary. What do you think about extracting this into a more descriptive method:

```swift
static class func updateProjectName(path: Path, pbxproj: PBXProj) {
 // Your logic
}
```
And maybe adding a comment on the method explaining that the attribute might be missing and it's needed to generate certain comments.

Sources/xcproj/PBXProj+Helpers.swift
allu22: I believe this part is actually more important than reading. When we write project to Path, the name of the project will come form the directory where `pbxproj` file is in, not from the `PBXProj` attribute `name`.

For examlpe. When you load load project from `Project.xcodeproj/project.pbxproj` and write it to `MyProject.xcodeproj/project.pbxproj` then the comments should be for `MyProject`. Otherwise when you resave it with Xcode, you will get an unwanted diff.

Sources/xcproj/Path+Helpers.swift
allu22: Done.

Sources/xcproj/XcodeProj.swift
allu22: I moved it to `Sources/xcproj/PBXProj+Helpers.swift` as helper method.

Sources/xcproj/PBXProj+Helpers.swift
pepibumur: I agree, then I'd keep it in only one place, the writing. What do you think? If I'm not wrong, I think it should be `path.parent.projectName()`, shouldn't it?

allu22: I changed getting project name from `Path` to `path.parent.projectName()` but kept it in internal extension, as this allows us to cover it with unit test and this will describe the needs/limitations of it.

In addition to writing the project name, we should read it also. It will come in handy when you want to serialise the project to Data, after reading it from file. In this case the original project name is kept in comments.

allu22:
@pepibumur please review this again. My last feedback: https://github.com/xcodeswift/xcproj/pull/152#discussion_r150310722

Sources/xcproj/Path+Helpers.swift
```diff
@@ -0,0 +1,16 @@
+import PathKit
+
+// MARK: - Path Extension (Extras)
+
+extension Path {
+
+    /// - Returns: Base name for xcodeproj directory
+    internal func projectName() -> String? {
```
pepibumur: I'm not sure about having this method. This method is thought to be used with `.pbxproj` paths, and that's not explicit in the definition.  Since it's just calling `lastComponentWithoutExtension` if it's an xcode project I'd move that line to the places where we call this method.

Sources/xcproj/XcodeProj.swift
```diff
@@ -26,6 +26,7 @@ public class XcodeProj {
         let pbxProjData = try Data(contentsOf: pbxprojPaths.first!.url)
         let plistDecoder = PropertyListDecoder()
         pbxproj = try plistDecoder.decode(PBXProj.self, from: pbxProjData)
+        pbxproj.updateProjectName(path: pbxprojPaths.first!)
```
pepibumur:
```swift
pbxproj.updateProjectName(path: pbxprojPaths.first!.parent().lastComponentWithoutExtension)
```

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -111,13 +111,28 @@ extension PBXProj {
         }
     }
 
+    /// Infers project name from Path and sets it as project name
+    ///
+    /// Project name is needed for certain comments when serialising PBXProj
+    ///
+    /// - Parameters:
+    ///   - path: path to .xcodeproj directory.
+    func updateProjectName(path: Path) {
```
pepibumur: 👏 Nice @allu22

Sources/xcproj/Path+Helpers.swift
allu22: I'm starting to think that we should not automatically set project name during writing. I would guess, it is not very common use case, where someone wan't so save project with new name. Considering this, maybe it should be users responsibility to set new project name, before serialising project.

What if I move
```swift
if parent().extension == "xcodeproj" {
  return parent().lastComponentWithoutExtension
} else {
  return nil
}
```
into https://github.com/xcodeswift/xcproj/pull/152/files/15bd83a1794242f4ddc37d3ce0880c9c7207b914#diff-682f894948ca1d5648a6d9bd7593c396R120
Turning it into:
```swift
func updateProjectName(path: Path) {
  guard path.parent().extension == "xcodeproj" else {
    return
  }
  let projectName = path.parent().lastComponentWithoutExtension
  let rootProject = projects.first(where: { $0.reference == rootObject })
  rootProject?.name = projectName
}
```

And add tests to check if `public init(path: Path) throws` updates project name.

allu22: @pepibumur thoughts?

pepibumur: Good point. I don't think people would be renaming the project once created, and if they do it's up to them to change the name accordingly. I think moving the updateProjectName in the init and testing it is a good idea. Go for it @allu22

allu22: @pepibumur i've made these changes. Would live if you could give one more look :)

PR #153, opened 2017-11-09
Cache pbxobject references during project writing

yonaskolb:
This is an alternative to the reference changes in #136.

This doesn't change the public API and simply generates an internal cache of reference that can be looked up while writing, improving performance.
It's a temporary internal solution while we lock down the final public API for pbx objects.

At the moment it's a generic function that requires a type. This could be changed to seperate typed caches for all the required types.

### DIFF
--- a/Sources/xcproj/PBXBuildFile.swift
+++ b/Sources/xcproj/PBXBuildFile.swift
@@ -88,11 +88,9 @@ extension PBXBuildFile: PlistSerializable {
     }
 
     private func name(fileRef: String, proj: PBXProj) -> String? {
-        let fileReference = proj.fileReferences.getReference(fileRef)
-        let variantGroup = proj.variantGroups.getReference(fileRef)
-        if let fileReference = fileReference {
+        if let fileReference: PBXFileReference = proj.getCachedReference(fileRef) {
             return fileReference.name ?? fileReference.path
-        } else if let variantGroup = variantGroup {
+        } else if let variantGroup: PBXVariantGroup = proj.getCachedReference(fileRef) {
             return variantGroup.name
         }
         return nil
--- a/Sources/xcproj/PBXBuildFile.swift
+++ b/Sources/xcproj/PBXBuildFile.swift
@@ -73,7 +73,7 @@ extension PBXBuildFile: PlistSerializable {
         dictionary["isa"] = .string(CommentedString(PBXBuildFile.isa))
         var fileName: String?
         if let fileRef = fileRef {
-            fileName = name(fileRef: fileRef, proj: proj)
+            fileName = proj.fileName(fileReference: fileRef)
             dictionary["fileRef"] = .string(CommentedString(fileRef, comment: fileName))
         }
         if let settings = settings {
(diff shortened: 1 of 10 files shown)


Sources/xcproj/PBXProjEncoder.swift
```diff
@@ -18,6 +18,10 @@ class PBXProjEncoder {
     var multiline: Bool = true
     
     func encode(proj: PBXProj) -> String {
+        // populate reference cache
+        proj.referenceCache.removeAll()
```
pepibumur: I like the idea of having a cache to prevent having to look up for objects every time we need to generate a comment. However, I'm not 100% convinced of the implementation. I think the caching is something we need to speed up writing, thus it shouldn't be part of the project. I think there should be an instance that we instantiate at the moment we start encoding the project, sort of an `ObjectsProvider` that we pass to other objects instead of the `PBXProj` to generate the comments:

```swift
protocol ObjectsProviding {
  func get(reference: String) -> PBXObject
}
class ObjectsProvider: ObjectsProviding {
  let proj: PBXProj
  func get(reference: String) -> PBXObject {}
}
```
And then at then, we could pass that provider to the objects:

```swift
let element = serializable.plistKeyAndValue(provider: ObjectsProviding)
```

Eventually, when we move from arrays to string, we could extend `PBXProj` to conform the `ObjectsProviding` protocol, and pass the project instead. What do you think @yonaskolb?

Sources/xcproj/PBXBuildFile.swift
```diff
@@ -88,11 +88,9 @@ extension PBXBuildFile: PlistSerializable {
     }
 
     private func name(fileRef: String, proj: PBXProj) -> String? {
-        let fileReference = proj.fileReferences.getReference(fileRef)
```
pepibumur: I didn't know we still had some objects with methods to get the names for the comments. Would you mind using the methods from `PBXProj+Helpers.swift`? I can also do it on another PR.

pepibumur:
Just a couple of comments, great idea @yonaskolb. I'd check the memory consumption just in case.

Sources/xcproj/PBXProjEncoder.swift
yonaskolb: Yeah, sounds a fine approach. I just didn't want to over-engineer this as it's designed as mostly a temporary hotfix

Sources/xcproj/PBXBuildFile.swift
yonaskolb: Yeah, I noticed that there are also functions in PBXProj `getName` and `getConfig`. `getName` actually mirrors the one in Helpers but has a slightly different implementation. I've removed these now

Sources/xcproj/PBXProjEncoder.swift
pepibumur: Good point, then feel free to leave it like that. What do you think about opening an GitHub issue to tackle this once #136 has been merged? Great job @yonaskolb

PR #156, opened 2017-11-09
Add buildPhase property to PBXBuildPhase

yonaskolb:
This adds the build phase enum as a property of each PBXBuildPhase subclass This makes it easier to work with the polymorphic PBXProj.buildPhases


pepibumur:
Great @yonaskolb 👏

PR #158, opened 2017-11-11
Danger

pepibumur:
### Short description 📝
Add Danger


pepibumur:
For maintainability reasons, we'll move the `Danger` configuration into an organization repository that can be shared across all the repositories. https://github.com/xcodeswift/danger

Issue #145, opened 2017-11-08
Build settings generated without quotation marks


pepibumur:
## Context 🕵️‍♀️
Xcode generates the build settings that are strings adding quotation marks. When xcproj saves them, it doesn't add those quotation marks.

## What 🌱
![image](https://user-images.githubusercontent.com/663605/32538998-5b376894-c468-11e7-9477-57ecc08b02a7.png)

## Proposal 🎉
Update how the build settings are generated to include those quotation marks.

yonaskolb:
I believe it only adds them if they contain special characters. And I think `/` is one of them. `$` may be too

pepibumur:
Yeah we can have a list of those special characters and expand it as we find more cases.

yonaskolb:
This CharacterSet should cover it

PR #132, opened 2017-10-27
Add buildPhases property to PBXProj

pepibumur:
### Short description 📝
I've added a convenient getter to get all the build phases.

### Solution 📦
Add the getter that aggregates all the build phases.

### Implementation 👩‍💻👨‍💻
- [x] Implement it.
- [x] Test it.

### GIF
gif

PR #159, opened 2017-11-13
Fix CommentedString escaped string transformation

allu22:
Resolves https://github.com/xcodeswift/xcproj/issues/145, https://github.com/xcodeswift/xcproj/issues/71

### Short description 📝
Currently, the logic and spec for transforming quotation mark is incorrect. This was witnessed by #145. This can be validated by re-saving a project with Xcode, after you have added quotation marks to build script.

### Solution 📦
Fixed quotation mark escaping. After that it was possible to use CommentedString with `PBXTarget productType` and `PlistValue key`.

Using this fix and #132, I am able to deserialise and serialise (simple new) project without generating any diff.

### GIF
gif

(PS! It is horrible to debug quotation marks 😭)

### DIFF
--- a/Sources/xcproj/CommentedString.swift
+++ b/Sources/xcproj/CommentedString.swift
@@ -31,15 +31,14 @@ struct CommentedString {
         }
 
         var escaped = string
+        // escape escape
+        escaped = escaped.replacingOccurrences(of: "\\", with: "\\\\")
+        // escape quotes
+        escaped = escaped.replacingOccurrences(of: "\"", with: "\\\"")
+        // escape tab
+        escaped = escaped.replacingOccurrences(of: "\t", with: "\\t")
         // escape newlines
         escaped = escaped.replacingOccurrences(of: "\n", with: "\\n")
-
-        // escape quotes
-        var range: Range<String.Index>?
-        if escaped.isQuoted && escaped.count > 1 {
-            range = escaped.index(after: escaped.startIndex)..<escaped.index(before: escaped.endIndex)
-        }
-        escaped = escaped.replacingOccurrences(of: "\"", with: "\\\"", range: range).replacingOccurrences(of: "\\\\\"", with: "\\\"")
 
         if !escaped.isQuoted && escaped.rangeOfCharacter(from: CommentedString.invalidCharacters) != nil {
             escaped = escaped.quoted
--- a/Tests/xcprojTests/PBXProjWriterSpec.swift
+++ b/Tests/xcprojTests/PBXProjWriterSpec.swift
@@ -1,5 +1,6 @@
 import Foundation
 import XCTest
+import PathKit
 
 @testable import xcproj
 
@@ -36,10 +37,12 @@ class PBXProjEncoderSpec: XCTestCase {
         let quote = "\""
         let escapedNewline = "\\n"
         let escapedQuote = "\\\""
+        let escapedEscape = "\\\\"
+        let escapedTab = "\\t"
 
         let values: [String: String] = [
             "a": "a",
-            "a".quoted: "a".quoted,
+            "a".quoted: "\(escapedQuote)a\(escapedQuote)".quoted,
             "@": "@".quoted,
             "[": "[".quoted,
             "<": "<".quoted,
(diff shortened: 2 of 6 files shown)


allu22:
@pepibumur Can I press Merge or should we wait for another reviewer?

Sources/xcproj/CommentedString.swift
```diff
@@ -31,16 +31,15 @@ struct CommentedString {
         }
 
         var escaped = string
+        // escape escape
+        escaped = escaped.replacingOccurrences(of: "\\", with: "\\\\")
```
yonaskolb: Won’t this now add more and more escaping characters every time a project is written?

Tests/xcprojTests/PBXProjWriterSpec.swift
```diff
@@ -36,10 +36,12 @@ class PBXProjEncoderSpec: XCTestCase {
         let quote = "\""
         let escapedNewline = "\\n"
         let escapedQuote = "\\\""
+        let escapedEscape = "\\\\"
+        let escapedTab = "\\t"
 
         let values: [String: String] = [
             "a": "a",
-            "a".quoted: "a".quoted,
+            "a".quoted: "\(escapedQuote)a\(escapedQuote)".quoted,
```
yonaskolb: Like the below cases this is escaping the surrounding quotes when it doesn’t need to. This would cause even more diffs when a project with any quoted values is opened and then saved, as this would happen every time

Tests/xcprojTests/PBXProjWriterSpec.swift
```diff
@@ -49,29 +51,30 @@ class PBXProjEncoderSpec: XCTestCase {
...
+            "a\(escapedQuote)a\(escapedQuote)": "a\(escapedEscape)\(escapedQuote)a\(escapedEscape)\(escapedQuote)".quoted,
             "a\na": "a\\na".quoted,
             "\n": escapedNewline.quoted,
             "\na": "\(escapedNewline)a".quoted,
             "a\n": "a\(escapedNewline)".quoted,
-            "a\na".quoted: "a\(escapedNewline)a".quoted,
-            "a\(escapedNewline)a": "a\(escapedNewline)a".quoted,
-            "a\(escapedNewline)a".quoted: "a\(escapedNewline)a".quoted,
+            "a\na".quoted: "\(escapedQuote)a\(escapedNewline)a\(escapedQuote)".quoted,
+            "a\(escapedNewline)a": "a\(escapedEscape)na".quoted,
```
yonaskolb: Here’s an example of an already escaped character getting escaped again

Sources/xcproj/CommentedString.swift
allu22: No, it wont. You can try it by adding build setting in Xcode with value \. It will become "\\" in proj file: 
https://github.com/RepositoryFork/xcproj/blob/d8a6995c23f8a5bd2f0aea68cb6461e7f08ee788/Fixtures/iOS/BuildSettings.xcodeproj/project.pbxproj#L103

Tests/xcprojTests/PBXProjWriterSpec.swift
allu22: This covers case where user has typed 
`a\na` 
not
```
a
a
```

Example: https://github.com/RepositoryFork/xcproj/blob/d8a6995c23f8a5bd2f0aea68cb6461e7f08ee788/Fixtures/iOS/BuildSettings.xcodeproj/project.pbxproj#L100

Tests/xcprojTests/PBXProjWriterSpec.swift
allu22: We need to escape quotes that have been inserted be user. This can bee seen in https://github.com/xcodeswift/xcproj/issues/71. It wont add additional quotes because the quotes that are meant for encoding are removed during decoding.

allu22:
@yonaskolb I added fixture to test the cases you pointed out. This fixture was saved in Xcode, so it should be a source of truth.

Here is how it looks like in Xcode:
img
And here is the raw format:
https://github.com/xcodeswift/xcproj/pull/159/files#diff-07993c7ba07ab356e6d2e793c253998a

Did I clear up your concerns?

yonaskolb:
Ok cool. Sounds like the encoding and decoding removes all the duplicate escaping characters and quotes. Great!

pepibumur:
👏

PR #160, opened 2017-11-14
Update Dangerfile to use the organization one

pepibumur:


yonaskolb:
Is the comment above supposed to look like that?

pepibumur:
Nop @yonaskolb. There was an issue with the Danger file. I'm re-running the job.

orta:
You could consider using [Danger Swift](https://github.com/danger/danger-swift) for a Dangerfile that simple? [Reference PR](https://github.com/Moya/Harvey/pull/13/files) ( use `npm install danger` not `npm install danger@alpha` though )

pepibumur:
Nice! I'll do that instead. Thanks @orta

pepibumur:
@orta just wondering. Can I use the `Danger.swift` definition reusing a `Dangerfile` define in another organization repository?

orta:
Nope, that'd require some interesting work, because you'd need to compile and evaluate those Swift files, which is tricky enough for just the original `Dangerfile.swift`. It can definitely be built into danger-swift though the DSL though.

PR #161, opened 2017-11-20
Use productRefGroup name as CommentedString comment

allu22:
### Short description 📝
At the moment `productRefGroup` comment is always `Products`. It probably covers 99.99% of use cases. But the comment should actually be a name of that group.
It will cause problems when you change the name of that group or don't have dedicated products group at all (In that case it falls back to main group).

### Solution 📦
Use productRefGroup name as CommentedString comment.

### GIF
gif

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 1.5.0 - (next release)
 
+### Fixed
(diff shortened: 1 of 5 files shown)


pepibumur:
It's ready to be merged @allu22

PR #164, opened 2017-11-21
Make final classes that are not extendible

pepibumur:
Partially resolves https://github.com/xcodeswift/xcproj/issues/163

### Short description 📝
Apple recommends you to make final those classes that are not extendible.

### Solution 📦
Review classes and make final those that are not supposed to be extended.

### Implementation 👩‍💻👨‍💻
- [x] Add `final` keyword to non-extendible classes

### GIF
gif

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 1.5.0 - (next release)
 
+### Added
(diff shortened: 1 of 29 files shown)


rahul-malik:
Awesome, thanks for the update!

PR #168, opened 2017-11-24
Integration testing

pepibumur:
### Short description 📝
Best way to test xcproj supports Xcode format is by running tests with real projects. The purpose of this PR is to create a new set of tests that run with real OSS projects.

### Solution 📦
Add integration tests.

### Implementation 👩‍💻👨‍💻
- [x] Create new tests 
- [x] Add some OSS projects to be tested.

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/168)


allu22:
I would like if we also tested if these project files will serialize without creating difference. But this may be a different PR.

But I don't like that the test is not reproducible. At the moment it takes the HEAD of the repos default branch every time you run it. It makes it very unpredictable and hard to track down why some test failed.

Sadly I don't have anything good to offer as replacement. Some options, that I can think of:
*  Checkout specific commit from the repo. Meaning we store that commits hash in our test.
*  Store all needed project files in our repo.

If you decide to keep it as is, then i recommend printing failing test repos current commits hash.

pepibumur:
Good points @allu22. This sort of tests come with flakiness because anything can change on their side and break the tests *(e.g. they rename the project)*. We could make them reproducible by pointing to a specific commit of the repo but by doing that we'd only test if our changes in `xcproj` introduce any regression. I think it's also interesting to run the test against the last version of the project. That way if any new version of Xcode introduces changes in the format we could detect it by getting tests failing. 
What do you think about improving the output and printing the commit hash when they fail and giving it a try? If we see that the flakiness is too high we can re-evaluate it.

allu22:
@pepibumur yeah, this will be a good start.

rahul-malik:
@pepibumur @allu22  - You're both right! :)

There is actually value in both approaches on CI (building off `HEAD` and specific commit). We do this a lot at Pinterest. 
- The specific commit is generally the test we care about and runs on every PR. 
- The PR that builds off of `HEAD` daily is used to test compatibility. If it is successful, automatically move the specific commit test to the latest commit on `HEAD`. If it fails, investigate the failure to determine what happened (project name changed, xcode version updates, etc.)

pepibumur:
I had an idea! What if after every merge into master, we branch off from `master` and run the integration tests there? We could even create an issue automatically on the repository if the integration tests failed. That way if the integration breaks it doesn't block developers from merging their changes into master. I have to check if this is feasible with travis-ci.

pepibumur:
@allu22 & @rahul-malik I've addressed your comments:
- We don't run those tests from master nor branches.
- If master succeeds on travis-ci it pushes the last version of master to a new remote branch `integration`.
- Travis-ci runs the integration tests on that branch.

PR #171, opened 2017-11-29
Add PBXLegacy

bkase:
### Short description 📝
This is models the "External Build Tool" target type (internally called
PBXLegacyTarget in Xcode). All fields are optional (Xcode doesn't crash
if you omit them all). The build for this target will always fail if
 you omit the `buildToolPath`, however.

### Solution 📦
Since this is another specialization of PBXTarget it is a subclass of PBXTarget just like PBXNativeTarget. Unlike PBXNativeTarget, there are extra fields.

### Implementation 👩‍💻👨‍💻

The implementation followed from the other PBX objects I found lying around.

- [x] Mess around with PBXLegacyTarget until I could see something in Xcode
- [x] Let everyone know on the `#xcodeswift` channel on Slack
- [x] Implement it for real, with all the properties
- [ ] Test properly?

I tested by playing with XcodeGen, let me know how to write real tests!

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/171)

### DIFF
--- a/Sources/xcproj/PBXLegacyTarget.swift
+++ b/Sources/xcproj/PBXLegacyTarget.swift
@@ -1,0 +1,108 @@
+import Foundation
+
+/// This is the element for a build target that according to Xcode is an "External Build System". You can use this target to run a script.
+final public class PBXLegacyTarget: PBXTarget {
+    /// Path to the build tool that is invoked (required)
+    public var buildToolPath: String?
+    
+    /// Build arguments to be passed to the build tool.
+    public var buildArgumentsString: String?
+    
+    /// Whether or not to pass Xcode build settings as environment variables down to the tool when invoked
+    public var passBuildSettingsInEnvironment: Bool
+    
+    /// The directory where the build tool will be invoked during a build
+    public var buildWorkingDirectory: String?
+    
+    public init(reference: String,
+                name: String,
+                buildToolPath: String? = nil,
+                buildArgumentsString: String? = nil,
+                passBuildSettingsInEnvironment: Bool = false,
+                buildWorkingDirectory: String? = nil,
+                buildConfigurationList: String? = nil,
+                buildPhases: [String] = [],
+                buildRules: [String] = [],
+                dependencies: [String] = [],
+                productName: String? = nil,
+                productReference: String? = nil,
+                productType: PBXProductType? = nil) {
+        self.buildToolPath = buildToolPath
+        self.buildArgumentsString = buildArgumentsString
+        self.passBuildSettingsInEnvironment = passBuildSettingsInEnvironment
+        self.buildWorkingDirectory = buildWorkingDirectory
+        super.init(reference: reference,
+                   name: name,
+                   buildConfigurationList: buildConfigurationList,
+                   buildPhases: buildPhases,
+                   buildRules: buildRules,
+                   dependencies: dependencies,
+                   productName: productName,
+                   productReference: productReference,
+                   productType: productType)
+    }
+    
+    // MARK: - Decodable
+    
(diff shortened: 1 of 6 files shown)


pepibumur:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry. <br />
You can find the guidelines at <a href="https://github.com/xcodeswift/contributors/blob/master/CHANGELOG_GUIDELINES.md">CHANGELOG_GUIDELINES.md</a>.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">Danger</a>
</p>

Sources/xcproj/PBXLegacyTarget.swift
```diff
@@ -0,0 +1,109 @@
...
+    public static func == (lhs: PBXLegacyTarget,
+                           rhs: PBXLegacyTarget) -> Bool {
+        return (lhs as PBXTarget) == (rhs as PBXTarget) &&
+            lhs.buildToolPath == rhs.buildToolPath &&
+            lhs.buildArgumentsString == rhs.buildArgumentsString &&
+            lhs.passBuildSettingsInEnvironment == rhs.passBuildSettingsInEnvironment &&
+            lhs.buildWorkingDirectory == rhs.buildWorkingDirectory
+    }
+    
+    override func plistValues(proj: PBXProj, isa: String) -> (key: CommentedString, value: PlistValue) {
```
pepibumur: We didn't use to test this object method but we started to do it recently because it's usually the source of inconsistencies when saving the project back *(e.g. bad comment, missing comment)*. `PlistValue` [is already equatable](https://github.com/xcodeswift/xcproj/blob/master/Sources/xcproj/PlistValue.swift#L74) so it should facilitate writing the test. I think it's important to test that the values have the proper comment and that are written with the proper key.

Sources/xcproj/PBXLegacyTarget.swift
```diff
@@ -0,0 +1,109 @@
...
+        case buildToolPath
+        case buildArgumentsString
+        case passBuildSettingsInEnvironment
+        case buildWorkingDirectory
+        case buildConfigurationList
+        case buildPhases
+        case buildRules
+        case dependencies
+        case name
+        case productName
```
yonaskolb: You don’t need the superclass keys

Sources/xcproj/PBXLegacyTarget.swift
bkase: If I'm understanding this right: You want me to confirm that the Plist comments are the same as what would exist if this target were created from Xcode? I just confirmed that all these new fields do not have any Plist comments.

Or are you saying that I should write a unit test somewhere? Can you point me to an example of a plistValues one that already exists (I can't seem to find it).

pepibumur: I meant more your second point. You can check out [these tests](https://github.com/xcodeswift/xcproj/blob/master/Tests/xcprojTests/PBXSourceTreeSpec.swift#L36) where we test that the returned `plistValues` are the expected ones.

pepibumur:
@bkase dont' forget to update the CHANGELOG with your changes. Moreover, it seems that there's an issue with swiftlint. Please, let me know if I can help you with these issues.

PR #173, opened 2017-11-30
Make PBXFileElement a superclass

gubikmic:
This PR makes PBXFileElement a superclass for PBXFileReference, PBXGroup, and PBXVariantGroup.

According to http://www.monobjc.net/xcode-project-file-format.html PBXFileElement is an abstract superclass, so it shouldn't occur as itself. Empirically this is supported by the fact that I grepped through all project.pbxproj files on my disk and there is none that uses this string.
(`find ~/src -name 'project.pbxproj' -exec rg PBXFileElement \{\} \;`)

The benefit of using the common parent is that the three shared fields (name, path, sourceTree) have to be dealt with only in one class.

Important:
I did this without thinking a lot, just change & break, then fix until all tests pass.
So please go through it with a fine comb and try to break things. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/173)


pepibumur:
Great catch @gubikmic. Looks good to me. Make sure you update the `CHANGELOG.md` and if everything passes it should be ready to be merged. Nice job 🎉

daveanderson:
Glad I saw this. I was about to submit a PR to add a path property to PBXVariantGroup because my project has one.

PR #175, opened 2017-12-02
Add getFileElement

yonaskolb:
- make `XCVersionGroup` subclass `PBXFileElement`
- add `PBXProj.Objects.getFileElement`
- use `getFileElement` in `func fileName(fileReference: String) -> String?`

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/175)

### DIFF
--- a/Sources/xcproj/PBXProj+Helpers.swift
+++ b/Sources/xcproj/PBXProj+Helpers.swift
@@ -22,17 +22,10 @@ extension PBXProj {
     /// - Parameter fileReference: file reference.
     /// - Returns: file name.
     func fileName(fileReference: String) -> String? {
-        if let variantGroup: PBXVariantGroup = objects.variantGroups.getReference(fileReference) {
-            return variantGroup.name
-        } else if let group: PBXGroup = objects.groups.getReference(fileReference) {
-            return group.name
-        } else if let versionGroup: XCVersionGroup = objects.versionGroups.getReference(fileReference) {
-            return versionGroup.name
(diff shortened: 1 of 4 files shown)


Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -22,17 +22,10 @@ extension PBXProj {
...
-            return group.name
-        } else if let versionGroup: XCVersionGroup = objects.versionGroups.getReference(fileReference) {
-            return versionGroup.name
-        } else if let fileReference: PBXFileReference = objects.fileReferences.getReference(fileReference) {
-            return fileReference.name ?? fileReference.path
-        } else {
+        guard let fileElement = objects.getFileElement(reference: fileReference) else {
             return nil
         }
+        return fileElement.name ?? fileElement.path
```
yonaskolb: Will this cause any issues? Before only the fileReference was returning the path if no name was present. If so it can be changed @pepibumur

pepibumur:
Awesome cleanup @yonaskolb 👌

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sources/xcproj/PBXProj+Helpers.swift
pepibumur: AFAIK it shouldn't @yonaskolb. I think groups, variant groups and version groups they should all have a name so the path wouldn't be returned in those cases.

PR #180, opened 2017-12-08
Skip writing if nothing changed

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/128

### Short description 📝
As described in the issue the purpose of this PR is to update the write logic to skip writing if nothing changed. This prevents Xcode from reloading the project unnecessarily.

### Solution 📦
Update implementation and add some tests.

### Implementation 👩‍💻👨‍💻
- [x] PBXProj
- [x] XCWorkspaceData
- [x] XCScheme

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/180)

### DIFF
--- a/Fixtures/iOS/Project.xcodeproj/project.pbxproj
+++ b/Fixtures/iOS/Project.xcodeproj/project.pbxproj
@@ -7,7 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		0403345F1F8F57CB0079CCF5 = {isa = PBXBuildFile; fileRef = 0403345D1F8F57CB0079CCF5; };
+		0403345F1F8F57CB0079CCF5 /* Model.xcdatamodeld in Sources */ = {isa = PBXBuildFile; fileRef = 0403345D1F8F57CB0079CCF5 /* Model.xcdatamodeld */; };
 		04D5C09F1F153824008A2F98 /* CoreData.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 04D5C09E1F153824008A2F98 /* CoreData.framework */; };
 		04D5C0A31F153924008A2F98 /* Public.h in Headers */ = {isa = PBXBuildFile; fileRef = 04D5C0A01F153915008A2F98 /* Public.h */; };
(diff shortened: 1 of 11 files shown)


keith:
One thing to note is that for anyone using CocoaPods the XML will never be the same since CocoaPods adds some references and build phases.

Fixtures/iOS/Project.xcodeproj/project.pbxproj
```diff
@@ -7,7 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		0403345F1F8F57CB0079CCF5 /* Model.xcdatamodeld in Sources */ = {isa = PBXBuildFile; fileRef = 0403345D1F8F57CB0079CCF5 /* Model.xcdatamodeld */; };
+		0403345F1F8F57CB0079CCF5 = {isa = PBXBuildFile; fileRef = 0403345D1F8F57CB0079CCF5; };
```
toshi0383: Why are these changes necessary?

toshi0383:
> One thing to note is that for anyone using CocoaPods the XML will never be the same since CocoaPods adds some references and build phases.

Yea CocoaPods even updates xcodeproj even when Podfile only contains command-line tools like Sourcery. 🤷‍♂️

rahul-malik:
@yonaskolb - Do you think XcodeGen could leverage this change to update a project vs delete / create a new one?

Fixtures/iOS/Project.xcodeproj/project.pbxproj
pepibumur: Nop, I've reverted the change. Good catch 👍

yonaskolb:
I’m actually going to re-implement this in XcodeGen anyway, to have some oversight if anything changed or not, for logging purposes.
I’m actually not 100% convinced this should live in xcproj. Feels weird to have an api do something only in certain cases. Maybe someone wants to rewrite the file for updated permissions or metadata like modified. 
At the very least there should be some way to find out that a file hasn’t actually been written.

pepibumur:
> Maybe someone wants to rewrite the file for updated permissions or metadata like modified.

Isn't modifying permissions or metadata something that you would do without xcproj?

pepibumur:
I'll close this PR leaving up to the developers using xcproj to decide whether they want to override the project all the times or check if anything changed before saving.

PR #181, opened 2017-12-08
Multithread parsing

pepibumur:
### Short description 📝
Update the `pbxproj` object parsing to do it in multiple threads.

### Solution 📦
I've defined a new object `PBXObjectsParser` responsible for parsing the objects. This one uses promises that do the asynchronous parsing in multiple threads. The result of all these promises is merged and returned synchronously.

### Implementation 👩‍💻👨‍💻
- [x] Implement `Promise`.
- [x] Implement `PBXObjectParser`.
- [ ] Test it.
- [ ] Measure impact.

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/181)


pepibumur:
I'm curious to know if it improves the parsing time at Pinterest @rahul-malik. Moreover, I'd like to ask you @yonaskolb & @rahul-malik what do you think about the asynchronous model using promises. Is there any other way you can think of that could simplify the current implementation?

pepibumur:
I didn't check yet but it'd be interesting to know as well the memory consumption. During the iteration when I create the promises I create copies of the dictionaries. I'm concerned that it might result in a memory spike.

bkase:
Hey @pepibumur , this seems great. I'd suggest using a mature async-monad library (Promise or Future or Defferred etc) as it's really really easy to miss subtle concurrency bugs when you roll your own (I've hit a few early-on in more than one async-monad library). I haven't used this recently, but my favorite has been the https://github.com/Thomvis/BrightFutures library. There are lots of useful primitives there and since error is parameterized you can specify `NoError` and get a contract at compile time that you don't need to handle the error case.

With `BrightFutures` you could do something like this:

```swift
    /// Parses the objects contained in the given dictionary.
    ///
    /// - Parameter objects: dictionary with the objects.
    /// - Returns: array with all the objects
    /// - Throws: an error if the parsing fails
    func parse(objects: [String: [String: Any]]) throws -> [PBXObject] {
        // we shouldn't need a batch size because the dispatch queue should handle that for us
        let allFuture = objects.map { 
             Future { complete in
                    DispatchQueue.global().async {
                    do {
                        let values = try copiedDictionary.flatMap { try PBXObject.parse(reference: $0.key, dictionary: $0.value) }
                        complete(.success(values))
                    } catch {
                        complete(.failure(error))
                    }
               }
           }.sequence()
        return allFuture.forced().dematerialize()
    }
```

pepibumur:
@rahul-malik I've tried the current implementation using Futures and parsing the objects in different threads and surprisingly the time it takes is slower compared to single-thread parsing. My guess is that creation of threads and synchronization of results is costly and ends up affecting the parsing time. Would you mind checking if the same happens to you?

By the way... I've created an `opensave` target that you can use for your tests.

rahul-malik:
@pepibumur - Cool! Haven't had a chance to try it out yet. I assume the performance depends on the size of the project as well. Since we're using `xcproj` through `XcodeGen` it's more likely that we're going to see perf improvements through the writing flows.

pepibumur:
I'll close this for now. If the discussion arises again we can reopen it in the future.

PR #184, opened 2017-12-12
Add convenience method to get targets with a given name

pepibumur:
### Short description 📝
Add convenience method to get targets with a given name

### Solution 📦
Extend `PBXProj.Objects` and add the convenience method in there.

### Implementation 👩‍💻👨‍💻
- [x] Add and test the method
- [x] Move internal helpers from `PBXProj` to `PBXProj.Objects`.

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/184)


Package.resolved
```diff
@@ -10,6 +10,15 @@
           "version": "4.1.0"
         }
       },
+      {
+        "package": "BrightFutures",
```
toshi0383: Why are these diffs in this PR?

allu22: +1, it seems these are unnecessary packages.

pepibumur: Ouch, true. I'll revert them.

toshi0383: You reverted it, but restored again.🤣

PR #84, opened 2017-09-20
Swift 4 migration

artemnovichkov:
**Issue:** [Link](https://github.com/swift-xcode/xcodeproj/issues/82)

### Short description
Migrate the project to Swift 4

### Solution
Build the project with Xcode 9, fix all warnings and errors and update SPM manifest syntax.

### Implementation

- [x] Update `Package.swift` file. Please, check versions of dependencies. In my opinion, it's good practice to update dependencies to next major version.
- [x] Fix new syntax errors in `sorted(by:)` and `NSTextCheckingResult.range(at:)` functions.

### GIF
Smile

PR #185, opened 2017-12-12
Remove Referenceable

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/31

### Short description 📝
Objects references were duplicated. They were defined in:
- The dictionary key that references the object.
- The reference property of the object.

By having them in two places it was possible to have an object in a dictionary under a key that didn't match the object reference. 

The goal of this PR is to remove such duplication aiming for a safer API, and make it easier for developers to use `xcproj` to generate deterministic references when creating new objects.

### Solution 📦
The `Referenceable` protocol has been removed. `PBXObject`s don't have a reference property anymore. The reference is the key in the dictionary that contains the object. The creation of the pair reference-object is done:
- When the project is parsed.
- When the developer adds a new object with a given reference. 

### Implementation 👩‍💻👨‍💻
- [x] Remove `Referenceable` protocol and update the classes accordingly
- [x] Remove deprecated methods *(read below)*.
- [x] Fix broken tests.

> ⚠️ This update will require a major bump in the version since it deletes and updates classes and methods from the public interface. 

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/185)


Sources/xcproj/PBXGroup.swift
```diff
@@ -39,13 +37,12 @@ final public class PBXGroup: PBXFileElement {
         self.usesTabs = usesTabs
         self.indentWidth = indentWidth
         self.tabWidth = tabWidth
-        super.init(reference: reference, sourceTree: sourceTree, path: path, name: name)
+        super.init(sourceTree: sourceTree, path: path, name: name)
     }
 
     public static func == (lhs: PBXGroup,
```
toshi0383: Just wondering: Is there any reason why your not using AutoEquatable from Sourcery?

Sources/xcproj/PBXObject.swift
```diff
@@ -1,27 +1,19 @@
...
-    public let reference: String
-
-    init(reference: String) {
-        self.reference = reference
-    }
+public class PBXObject: Decodable {
 
+    // MARK: - Init
+    
+    init() {}
```
yonaskolb: This class should not actually be initiable as it's abstract, so we can remove this

Package.resolved
```diff
@@ -10,6 +10,15 @@
           "version": "4.1.0"
         }
       },
+      {
+        "package": "BrightFutures",
```
yonaskolb: This is from a different PR isn't it?

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -110,6 +110,17 @@ extension PBXProj {
             return type?.rawValue
         }
     }
+    
+    /// Returns the object with the given configuration list (project or target)
+    ///
+    /// - Parameter reference: configuration list reference.
+    /// - Returns: target or project with the given configuration list.
+    func objectWithConfigurationList(reference: String) -> PBXObject? {
```
yonaskolb: Should this be under objects, as #84 moves all these helpers

Sources/xcproj/PBXProj.swift
```diff
@@ -73,33 +74,65 @@ final public class PBXProj: Decodable {
...
         }
+        
+        /// Generates a deterministic reference from an object type and identifier.
+        /// It ensures that the generated reference doesn't collide with any existing one.
+        ///
+        /// - Parameters:
+        ///   - element: object type.
+        ///   - id: object identifier (e.g. path or name)
+        /// - Returns: reference.
+        public func generateReference<T: PBXObject>(_ element: T.Type, _ id: String) -> String {
```
yonaskolb: Will the old generation function get deleted?

Sources/xcproj/PBXGroup.swift
pepibumur: Not really. I actually think it's a good idea to automate the generation of init and the conformance of Equatable. Creating an issue 😛

Sources/xcproj/PBXObject.swift
pepibumur: Good point, removing.

Package.resolved
pepibumur: Yeah, reverting.

Sources/xcproj/PBXProj.swift
pepibumur: Yes it will, I'll remove it.

Sources/xcproj/PBXProj+Helpers.swift
pepibumur: Do you mean [this PR](https://github.com/xcodeswift/xcproj/pull/184/files)? Yeah it should be there. My idea was to merge that PR first and then update this one moving the helper under objects.

Sources/xcproj/PBXGroup.swift
pepibumur: https://github.com/xcodeswift/xcproj/issues/186

Sources/xcproj/PBXObject.swift
yonaskolb: Ah ok. I guess it’s not public anyway so not a big issue

Sources/xcproj/PBXProj+Helpers.swift
yonaskolb: Yeah, whoops, meant that one

pepibumur: Nice, I'll merge that one first and move this function before merging 👍

Sources/xcproj/PBXProj.swift
pepibumur: Removed

Sources/xcproj/PBXProj.swift
```diff
@@ -111,14 +111,14 @@ final public class PBXProj: Decodable {
         /// It ensures that the generated reference doesn't collide with any existing one.
         ///
         /// - Parameters:
-        ///   - element: object type.
+        ///   - object: the object to generate the reference for.
         ///   - id: object identifier (e.g. path or name)
         /// - Returns: reference.
-        public func generateReference<T: PBXObject>(_ element: T.Type, _ id: String) -> String {
+        public func generateReference<T: PBXObject>(_ object: T, _ id: String) -> String {
```
yonaskolb: No need for this function to be generic anymore. T can just be a PBXObject

pepibumur: Lol @yonaskolb I just updated it before I saw your comment. It's not generic anymore.

PR #189, opened 2017-12-13
Equatable conformate using Sourcery

pepibumur:
Resolves https://github.com/xcodeswift/xcproj/issues/186

### Short description 📝
I've added Sourcery to generate the code to compare `PBXObjects`.


toshi0383:
I think you'd better to add Sourcery via Package.swift or Podfile. Users don't want to clone Sourcery binary along side of source-codes.
But make sure you don't expose Sourcery in Package.swift. That would cause dependency conflicts on users side. You can set&check environment variable for that purpose.
See: https://github.com/kylef/Commander/pull/63

pepibumur:
Great idea @toshi0383. I didn't think about the fact that users would get the binary when SPM clones the repository. I'll do what you suggested and move it to the `Package.swift` using environment variables. Thanks for the idea!

pepibumur:
@toshi0383 I ended up using @yonaskolb's Mint. I added a Rake task that uses Mint to run Sourcery.

Rakefile
```diff
@@ -98,6 +98,11 @@ task :generate_carthage_project do
   generate_carthage_project()
 end
 
+desc "Generate Sourcery files"
+task :sourcery do
+  sh "mint run krzysztofzablocki/Sourcery sourcery"
```
toshi0383: should pin to specific version of Sourcery?

pepibumur: Yep @toshi0383. I'm waiting for this to be merged to pin to the proper version https://github.com/krzysztofzablocki/Sourcery

Package.resolved
```diff
@@ -19,6 +28,15 @@
           "version": "0.8.0"
         }
       },
+      {
+        "package": "Result",
```
yonaskolb: I think there's something wrong with your local repo, as these changes keep occurring in new PR's

toshi0383: 🙈

pepibumur: 🙄🙄🙄🙄

yonaskolb:
This would be good to get in as I just noticed a missing equatable check for `PBXTarget.productName`, which led to a test not failing when it was supposed to in XcodeGen

pepibumur:
Definitively @yonaskolb. My goal is to work on this one after updating XcodeGen to support xcproj 2.0. Once we get equatable I'll also implement the template for hashable so that we can simplify the reference generation.

Sources/xcproj/PBXBuildFile.swift
```diff
@@ -9,6 +9,7 @@ final public class PBXBuildFile: PBXObject {
     public var fileRef: String?
 
     /// Element settings
+    // sourcery: notEquatableDictionary
```
yonaskolb: Is this type not supported by Sourcery?

Sources/xcproj/PBXProject.swift
```diff
@@ -41,6 +41,7 @@ final public class PBXProject: PBXObject {
     public var targets: [String]
     
     /// Project attributes.
+    // sourcery: notEquatableDictionary
     public var attributes: [String: Any]
```
yonaskolb: Is this not supported by Sourcery?

Sources/xcproj/PBXBuildFile.swift
pepibumur: Sourcery doesn't know anything about how to generate the models. It uses the templates that you specify. I could have mad the template more complex by traversing this type and checking if it's a non-equatable dictionary, but it was simpler using annotations.

Sources/xcproj/PBXProject.swift
pepibumur: Same comment here.

Sources/xcproj/PBXBuildFile.swift
yonaskolb: I meant don't we want to check for this value as well? Otherwise we're losing functionality

Templates/Equality.stencil
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+{% for type in types.implementing.AutoEquatable|class|!annotated:"skipEquality" %}
+extension {{ type.name }} {
+    /// :nodoc:
+    @objc override {{ type.accessLevel }} func isEqual(to object: Any?) -> Bool {
+      guard let rhs = object as? {{ type.name }} else { return false }
+      {% for variable in type.storedVariables %}
+      {% if variable|annotated:"notEquatableDictionary" %}
+      if !NSDictionary(dictionary: self.{{ variable.name}} {% if variable.typeName.isOptional %}?? [:]{% endif %}).isEqual(to: rhs.{{ variable.name }} {% if variable.typeName.isOptional %}?? [:]{% endif %}) { return false }
```
ilyapuchka: I would just check for `[String: Any]` type, that's the only type you need this for. Just use `actualTypeName` instead of `typeName` to include typealias too.

Sources/xcproj/PBXBuildFile.swift
yonaskolb: Oh, I see. You are checking for it, but using this tag to differentiate it in the template. Got it

Templates/Equality.stencil
pepibumur: How can I write that condition @ilyapuchka? I tried this but doesn't seem to work:

```
{% if variable|actualTypeName:"[String: Any]" %}
```

pepibumur:
That was the problem @ilyapuchka. I've updated it and it works now 😀. Thank you very much.

yonaskolb:
Awesome, great work @pepibumur 🎉

yonaskolb:
@pepibumur could we make `XcodeProj` AutoEquatable as well to resolve #200

PR #191, opened 2017-12-18
Optimize PBXObject lookups

kastiglione:
### Short description 📝

This change has the effect of, on my machine with my project, bringing XcodeGen's **runtime down by ~80%**, and **reducing total allocated memory down by 2.1gb**.

### Solution 📦

`PBXProj` uses a pattern which collects dictionaries into array in order to search for a specific key.

The change to `getFileElement()` replaces this search pattern with `??` chaining, because profiling shows this existing search pattern to be poorly optimized by swift, and thus `getFileElement()` is a large hot spot.

In particular, for project generation with XcodeGen, the project I'm working with went from 9-10s on my machine, to 1.5-2s. Total memory allocations went down from 2.6gb to 0.5gb.

In Instruments, the category `_HashableTypedNativeDictionaryStorage<String, PBXFileElement>` had a total allocation count of 61595, for a sum total of 2.13gb.  With this change, Instruments shows no such allocations.

All these allocations were wasting cpu as well. With this change, `_swift_release_` and `_swift_retain_` were reduced to 85ms each, down from 2.3s and 1.9s respectively.

The allocation was caused by a function named `_dictionaryUpCast()`, and this is presumably happening because these dictionaries, which are each typed to a specific subclass of `PBXObject`, are being added to an array of `[String: PBXObject]`.

### Implementation 👩‍💻👨‍💻

- [x] Profile XcodeGen on an internal project, read through the data to find potential optimizations

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/191)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Next version
 
 ### Fixed
-- Optimized performance of `PBXProj.getFileElement()` https://github.com/xcodeswift/xcproj/pull/191 by @kastiglione
+- Optimised performance of object lookups https://github.com/xcodeswift/xcproj/pull/191 by @kastiglione
 ### Added
 - Support more indentation options on PBXGroups https://github.com/xcodeswift/xcproj/pull/168 by @bkase.
 - Support `PBXLegacyTarget` https://github.com/xcodeswift/xcproj/pull/171 by @bkase.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next version
 
+### Fixed
+- Optimized performance of `PBXProj.getFileElement()` https://github.com/xcodeswift/xcproj/pull/191 by @kastiglione
 ### Added
 - Support more indentation options on PBXGroups https://github.com/xcodeswift/xcproj/pull/168 by @bkase.
 - Support `PBXLegacyTarget` https://github.com/xcodeswift/xcproj/pull/171 by @bkase.
--- a/Sources/xcproj/PBXProj.swift
+++ b/Sources/xcproj/PBXProj.swift
@@ -110,13 +110,10 @@ final public class PBXProj: Decodable {
         }
 
         public func getFileElement(reference: String) -> PBXFileElement? {
-            let caches: [[String: PBXFileElement]] = [
-                fileReferences,
-                groups,
-                variantGroups,
-                versionGroups,
-                ]
-            return caches.first { cache in cache[reference] != nil }?[reference]
+            return fileReferences[reference] ??
+                groups[reference] ??
+                variantGroups[reference] ??
+                versionGroups[reference]
         }
 
         public func getReference(_ reference: String) -> PBXObject? {
--- a/Sources/xcproj/PBXProj.swift
+++ b/Sources/xcproj/PBXProj.swift
@@ -114,6 +114,10 @@ final public class PBXProj: Decodable {
         }
 
         public func getReference(_ reference: String) -> PBXObject? {
+            // The ?? chaining below is partitioned into separate statments because, with Swift 4, a single
+            // expression resulted in this compiler error:
+            //     Expression was too complex to be solved in reasonable time;
+            //     consider breaking up the expression into distinct sub-expressions
             let part1 = buildFiles[reference] ??
                 aggregateTargets[reference] ??
                 legacyTargets[reference] ??
--- a/Sources/xcproj/PBXProj.swift
+++ b/Sources/xcproj/PBXProj.swift
@@ -114,29 +114,29 @@ final public class PBXProj: Decodable {
         }
 
         public func getReference(_ reference: String) -> PBXObject? {
-            let caches: [[String: PBXObject]] = [
-                buildFiles,
-                aggregateTargets,
-                legacyTargets,
-                containerItemProxies,
-                groups,
-                configurationLists,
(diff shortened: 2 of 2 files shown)


kastiglione:
Note that only `getFileElement` was changed, but `getTarget` and `getReference` also use the same non-optimal pattern. This is for two reasons, 1) only `getFileElement` was a hot spot, and 2) using `??` chaining in `getReference` causes the swift compiler to error because it can't type check the expression without hitting its type checking time limit. Interested in suggestions!

If the result of this pull request is that only `getFileElement` is updated, then I will add comments to these functions for future reference.

yonaskolb:
Very cool! Thanks @kastiglione!
If we can get the others changed as well it would be awesome. Does casting each lookup to the return type help the compiling issue?
@rahul-malik

kastiglione:
> Does casting each lookup to the return type help the compiling issue?

Thanks for the suggestion. I just tried it but it did not work.

yonaskolb:
Then I guess just a bunch of if lets will do the trick

kastiglione:
The last commit has the `??` chaining, but partitioned into separate statements. How do you feel about that?

Sources/xcproj/PBXProj.swift
```diff
@@ -101,48 +101,46 @@ final public class PBXProj: Decodable {
...
-                referenceProxies,
-                copyFilesBuildPhases,
-                shellScriptBuildPhases,
-                resourcesBuildPhases,
-                frameworksBuildPhases,
-                headersBuildPhases,
-                sourcesBuildPhases
-            ]
-            return caches.first { cache in cache[reference] != nil }?[reference]
+            // The ?? chaining below is partitioned into separate statments because, with Swift 4, a single
```
yonaskolb: This may be better served with: 
```
if let object = buildFiles[reference] {
    return object
} else if let object = aggregateTargets[reference] {
    return object
} ...

```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Next version
 
+### Fixed
+- Optimized performance of `PBXProj.getFileElement()` https://github.com/xcodeswift/xcproj/pull/191 by @kastiglione
```
yonaskolb: I guess you can rename this to `Optimised performance of object lookups`

Sources/xcproj/PBXProj.swift
```diff
@@ -101,48 +101,46 @@ final public class PBXProj: Decodable {
...
+            // expression resulted in this compiler error:
+            //     Expression was too complex to be solved in reasonable time;
```
pepibumur: Out of curiosity is the error thrown when you try to build from Xcode or when you build from the console? @kastiglione

kastiglione: Xcode. I haven't yet built from console.

pepibumur: 👍

Sources/xcproj/PBXProj.swift
```diff
@@ -114,33 +114,53 @@ final public class PBXProj: Decodable {
         }
 
         public func getReference(_ reference: String) -> PBXObject? {
-            // The ?? chaining below is partitioned into separate statments because, with Swift 4, a single
-            // expression resulted in this compiler error:
+            // This if-let expression is used because the equivalent chain of `??` separated lookups causes,
+            // with Swift 4, this compiler error:
```
yonaskolb: I don't think this comment is needed anymore. If let'ing like this is a standard practice, so no need to explain it 👍

kastiglione: The other two functions aren't using `if-let`s, so I wanted to explain why this one is different. Prior to this investigation, I wouldn't have been able to look at this code and know why it ended up being written this way. An `if-let` this large looks a bit off in my opinion.

kastiglione:
Note I updated the description with a bit of detail about the cause of the inefficiency. See the part about `_dictionaryUpCast`.

Sources/xcproj/PBXProj.swift
kastiglione: Another reason I thinks this comment is valuable is it would be good if this `if-let` could one day be converted to use a chain of `??` lookups. This comment provides that context.

yonaskolb: Fair enough 👍

rahul-malik:
@kastiglione - Great fix! I've been trying to track down why XcodeGen was consuming so much memory. Looking forward to re-testing after this lands 💯

kastiglione:
Thanks!

yonaskolb:
Yeah great work @kastiglione! @rahul-malik, I'd love to hear some updates about generation time, referring to the times posted here https://github.com/yonaskolb/XcodeGen/pull/158

rahul-malik:
@yonaskolb - I haven't done as thorough of a benchmark as before but this patch actually addresses what would have been a regression for us. The original `getFileElement` addition spiked our generation time from 30 seconds to nearly 300 seconds. This patch brought us back down to 30 seconds where we were before.

yonaskolb:
Oh ok, this is almost a hotfix then. 
@pepibumur It might even be worth releasing a `1.7.1` if `2.0.0` is going to take a while

pepibumur:
Yeah @yonaskolb. If I don't manage to finish this #185 PR today I'll release the patch at the end of the day.

PR #194, opened 2017-12-19
Better Support Loading Legacy Schemes

briantkelley:
### Short description 📝
Xcode accepts just about any scheme file, filling the model with default values if necessary. xcproj rejected a number of schemes Xcode accepted in a codebase with 279 Xcode projects.

### Solution 📦
Update the XCScheme XML initialization to better match Xcode's behavior.

Note that the `buildableProductRunnable` property/element isn't required. It's possible to select "None" for the executable in the Scheme Editor, in which case Xcode won't generate the element.

The xcproj tests test xcodegen, which creates a circular dependency for validation. Because this change breaks the API, we have to fix up the xcodegen source files until it adopts the version of xcproj that contains this change.

### Implementation 👩‍💻👨‍💻
- [x] Add MinimalInformation.xcscheme, which is somewhat of a worse case scheme, to validate the changes to the scheme loading code
- [x] Open MinimalInformation.xcscheme in Xcode's Scheme Editor and close the sheet, which causes it to rewrite the scheme with default values.
- [x] Integrate the default values added by Xcode in to the XCScheme loading code as fallbacks for when the attribute or element doesn't exist.
- [x] Update the XCScheme integration test to test reading and writing the underspecified scheme to ensure it can be successfully loaded with the correct default values.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/194)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 ### Added
 - Add breakpoint `condition` parameter by [@alexruperez](https://github.com/alexruperez).
 
+### Changed
+- **Breaking:** The `buildableProductRunnable` property on`XCScheme.LaunchAction` and `XCScheme.ProfileAction` is now optional. Similarly, `macroExpansion` on `XCScheme.TestAction` is also optional. https://github.com/xcodeswift/xcproj/pull/194 by @briantkelley
+- The `XCScheme` initialization from an XML file has been relaxed, better matching Xcode's behavior. Default values will be used if the XML file is missing the relevant element or attribute. https://github.com/xcodeswift/xcproj/pull/194 by @briantkelley
+
 ## 1.7.0
 
 ### Added
--- a/Fixtures/Schemes/MinimalInformation.xcscheme
+++ b/Fixtures/Schemes/MinimalInformation.xcscheme
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   version = "1.3">
+   <BuildAction>
+      <BuildActionEntries>
+         <BuildActionEntry
+            buildForRunning = "NO"
+            buildForTesting = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "651BA89E1E459798004EAFE5"
+               BuildableName = "ais.xctest"
+               BlueprintName = "ais-Tests"
+               ReferencedContainer = "container:ais.xcodeproj">
+            </BuildableReference>
+         </BuildActionEntry>
+      </BuildActionEntries>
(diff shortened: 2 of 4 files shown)


pepibumur:
Thanks @briantkelley for improving the XCScheme initialization to match the Xcode behavior. We did that for the projects but we forgot about the schemes. Regarding your comment on the breaking change I've updated your PR to have `release/2.0.0` as the merge base. There's another breaking change #185 that will also go into that release. We'll merge these two PRs into `release/2.0.0` and get XcodeGen working with that version before doing the release.

Fixtures/Schemes/AIS.xcscheme
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
```
pepibumur: Out of curiosity, what does `AIS` stand for?

Tests/xcprojIntegrationTests/XcodeGenTests.swift
```diff
@@ -34,6 +34,14 @@ final class XcodeGenTests: XCTestCase {
         try packageSwiftWithLocalDependency.write(to: packageSwiftPath.url,
                                                   atomically: true,
                                                   encoding: .utf8)
+        // Testing xcodegen as part of xcproj creates a circular dependency; fix up source files affected by API changes
+        let projectGeneratorTestsPath = clonePath + Path("Tests/XcodeGenKitTests/ProjectGeneratorTests.swift")
```
pepibumur: I'm sorry that how we test xcproj is not well documented @briantkelley. `XcodeGenTests` is part of a suite of tests, integration tests, that we don't run from PRs. These tests are run after very merge into master to know if there was any change that is breaking. We test that by using XcodeGen and updating the `Package.swift` to point to our changes.
If you revert these changes this PR should pass *(ignore the local failures since they are expected)*

briantkelley: Thanks for the info. I saw `swift test` failed and thought I should fix it. But merging it in to the 2.0 branch and updating the framework makes so much more sense.

Fixtures/Schemes/AIS.xcscheme
briantkelley: Automation Infrastructure Shared, which is an internal library for testing support. I wasn't sure if the name of a file loaded during a test case had much value, but I'd be happy to rename it to something a bit more useful. Perhaps LotsOfMissingInformation.xcscheme?

briantkelley:
Thanks for the feedback @pepibumur. I renamed the scheme to something a bit more descriptive and removed the hack for the tests given the branching strategy.

PR #195, opened 2017-12-19
Optimization: Only escape CommentedString when necessary

kastiglione:
### Short description 📝
Reduce unnecessary string allocations in `CommentedString.validString`. For the project I tested on, this reduced the total allocated memory used by XcodeGen by just under 60mb, from ~459mb to ~402mb. Instruments showed a 48% reduction in total `Swift._NSContiguousString` allocations, down from ~1.381 million to ~716k.

### Solution 📦
Every call to `replacingOccurrences()` creates a new copy of the string, even when no replacements occur. For `CommentedString.validString`, this means 4 copies of the string are made every time, even in a common case where no escaping is required. This code is called frequently and is a hotspot.

Additionally, Instruments is showing an allocation for each swift `String` bridged to objective-c. This means each avoided `replacingOccurrences()` call is also saving 2 string allocations for the pattern and replacement strings. The bridging cost is why the `contains()` checks are explicitly typed to `Character`. Without the explicit type, the `String` version of `contains()` is called.

### Implementation 👩‍💻👨‍💻

- [x] Ran Instruments with the Allocations template, then starting from the big numbers (of allocations), I followed the data trail

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/195)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next version
 
 ### Fixed
+- Optimised escaping of CommentedString https://github.com/xcodeswift/xcproj/pull/195 by @kastiglione
 - Optimised performance of object lookups https://github.com/xcodeswift/xcproj/pull/191 by @kastiglione
 ### Added
 - Add breakpoint `condition` parameter by [@alexruperez](https://github.com/alexruperez).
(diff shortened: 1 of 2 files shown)


kastiglione:
This makes XcodeGen run a bit faster too. My project runs in around ~1.5s but the run times vary by enough that I didn't want to label a number on how much faster.

kastiglione:
Thanks!

PR #197, opened 2017-12-19
Add Support for Custom Build Rules

briantkelley:
### Short description 📝
Add support for custom build rules, which Xcode uses to automatically invoke a tool when a matching input file changes.

### Solution 📦
Add a PBXObject subclass, PBXBuildRule, which can load and store the custom build rule configuration.

### Implementation 👩‍💻👨‍💻
- [x] Create `PBXBuildRule`, which is a subclass of `PBXObject` and add properties for all possible fields in the Xcode project file. Implement Equatable, Decodable, and encoding support.
- [x] Update `PBXObject` and `PBXProj` to support decoding
- [x] Update `PBXProj` and `PBXProjEncoder` to support encoding
- [x] Add and update unit tests
- [x] Manually verify xcproj correctly round trips a project with custom build rules (custom pattern/custom tool, custom pattern/built-in tool, built-in pattern/custom tool, built-in pattern/built-in tool)

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/197)

### DIFF
--- a/Sources/xcproj/PBXBuildRule.swift
+++ b/Sources/xcproj/PBXBuildRule.swift
@@ -1,0 +1,136 @@
+import Foundation
+
+/// A PBXBuildRule is used to specify a method for transforming an input file in to an output file(s).
+final public class PBXBuildRule: PBXObject, Equatable {
+
+    // MARK: - Attributes
+
+    /// Element compiler spec.
+    public var compilerSpec: String
+
+    /// Element file patterns.
+    public var filePatterns: String?
+
+    /// Element file type.
+    public var fileType: String
+
+    /// Element is editable.
+    public var isEditable: Bool
+
+    /// Element name.
+    public var name: String?
+
+    /// Element output files.
+    public var outputFiles: [String]
+
+    /// Element output files compiler flags.
+    public var outputFilesCompilerFlags: [String]?
+
+    /// Element script.
+    public var script: String?
+
+    // MARK: - Init
+
+    public init(reference: String,
+                compilerSpec: String,
+                filePatterns: String? = nil,
+                fileType: String,
+                isEditable: Bool = true,
+                name: String? = nil,
+                outputFiles: [String] = [],
+                outputFilesCompilerFlags: [String]? = nil,
+                script: String? = nil) {
+        self.compilerSpec = compilerSpec
+        self.filePatterns = filePatterns
+        self.fileType = fileType
+        self.isEditable = isEditable
+        self.name = name
+        self.outputFiles = outputFiles
+        self.outputFilesCompilerFlags = outputFilesCompilerFlags
+        self.script = script
+        super.init(reference: reference)
+    }
+
+    // MARK: - Decodable
+
+    enum CodingKeys: String, CodingKey {
+        case compilerSpec
+        case filePatterns
+        case fileType
+        case isEditable
+        case name
+        case outputFiles
+        case outputFilesCompilerFlags
+        case script
+    }
+
+    public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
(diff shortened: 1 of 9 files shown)


Sources/xcproj/PBXBuildRule.swift
```diff
@@ -0,0 +1,134 @@
...
+    public var compilerSpec: String
+
+    /// Element file patterns.
+    public var filePatterns: String?
+
+    /// Element file type.
+    public var fileType: String
+
+    /// Element is editable.
+    public var isEditable: UInt
```
yonaskolb: Can this take any other values than 1 or 0? Otherwise we can make it a Bool

Sources/xcproj/PBXBuildRule.swift
```diff
@@ -0,0 +1,134 @@
...
+    public var name: String?
+
+    /// Element output files.
+    public var outputFiles: [String]
+
+    /// Element output files compiler flags.
+    public var outputFilesCompilerFlags: [String]?
+
+    /// Element script.
+    public var script: String
```
yonaskolb: Script is also optional right? Is there an advantage to making it non optional?

Sources/xcproj/PBXBuildRule.swift
briantkelley: I believe it's only ever a 1 or 0. There were some existing seemingly Bool properties were using (U)Int but definitely happy to change it.

Sources/xcproj/PBXBuildRule.swift
briantkelley: Nice catch. This should definitely be optional.

briantkelley:
Thanks for the feedback @yonaskolb. Updated the PR.

Sources/xcproj/PBXBuildRule.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Element script.
+    public var script: String?
+
+    // MARK: - Init
+
+    public init(reference: String,
+                compilerSpec: String,
+                filePatterns: String? = nil,
+                fileType: String,
```
yonaskolb: One more very minor thing. As `fileType` is non optional (I don't know if it should be or not), can we move it before optional `filePatterns`?

Sources/xcproj/PBXBuildRule.swift
yonaskolb: Yeah, those other ones could be changed in 2.0 as it's a breaking change

yonaskolb: https://github.com/xcodeswift/xcproj/issues/204

Sources/xcproj/PBXBuildRule.swift
```diff
@@ -0,0 +1,136 @@
...
+/// A PBXBuildRule is used to specify a method for transforming an input file in to an output file(s).
+final public class PBXBuildRule: PBXObject, Equatable {
+
+    // MARK: - Attributes
+
+    /// Element compiler spec.
+    public var compilerSpec: String
+
+    /// Element file patterns.
+    public var filePatterns: String?
```
pepibumur: Out of curiosity. Did you check the optionality of the attributes with Xcode?

briantkelley: Thanks for fixing this up and merging. @pepibumur, yes, I tried examining the encoded form of of a number of different permutations. `compilerSpec` is required so specify how the file should be processed, but `filePatterns` is optional since it's a filter on `fileType`.

pepibumur: No problem @briantkelley 😛  it was just curiosity. The changes should be in xcproj 1.8.0 ready to be used :)

PR #198, opened 2017-12-20
Allow custom Source Tree Specifications

briantkelley:
### Short description 📝
An esoteric feature of Xcode allows people to specify a custom variable for the source tree (e.g. `DERIVED_FILE_DIR`), which wasn't supported by xcproj.

### Solution 📦
 Convert `PBXSourceTree` to an enum with an associated value so we can load this information.

With this change, loading a `PBXSourceTree` from a string always succeeds, which identified a number of obsolete tests that were attempting to check for a throw when a value was missing from the object's encoded representation. I removed those tests.

### Implementation 👩‍💻👨‍💻
- [x] Convert `PBXSourceTree` to manually initialize from a string and convert to a string
- [x] Update `PBXSourceTree` to manually determine equality. Perhaps this is another candidate for Sourcery.
- [x] Add support decoding and encoding custom PBXSourceTree values.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/198)

### DIFF
--- a/Sources/xcproj/PBXSourceTree.swift
+++ b/Sources/xcproj/PBXSourceTree.swift
@@ -2,14 +2,96 @@ import Foundation
 
 /// Specifies source trees for files
 /// Corresponds to the "Location" dropdown in Xcode's File Inspector
-public enum PBXSourceTree: String, Hashable, Decodable {
-    case none = ""
-    case absolute = "<absolute>"
-    case group = "<group>"
-    case sourceRoot = "SOURCE_ROOT"
-    case buildProductsDir = "BUILT_PRODUCTS_DIR"
-    case sdkRoot = "SDKROOT"
-    case developerDir = "DEVELOPER_DIR"
+public enum PBXSourceTree: CustomStringConvertible, Equatable, Decodable {
+
+    case none
+    case absolute
+    case group
+    case sourceRoot
+    case buildProductsDir
+    case sdkRoot
+    case developerDir
+    case custom(String)
+
+    private static let noneValue = ""
+    private static let absoluteValue = "<absolute>"
+    private static let groupValue = "<group>"
+    private static let sourceRootValue = "SOURCE_ROOT"
+    private static let buildProductsDirValue = "BUILT_PRODUCTS_DIR"
+    private static let sdkRootValue = "SDKROOT"
+    private static let developerDirValue = "DEVELOPER_DIR"
+
+    public init(value: String) {
+        switch value {
+            case PBXSourceTree.noneValue:
+                self = .none
+            case PBXSourceTree.absoluteValue:
+                self = .absolute
+            case PBXSourceTree.groupValue:
+                self = .group
+            case PBXSourceTree.sourceRootValue:
+                self = .sourceRoot
+            case PBXSourceTree.buildProductsDirValue:
+                self = .buildProductsDir
+            case PBXSourceTree.sdkRootValue:
+                self = .sdkRoot
+            case PBXSourceTree.developerDirValue:
+                self = .developerDir
+            default:
+                self = .custom(value)
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


Sources/xcproj/CommentedString.swift
```diff
@@ -32,13 +32,21 @@ struct CommentedString {
 
         var escaped = string
         // escape escape
-        escaped = escaped.replacingOccurrences(of: "\\", with: "\\\\")
+        if escaped.contains("\\" as Character) {
```
pepibumur: Isn't this part of another PR?

Sources/xcproj/PBXProductType.swift
```diff
@@ -21,6 +21,7 @@ public enum PBXProductType: String, Decodable {
     case stickerPack = "com.apple.product-type.app-extension.messages-sticker-pack"
     case xpcService = "com.apple.product-type.xpc-service"
     case ocUnitTestBundle = "com.apple.product-type.bundle.ocunit-test"
+    case xcodeExtension = "com.apple.product-type.xcode-extension"
```
pepibumur: I think somehow you got in this branch changes that were merged into master. You should branch off from release/2.0.0 rather than master.

Sources/xcproj/PBXSourceTree.swift
```diff
@@ -2,22 +2,100 @@ import Foundation
...
+                 (.sourceRoot, .sourceRoot),
+                 (.buildProductsDir, .buildProductsDir),
+                 (.sdkRoot, .sdkRoot),
+                 (.developerDir, .developerDir):
+                return true
+
+            case (.custom(let lhsValue), .custom(let rhsValue)):
+                return lhsValue == rhsValue
+
+            case (.none, _),
```
pepibumur: Did you use this case instead of `default` to make sure we don't forget any type?

Sources/xcproj/PBXSourceTree.swift
```diff
@@ -2,22 +2,100 @@ import Foundation
...
+                 (.sourceRoot, _),
+                 (.buildProductsDir, _),
+                 (.sdkRoot, _),
+                 (.developerDir, _),
+                 (.custom, _):
+                return false
+        }
+    }
+
+    public var hashValue: Int {
```
pepibumur: Do we need the `Hashable` conformance? Unless you need it I wouldn't add it for now. We have plans to automate the conformance using Sourcery.

Sources/xcproj/PBXSourceTree.swift
```diff
@@ -2,22 +2,100 @@ import Foundation
...
+        return description.hashValue
+    }
+
+    public var description: String {
+        switch self {
+            case .none:
+                return ""
+            case .absolute:
+                return "<absolute>"
```
pepibumur: Since these values are defined in the `init` I think we should have them in a single place, for example as constants in the enum.

Sources/xcproj/CommentedString.swift
briantkelley: Yup. I made the change in master so when targeting 2.0.0 it wants to merge everything. I have that fixed and will push soon.

Sources/xcproj/PBXSourceTree.swift
briantkelley: Yup. I'm a big fan of the "switch is not exhaustive" compile error 😁

Sources/xcproj/PBXSourceTree.swift
briantkelley: Nope, don't need `Hashable`, just an implementation of `Equatable` for now. Fixed.

briantkelley:
@pepibumur Any more feedback on the updated code?

pepibumur:
Nothing to add! Great job @briantkelley

PR #208, opened 2017-12-27
Change CLI fileExtension to nil

keith:
The fileExtension property is used upstream to determine the
explicitFileType of the file reference. If that property exists but is
empty Xcode crashes. Previously that was the case with the tool file
type since the product target had an empty explicitFileType. With this
change it now omits that property instead, and still has no extension.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/208)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Fix Xcode 9.2 warning https://github.com/xcodeswift/xcproj/pull/209 by @keith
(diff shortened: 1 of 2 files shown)


keith:
There's another problem with this that this PR does not solve. CLI target's explicitFileType _should_ be set to `compiled.mach-o.executable`, but as long as this is coupled to the fileExtension, that is a problem.

PR #210, opened 2018-01-02
Add XCWorkspace group and file reference parsing

allu22:
### Short description 📝
Currently `XCWorkspace.Data` only supports `FileRef` elements from `workspacedata` file. And in some cases is incorrectly parses `Group` element as `FileRef`.

### Solution 📦
Add support for both `Group` and `FileRef` elements in `XCWorkspace.Data`.

I moved `XCWorkspace.Data` to `XCWorkspaceData`, because I needed to add additional models _under_ `Data`. If I were to keep nesting them, it would become less readable (If you think otherwise, let me know.)

Currently all `XCWorkspaceData` encoding is done in `XCWorkspaceData.swift` by adding private extensions for other necessary types. This is done to not expose `AEXML` types outside of this file. I am not sure if this is necessary, but decided this way because I personally don't see need to initialise these models from `AEXMLElement`.

One thing I am not sure about is `XCWorkspaceDataElement` enum. Alternative would be to use a protocol but then we would need to downcast when we want concrete types.

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/210)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - macOS CLI targets now have a nil extension, instead of an empty string https://github.com/xcodeswift/xcproj/pull/208 by @keith
 - Fix unnecessary quotations in CommentedString https://github.com/xcodeswift/xcproj/pull/211 by @allu22
 
+### Changed
+- **Breaking:** `XCWorkspace.Data` renamed to `XCWorkspaceData` and removed `references`.
+
 ## 2.0.0
 
 ### Added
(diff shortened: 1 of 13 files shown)


yonaskolb:
You'll need to regenerate the carthage project due to the new files

allu22:
Fixed the carthage project.

pepibumur:
Sorry I couldn't review this on time. Great addition @allu22. My only concern regarding this change is that it's breaking and that will require a major bump. Are you ok with introducing these major changes that often or do you think we should avoid that and avoid doing that many major releases? cc @yonaskolb

allu22:
If we are following semver strictly then yes, this should go into major release.

PR #211, opened 2018-01-07
Improve character set for quotation

### LINKED ISSUE #205: change the organization name unexpectly
i can't find the Organization Name setting in xcproj, what should i do execpt rename `ORGANIZATIONNAME` in `project.pbxproj`

image

allu22:
Resolves #205

### Short description 📝
Current implementation does not escape string where Xcode does. As can be seen by #205 

### Solution 📦
I created list of build settings with values ranging from unicode character point 20 to 255. After saving it with Xcode 9.2 (9C40b), i noticed that there seems to be limited list of characters that are not wrapped in quotation.

Based on that I constructed new list of invalid characters.

TLDR: List of character that are not surrounded by quotes: `./_$` `0-9` `a-z` `A-Z`

### GIF
gif

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/211)


Tests/xcprojTests/PBXProjWriterSpec.swift
```diff
@@ -49,7 +49,7 @@ class PBXProjEncoderSpec: XCTestCase {
             ">": ">".quoted,
             ";": ";".quoted,
             "&": "&".quoted,
-            "$": "$".quoted,
+            "$": "$",
```
yonaskolb: Could you add the example text from #205

Sources/xcproj/CommentedString.swift
```diff
@@ -19,8 +19,14 @@ struct CommentedString {
         self.comment = comment
     }
 
-    private static let invalidCharacters = CharacterSet.alphanumerics.inverted
-        .subtracting(CharacterSet(charactersIn: "_./"))
+    private static var invalidCharacters: CharacterSet = {
+        var invalidSet = CharacterSet(charactersIn: "_$")
```
yonaskolb: It's interesting the _ and $ get special treatment. It must be because Xcode uses these extensively in build settings

Tests/xcprojTests/PBXProjWriterSpec.swift
allu22: Done

pepibumur:
Great fix @allu22. We've got a few bugs related this since people started using it. Your implementation seems to be aligned better with the Xcode's behavior. Fantastic job!

PR #212, opened 2018-01-09
Improved README examples

ilyapuchka:
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/212)

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,24 +156,65 @@ try project.write(path: "MyApp.xcodeproj")
 
 ```swift
 guard var sourcesGroup = project.pbxproj.objects.groups.first(where: {$0.value.name == "Sources" || $0.value.path == "Sources"})?.value else { return }    
+let homeGroup = PBXGroup(children: [], sourceTree: .group, path: "Home")
+let groupRef = pbxproj.objects.generateReference(homeGroup, "Home")
+sourcesGroup.children.append(homeGroup, reference: groupRef)
+project.pbxproj.objects.addObject(groupRef)
+```
+
+<details>
+<summary>Versions <2.0</summary>
+
+```swift
+guard var sourcesGroup = project.pbxproj.objects.groups.first(where: {$0.value.name == "Sources" || $0.value.path == "Sources"})?.value else { return }    
 let homeGroup = PBXGroup(reference: project.pbxproj.generateUUID(for: PBXGroup.self), children: [], sourceTree: .group, path: "Home")
 sourcesGroup.children.append(homeGroup.reference)
 project.pbxproj.objects.addObject(homeGroup)
 ```
 
+</details>
+
 #### Add `HomeViewController.swift` file inside `HomeGroup`
 
 ```swift
-let homeGroup = PBXGroup(reference: "xxx", children: [], sourceTree: .group, path: "Home")
-let homeViewController = PBXFileReference(reference: project.pbxproj.generateUUID(for: PBXFileReference.self), sourceTree: .group, name: "HomeViewController.swift", path: "HomeViewController.swift")
-homeGroup.children.append(homeViewController.reference)
-pbxproj.objects.addObject(homeViewController)
+let homeViewController = PBXFileReference(sourceTree: .group, name: "HomeViewController.swift", path: "HomeViewController.swift")
+let fileRef = pbxproj.objects.generateReference(homeViewController, "HomeViewController.swift")
+homeGroup.children.append(fileRef)
+project.pbxproj.objects.addObject(homeViewController, reference: fileRef)
 ```
 
-#### Add `HomeViewController.swift` file to `MyApp` target
+<details>
+<summary>Versions <2.0</summary>
 
 ```swift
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -155,8 +155,8 @@ try project.write(path: "MyApp.xcodeproj")
 #### Adding `Home` group inside `Sources` group
 
 ```swift
-guard var sourcesGroup = project.pbxproj.objects.groups.first(where: {$0.value.name == "Sources"})?.value else { return }    
-let homeGroup = PBXGroup(reference: "xxx", children: [], sourceTree: .group, path: "Home")
+guard var sourcesGroup = project.pbxproj.objects.groups.first(where: {$0.value.name == "Sources" || $0.value.path == "Sources"})?.value else { return }
```
ilyapuchka: In Xcode 9 groups are created with paths by default, unless they are added via "New Group without Folder". So `name` property is not stored in project xml and then is not decoded.

But before Xcode 9 when creating folder with File-New-Group menu (instead of dragging folder into the project) only `name` property will be set and `path` will be nil.

README.md
```diff
@@ -155,8 +155,8 @@ try project.write(path: "MyApp.xcodeproj")
 #### Adding `Home` group inside `Sources` group
 
 ```swift
-guard var sourcesGroup = project.pbxproj.objects.groups.first(where: {$0.value.name == "Sources"})?.value else { return }    
-let homeGroup = PBXGroup(reference: "xxx", children: [], sourceTree: .group, path: "Home")
+guard var sourcesGroup = project.pbxproj.objects.groups.first(where: {$0.value.name == "Sources" || $0.value.path == "Sources"})?.value else { return }    
+let homeGroup = PBXGroup(reference: project.pbxproj.generateUUID(for: PBXGroup.self), children: [], sourceTree: .group, path: "Home")
```
ilyapuchka: I've spent some time searching for a way to generate references until found this method.

README.md
```diff
@@ -165,14 +165,15 @@ project.pbxproj.objects.addObject(homeGroup)
 
 ```swift
 let homeGroup = PBXGroup(reference: "xxx", children: [], sourceTree: .group, path: "Home")
-let homeViewController = PBXFileReference(reference: "xxx", sourceTree: .group, path: "HomeViewController.swift")
+let homeViewController = PBXFileReference(reference: project.pbxproj.generateUUID(for: PBXFileReference.self), sourceTree: .group, name: "HomeViewController.swift", path: "HomeViewController.swift")
 homeGroup.children.append(homeViewController.reference)
+pbxproj.objects.addObject(homeViewController)
```
ilyapuchka: Without this file will be not properly added and build file will point to a dangling reference.

README.md
```diff
@@ -165,14 +165,15 @@ project.pbxproj.objects.addObject(homeGroup)
...
 ```
 
 #### Add `HomeViewController.swift` file to `MyApp` target
 
 ```swift
-let homeViewController = PBXFileReference(reference: "xxx", sourceTree: .group, path: "HomeViewController.swift")
+let homeViewController = PBXFileReference(reference: project.pbxproj.generateUUID(for: PBXFileReference.self), sourceTree: .group, name: "HomeViewController.swift", path: "HomeViewController.swift")
```
ilyapuchka: Name and path of the file may be different, i.e. when file is added with absolute path which is not in a group's folder

README.md
pepibumur: Thanks for the clarification. Makes sense to me! 👍

README.md
pepibumur: Yeah we introduced it in the version 2.0 but we forgot to update the readme. The function is `generateReference` instead of `generateUUID` and it should be used like this:

```
project.pbxproj.generateReference(PBXGroup.self, "home")
```

The second argument must be a string that uniquely identifies the object. In this case we can just use "home".

README.md
```diff
@@ -165,14 +165,15 @@ project.pbxproj.objects.addObject(homeGroup)
 
 ```swift
 let homeGroup = PBXGroup(reference: "xxx", children: [], sourceTree: .group, path: "Home")
-let homeViewController = PBXFileReference(reference: "xxx", sourceTree: .group, path: "HomeViewController.swift")
+let homeViewController = PBXFileReference(reference: project.pbxproj.generateUUID(for: PBXFileReference.self), sourceTree: .group, name: "HomeViewController.swift", path: "HomeViewController.swift")
```
pepibumur: Same comment regarding the `generateUUID`

README.md
pepibumur: Good catch!

README.md
pepibumur: Same comment regarding the reference generation

README.md
ilyapuchka: Yeah, looks like I was using 1.7.0 version

ilyapuchka: Unfortunately it means that readme examples are completely out of date. I would suggest to keep section with examples for 1.x version and with 2.x

pepibumur: Thank you very much @ilyapuchka

Issue #40, opened 2017-08-09
Fully qualified Names of source files


olbrichj:
Hey,

sorry if this is the wrong place to ask, but I can't find anything anywhere else.

Is there an option to get the fully qualified name of the sourcefiles? In project.pbxproj.objects I can only find the direct filename.

Thanks

yonaskolb:
Do you mean the path to the file? If so you would have to traverse the groups it exists in, and build up the path that way. 
That would be a good helper method, we should add it

olbrichj:
Thanks :)
Yes I'm referring to the path to the file. Either starting at root of the project, or even better at root of the system /Users/xx/project/.../main.swift.
I'll check if I figure out how to traverse it. Currently I'm struggling with the documentation.

olbrichj:
I've written a small example. Not sure if I got everything. Is there an option to separate by target?

let project = try! XcodeProj(path: "Demo.xcodeproj")
let pbxproj = project.pbxproj
let buildGroups = pbxproj.objects.groups

for group in buildGroups {
    if let path = group.path {
        for child in group.children {
            if let found = pbxproj.objects.index(where: { $0.reference == child }) {
                var name = ""
                switch pbxproj.objects[found] {
                case .pbxFileReference(let file): name = file.path!
                default:
                    break
                }
                print(path + "/" + name)
            }
        }
    }
}

yonaskolb:
- instead of `pbxproj.objects.index(where:` you could use `pbxproj.objects.first(where:`
- you'd have to do this recursively to handle nested groups, and build up the path along the way
- you'd have to handle the path differently base on the group `sourceTree`
- if you want to filter by target, you'd have to check if there was a build file for a certain target

What's your use case? Maybe there's a different way to achieve what you're after

olbrichj:
Thanks for your input. 
Currently I'm trying to write some kind of IDE for iOS/OSX development. To get a list of all source files I'd love to import the Xcode.proj. This way I would know, which files are actually used and not just part of the structure.

pepibumur:
This shouldn't be closed. Reopening.

ilyapuchka:
I'm using this code to get full path of file element:

```swift
extension PBXProj.Objects {
    func fullPath(fileElement: PBXFileElement, reference: String, sourceRoot: Path) -> Path? {
        switch fileElement.sourceTree {
        case .absolute?:
            return fileElement.path.flatMap({ Path($0) })
        case .sourceRoot?:
[...]

PR #213, opened 2018-01-09
Helper methods required for adding build file to the target

ilyapuchka:
Resolves https://github.com/xcodeswift/xcproj/issues/120, #40 

### Short description 📝
Added some helper methods required for adding build files

### Solution 📦
With these methods adding a new build file requires just few calls:

```swift
// 1. get target
guard let target = project.pbxproj.objects.targets(named: "MyApp").first else { return }

// 2. if needed add new group for this file
let groups = project.pbxproj.objects.addGroup(named: "Sources/Home", to: project.pbxproj.rootGroup)

// 3. add file to the project, if needed
let (fileRef, _) = project.pbxproj.objects.addFile(at: "Sources/HomeViewController.swift")

// 4. add file to the group, if needed
project.pbxproj.objects.addFile(toGroup: groups[1].group, reference: fileRef)

// 5. add file to the target, if needed
project.pbxproj.objects.addBuildFile(toTarget: target, reference: fileRef)
```

If approved for merge I'll add some integration tests.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/213)

### DIFF
--- a/Sources/xcproj/PBXProj+Helpers.swift
+++ b/Sources/xcproj/PBXProj+Helpers.swift
@@ -1,6 +1,18 @@
 import Foundation
 import PathKit
 
+// MARK: - PBXProj Extension (Public)
+
+public extension PBXProj {
+
+    /// Returns project's root group
+    public var rootGroup: PBXGroup {
+        let groupRef = objects.projects[rootObject]!.mainGroup
+        return objects.groups[groupRef]!
+    }
+
+}
+
 // MARK: - PBXProj Extension (Getters)
 
 extension PBXProj {
--- a/Sources/xcproj/PBXProj+Helpers.swift
+++ b/Sources/xcproj/PBXProj+Helpers.swift
@@ -5,10 +5,20 @@ import PathKit
 
 public extension PBXProj {
 
-    /// Returns project's root group
+    /// Returns root project.
+    public var rootProject: PBXProject? {
+        return objects.projects[rootObject]
+    }
+
+    /// Returns root project's root group.
     public var rootGroup: PBXGroup {
-        let groupRef = objects.projects[rootObject]!.mainGroup
-        return objects.groups[groupRef]!
+        guard let rootProject = self.rootProject else {
+            fatalError("Missing root project")
+        }
+        guard let rootGroup = objects.groups[rootProject.mainGroup] else {
+            fatalError("Root project has no root group")
+        }
+        return rootGroup
     }
 
 }
--- a/Sources/xcproj/PBXProjObjects+Helpers.swift
+++ b/Sources/xcproj/PBXProjObjects+Helpers.swift
@@ -93,7 +93,12 @@ public extension PBXProj.Objects {
     ///   - sourceTree: file sourceTree, default is `.group`
     ///   - sourceRoot: path to project's source root.
     /// - Returns: new or existing file and its reference.
-    public func addFile(at filePath: Path, toGroup: PBXGroup, sourceTree: PBXSourceTree = .group, sourceRoot: Path) throws -> (reference: String, file: PBXFileReference) {
+    public func addFile(
+        at filePath: Path,
+        toGroup: PBXGroup,
+        sourceTree: PBXSourceTree = .group,
+        sourceRoot: Path) throws -> (reference: String, file: PBXFileReference) {
+
         guard filePath.isFile else {
             throw XCodeProjEditingError.notAFile(path: filePath)
         }
--- a/Sources/xcproj/PBXProjObjects+Helpers.swift
+++ b/Sources/xcproj/PBXProjObjects+Helpers.swift
@@ -50,86 +50,60 @@ public extension PBXProj.Objects {
     /// Adds new group with the give name to the given parent group.
     /// Group name can be a path with components separated by `/`.
     /// Will create new groups for intermediate path components or use existing groups.
-    /// Returns new group and its reference.
+    /// Returns all new or existing groups in the path and their references.
     ///
     /// - Parameters:
     ///   - groupName: group name, can be a path with components separated by `/`
-    ///   - children: group children, default is empty array.
-    ///   - sourceTree: group source tree, default is `.group`.
-    ///   - name: group name.
-    ///   - path: group path.
-    ///   - usesTabs: group uses tabs.
+    ///   - toGroup: parent group
     ///   - options: additional options, default is empty set.
-    /// - Returns: new group and its reference.
-    public func addGroup(
-        named groupName: String,
-        to toGroup: PBXGroup,
-        children: [String] = [],
-        sourceTree: PBXSourceTree = .group,
-        usesTabs: Int? = nil,
-        indentWidth: Int? = nil,
-        tabWidth: Int? = nil,
-        options: GroupAddingOptions = []) -> (String, PBXGroup) {
+    /// - Returns: all new or existing groups in the path and their references.
+    public func addGroup(named groupName: String, to toGroup: PBXGroup, options: GroupAddingOptions = []) -> [(String, PBXGroup)] {
+        return addGroups(groupName.components(separatedBy: "/"), to: toGroup, options: options)
+    }
 
-        var toGroup = toGroup
-        var newGroup: PBXGroup!
-        var reference: String!
-        for groupName in groupName.components(separatedBy: "/") {
-            if let existingGroup = group(named: groupName, inGroup: toGroup) {
-                reference = existingGroup.0
-                newGroup = existingGroup.1
-            } else {
-                newGroup = PBXGroup(
-                    children: children,
-                    sourceTree: .group,
-                    name: groupName,
-                    path: options.contains(.withoutFolder) ? nil : groupName,
-                    usesTabs: usesTabs,
-                    indentWidth: indentWidth,
-                    tabWidth: tabWidth
-                )
-                reference = generateReference(newGroup, groupName)
-                addObject(newGroup, reference: reference)
-                toGroup.children.append(reference)
-            }
-            toGroup = newGroup
+    private func addGroups(_ groupNames: [String], to toGroup: PBXGroup, options: GroupAddingOptions) -> [(String, PBXGroup)] {
+        guard !groupNames.isEmpty else { return [] }
+        let newGroup = createOrGetGroup(named: groupNames[0], in: toGroup, options: options)
+        return [newGroup] + addGroups(Array(groupNames.dropFirst()), to: newGroup.1, options: options)
+    }
+
+    private func createOrGetGroup(named groupName: String, in parentGroup: PBXGroup, options: GroupAddingOptions) -> (String, PBXGroup) {
+        if let existingGroup = group(named: groupName, inGroup: parentGroup) {
+            return existingGroup
         }
+
+        let newGroup = PBXGroup(
+            children: [],
+            sourceTree: .group,
+            name: groupName,
+            path: options.contains(.withoutFolder) ? nil : groupName
+        )
+        let reference = generateReference(newGroup, groupName)
+        addObject(newGroup, reference: reference)
+        parentGroup.children.append(reference)
         return (reference, newGroup)
     }
 
-    /// Adds file at the give path to the project. Returns new file and its reference.
+    /// Adds file at the give path to the project or returns existing file and its reference.
     ///
     /// - Parameters:
     ///   - filePath: path to the file
     ///   - sourceTree: file source tree, default is `.group`.
     ///   - name: file name, by default gets file name from the path
-    /// - Returns: new file and its reference.
-    public func addFile(at filePath: Path,
-        sourceTree: PBXSourceTree = .group,
-        name: String? = nil,
-        fileEncoding: Int? = nil,
-        explicitFileType: String? = nil,
-        lastKnownFileType: String? = nil,
-        includeInIndex: Int? = nil,
-        usesTabs: Int? = nil,
-        lineEnding: Int? = nil,
-        xcLanguageSpecificationIdentifier: String? = nil
-        ) -> (String, PBXFileReference) {
+    /// - Returns: new or existing file and its reference.
+    public func addFile(at filePath: Path) throws -> (String, PBXFileReference) {
+        guard filePath.isFile else { throw XCodeProjEditingError.notAFile(path: filePath) }
 
         if let existingFileReference = fileReferences.first(where: { $0.value.path == filePath.string }) {
             return existingFileReference
         }
 
         let fileReference = PBXFileReference(
-            sourceTree: sourceTree,
-            name: name ?? filePath.lastComponent,
-            fileEncoding: fileEncoding,
-            explicitFileType: explicitFileType ?? PBXFileReference.fileType(path: filePath),
-            lastKnownFileType: lastKnownFileType ?? PBXFileReference.fileType(path: filePath),
-            path: filePath.string,
-            includeInIndex: includeInIndex,
-            usesTabs: usesTabs,
-            xcLanguageSpecificationIdentifier: xcLanguageSpecificationIdentifier
+            sourceTree: .group,
+            name: filePath.lastComponent,
+            explicitFileType: PBXFileReference.fileType(path: filePath),
+            lastKnownFileType: PBXFileReference.fileType(path: filePath),
+            path: filePath.string
         )
         let reference = generateReference(fileReference, filePath.string)
         addObject(fileReference, reference: reference)
@@ -146,19 +120,23 @@ public extension PBXProj.Objects {
         group.children.append(reference)
     }
 
-    /// Adds file to the given target's sources build phase
-    /// If target's sources build phase can't be found or it already contains file with given reference method does nothing.
+    /// Adds file to the given target's sources build phase or returns existing build file and its reference.
+    /// If target's sources build phase can't be found returns nil.
     ///
     /// - Parameter target: target obejct
     /// - Parameter reference: file reference
-    public func addBuildFile(toTarget target: PBXTarget, reference: String) {
-        guard let sourcesBuildPhase = sourcesBuildPhase(target: target) else { return }
-        guard !buildFiles.contains(where: { $0.value.fileRef == reference }) else { return }
+    /// - Returns: new or existing build file and its reference
+    public func addBuildFile(toTarget target: PBXTarget, reference: String) -> (String, PBXBuildFile)? {
+        guard let sourcesBuildPhase = sourcesBuildPhase(target: target) else { return nil }
+        if let existingBuildFile = buildFiles.first(where: { $0.value.fileRef == reference }) {
+            return existingBuildFile
+        }
 
         let buildFile = PBXBuildFile(fileRef: reference)
-        let buildFileRef = generateReference(buildFile, reference)
-        addObject(buildFile, reference: buildFileRef)
-        sourcesBuildPhase.files.append(buildFileRef)
+        let reference = generateReference(buildFile, reference)
+        addObject(buildFile, reference: reference)
+        sourcesBuildPhase.files.append(reference)
+        return (reference, buildFile)
     }
 
 }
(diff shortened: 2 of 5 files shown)


Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -1,6 +1,18 @@
 import Foundation
 import PathKit
 
+// MARK: - PBXProj Extension (Public)
+
+public extension PBXProj {
+
+    /// Returns project's root group
+    public var rootGroup: PBXGroup {
+        let groupRef = objects.projects[rootObject]!.mainGroup
```
pepibumur: I'd make this property optional and remove the force unwraps from this getter

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,160 @@ public extension PBXProj.Objects {
         targets.append(contentsOf: Array(aggregateTargets.values) as [PBXTarget])
         return targets.filter { $0.name == name }
     }
-    
+
+    /// Retruns target's sources build phase.
+    ///
+    /// - Parameter target: target object.
+    /// - Returns: target's sources build phase, if found.
+    public func sourcesBuildPhase(target: PBXTarget) -> PBXSourcesBuildPhase? {
```
pepibumur: Nice additions @ilyapuchka 👏

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,160 @@ public extension PBXProj.Objects {
...
+    ///   - groupName: group name, can be a path with components separated by `/`
+    ///   - children: group children, default is empty array.
+    ///   - sourceTree: group source tree, default is `.group`.
+    ///   - name: group name.
+    ///   - path: group path.
+    ///   - usesTabs: group uses tabs.
+    ///   - options: additional options, default is empty set.
+    /// - Returns: new group and its reference.
+    public func addGroup(
+        named groupName: String,
```
pepibumur: The problem that I see exposing this method with all these arguments is that if any property is introduced in the `PBXGroup` class, we need to update this method to support it. Given that xcproj represents the objects with mutable classes what do you think about simplifying it:

```swift
public func addGroup(named groupName: String,
  to toGroup: PBXGroup, 
  withFolder: Bool = false) -> [(String, PBXGroup)]
```
Notice that the method returns an array with all the groups that got created. If the developer passes a path like this `/group_a/group_b/group_c` then it'd return `[GroupA, GroupB, GroupC]` corresponding the last element in the array to the last path component. If the developer is interested in updating let's say, the `usesTabs` property of all the created groups, it can be easily done iterating with a `forEach` on the returned array.

On the other side, I think the implementation could be simplified adding some recursion:

```
// Given the path /component_a/component_b/component_c

1. Generate/Get the first component group: GroupA
2. Concatenate the group with the result of calling the method with GroupA and name /component_b/component_c: [GroupA, result[0], result[1], ...]
```

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,160 @@ public extension PBXProj.Objects {
...
+        guard !group.children.contains(reference) else { return }
+        group.children.append(reference)
+    }
+
+    /// Adds file to the given target's sources build phase
+    /// If target's sources build phase can't be found or it already contains file with given reference method does nothing.
+    ///
+    /// - Parameter target: target obejct
+    /// - Parameter reference: file reference
+    public func addBuildFile(toTarget target: PBXTarget, reference: String) {
```
pepibumur: What do you think about returning the `PBXBuildFile` here?

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,160 @@ public extension PBXProj.Objects {
...
+        fileEncoding: Int? = nil,
+        explicitFileType: String? = nil,
+        lastKnownFileType: String? = nil,
+        includeInIndex: Int? = nil,
+        usesTabs: Int? = nil,
+        lineEnding: Int? = nil,
+        xcLanguageSpecificationIdentifier: String? = nil
+        ) -> (String, PBXFileReference) {
+
+        if let existingFileReference = fileReferences.first(where: { $0.value.path == filePath.string }) {
```
pepibumur: I think this method does a lot of assumptions. Let's say we have the following file structure:

```
group_a/
   group_b/
     file_a
```
Where file_a is represented as a `PBXFileReference` with name `file_a` and sourceTree `.group`. `file_a` is a child of the group b. 

Now you call this method saying that you want to add a file with path `/group_a/group_b/file_a` which already exists in our project. We won't get any existing reference when we actually should *(because there's already a reference for that file)*. Moreover, if the path is relative the group, the group where reference gets added to as a child matters *(and the developer might not know about it)*.

I'd aim for an implementation similar to the one above for groups. You'd call a method like this:

```swift
func addFile(named fileName: String, to toGroup: PBXGroup)
```
And it'd recursively generate the groups and file references and return them.

What do you think?

Sources/xcproj/PBXProj+Helpers.swift
ilyapuchka: I was considering that of course but it will lead to always force unwrapping or checking for nil when using this API, which will make it inconvenient. Is it possible that a project can have no rootObject or group with such reference? If so probably it's a malformed project, so failing should be expected in this case or this should be validated even earlier. As a user I would prefer convenience over safety in this particular case.

Sources/xcproj/PBXProjObjects+Helpers.swift
ilyapuchka: Right, I didn't notice that those types are mutable classes, so it will makes sense to remove all those parameters. And I like your idea about returning array.

Sources/xcproj/PBXProjObjects+Helpers.swift
ilyapuchka: I'll return file and reference to match other APIs

Sources/xcproj/PBXProjObjects+Helpers.swift
ilyapuchka: This falls into the problem of calculating file paths, indeed path property can be relative in this case and so will not match the absolute path in the predicate. Maybe there is another way to do such check that I'm missing?

Regarding your suggestion to create groups and files together in this method I was doing that originally, but decided to break it into separate methods not to violate single responsibility. And it will hide details of project structure from developer that you mentioned in the issue discussion saving only one call for the user. So I would keep it as it is, just remove additional parameters as you suggested for another method.

Sources/xcproj/PBXProj+Helpers.swift
allu22: I also don't like forced unwapping. It can be legitimately be missing when you are constructing project file from scratch. In that case it would be bad behaviour to crash in runtime. Instead this should be either optional variable or a function that throws an error.

Sources/xcproj/PBXProjObjects+Helpers.swift
allu22: `sourcesBuildPhases` is deprecated. Instead you should use `objects.sourcesBuildPhases`.

Sources/xcproj/PBXProjObjects+Helpers.swift
allu22: As Xcode allows you to use `/` in group name, i think this should not be used as separator.

Sources/xcproj/PBXProjObjects+Helpers.swift
ilyapuchka: this is exactly what is being used, this is extension on PBXObjects.

Sources/xcproj/PBXProjObjects+Helpers.swift
ilyapuchka: Can you show an example? At least Xcode 9 does not allow to rename groups using `/`.

Sources/xcproj/PBXProj+Helpers.swift
allu22: I would create helper for accessing `rootProject`.
``` swift
	var rootProject: PBXProject? {
		return objects.projects[rootObject]
	}
```
This would make usage of `rootObject` different. Instead you would need to access it through project: `.rootProject?.mainGroup`. But it would be more clear what is optional.

Sources/xcproj/PBXProjObjects+Helpers.swift
allu22: 👍 sorry, missed that.

Sources/xcproj/PBXProjObjects+Helpers.swift
allu22: Once you point a group to a folder with name different than group name, you can change group name, without changing the folder name. 
Example: https://imgur.com/oveZwsI

ilyapuchka: Ok. This seems to be allowed also when group is created without folder. What is the alternative then? We can require passing in array of group names instead. But then if `/` is used in one of them we should not create a group with such name if it is created with folder, as it will probably lead to invalid project. We should probably normalise each item. But if it is created without folder we should allow that.

Sources/xcproj/PBXProj+Helpers.swift
pepibumur: > Is it possible that a project can have no rootObject or group with such reference?

There are two usages of xcproj that I can think of right now:
- Opening a project and modify it.
- Create a new project.

In the first case, it's very unlikely that we force unwrap a nil value since the project already has a root group. However, in the second case, it might happen. This is a bad usage sign because the developer should have created the root group before accessing it. 

It depends on whether we want the library to crash at runtime, or execute with unexpected behavior. Thinking about a developer consuming the API, I'd prefer the crash rather than spending time debugging where the unexpected behavior might be coming from.

What do you think @allu22 ?

Sources/xcproj/PBXProjObjects+Helpers.swift
pepibumur: I think the safest way would be to compare the absolute paths but that's a bit expensive because you'd need to generate the absolute path for all the file references that you want to compare. We could use the method that you showed in another PR  for getting the absolute path and optimize it. That way your first check would be legitimate.

Regarding your point on creating files and groups I agree with you so forget what I said. I wish we could leverage the type system to prevent the developer from passing a file name such as `path/to/my/file.swift` failing at compilation time. What do you think about throwing in case the string doesn't represent a file @ilyapuchka ?

ilyapuchka: I was thinking about using this method too, but yeah, it may be too expensive.
Regarding name we can make a check check if path is file path with `path.isFile` and then throw if it is not.

Sources/xcproj/PBXProj+Helpers.swift
allu22: Example where `rootObject` `id` is set and root project does not exist until `generate` is called: 
https://github.com/yonaskolb/XcodeGen/blob/46b1512c7c4dde0621c25e4388d5e02823d3f6ab/Sources/XcodeGenKit/PBXProjGenerator.swift#L33

I think there is no right or wrong approach here. The question is, how defensive you want to be in the API. Either way is fine by me. But in case of unwrapping it here, maybe call `fatalError` with message that hints that you are missing a project object?

ilyapuchka: How about this?

```swift
    public var rootGroup: PBXGroup {
        guard let rootProject = self.rootProject else {
            fatalError("Missing root project")
        }
        guard let rootGroup = objects.groups[rootProject.mainGroup] else {
            fatalError("Root project has no root group")
        }
        return rootGroup
    }
```

ilyapuchka:
Rebased and addressed all the comments, except related to using `/` in the group names. I think we can just disallow that for now to make it simpler. It seems unlikely to me that using `/` in group names is a common practice.

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,176 @@ public extension PBXProj.Objects {
...
+
+    public var description: String {
+        switch self {
+        case .notAFile(let path):
+            return "\(path) is not a file path"
+        }
+    }
+}
+
+extension Path {
```
allu22: I think it should be made `fileprivate` or moved to separate file.

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,176 @@ public extension PBXProj.Objects {
...
+        )
+        let reference = generateReference(newGroup, groupName)
+        addObject(newGroup, reference: reference)
+        parentGroup.children.append(reference)
+        return (reference, newGroup)
+    }
+
+    /// Adds file at the give path to the project or returns existing file and its reference.
+    ///
+    /// - Parameters:
```
allu22: These parameter descriptions do not match function.

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,176 @@ public extension PBXProj.Objects {
...
+    /// - Parameter reference: file reference
+    public func addFile(toGroup group: PBXGroup, reference: String) {
+        guard !group.children.contains(reference) else { return }
+        group.children.append(reference)
+    }
+
+    /// Adds file to the given target's sources build phase or returns existing build file and its reference.
+    /// If target's sources build phase can't be found returns nil.
+    ///
+    /// - Parameter target: target obejct
```
allu22: type on work `obejct`

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,176 @@ public extension PBXProj.Objects {
...
+        }
+
+        let buildFile = PBXBuildFile(fileRef: reference)
+        let reference = generateReference(buildFile, reference)
+        addObject(buildFile, reference: reference)
+        sourcesBuildPhase.files.append(reference)
+        return (reference, buildFile)
+    }
+
+    public func fullPath(fileElement: PBXFileElement, reference: String, sourceRoot: Path) -> Path? {
```
allu22: could you add few unit tests to this function. I am not sure what cases it has to cover and how. Especially with groups.

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,176 @@ public extension PBXProj.Objects {
...
+    }
+
+    /// Adds file at the give path to the project or returns existing file and its reference.
+    ///
+    /// - Parameters:
+    ///   - filePath: path to the file
+    ///   - sourceTree: file source tree, default is `.group`.
+    ///   - name: file name, by default gets file name from the path
+    /// - Returns: new or existing file and its reference.
+    public func addFile(at filePath: Path, sourceRoot: Path) throws -> (String, PBXFileReference) {
```
allu22: Please add unit test for this function. At the moment, I have a feeling that `filePath == fullPath(fileElement: $0.value, reference: $0.key, sourceRoot: sourceRoot) ` check will fail unless `filePath` is initialized as absolute path. But it should not be a absolute path as `PBXFileReference ` is created with `sourceTree: .group `

ilyapuchka:
@allu22 done

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,182 @@ public extension PBXProj.Objects {
...
+        guard filePath.isFile else { throw XCodeProjEditingError.notAFile(path: filePath) }
+
+        if let existingFileReference = fileReferences.first(where: {
+            filePath == fullPath(fileElement: $0.value, reference: $0.key, sourceRoot: sourceRoot)
+        }) {
+            return (existingFileReference.key, existingFileReference.value)
+        }
+
+        let fileReference = PBXFileReference(
+            sourceTree: .absolute,
```
allu22: Adding absolute path is very restrictive. This will restrict the usage of this project to single path in a filesystem. It would be useful only for situations where project is always generated.

I recommend rethinking this helper function.
My suggestion:
`addFile` function will take filePath and optional PBXGroup where nil will be a fallback to mainGroup.

ilyapuchka: @allu22 indeed! I improved this method by adding group and source tree parameters. Source tree is required to calculate initial value for file path property, as it will be different for different source tree values, i.e. it can be relative to group or to project. Also added helper method to calculate relative paths. I made it public as this method can be used for example to update file path when moving it to a different group.
Some of the source tree values are not supported in file path calculation (file will be added with nil path), but we can extend it later, it feels that use cases for these values should be rare.

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -15,7 +16,224 @@ public extension PBXProj.Objects {
...
+    /// Adds file at the give path to the project and given group or returns existing file and its reference.
+    ///
+    /// - Parameters:
+    ///   - filePath: path to the file.
+    ///   - toGroup: group to add file to.
+    ///   - sourceTree: file sourceTree, default is `.group`
+    ///   - sourceRoot: path to project's source root.
+    /// - Returns: new or existing file and its reference.
+    public func addFile(at filePath: Path, toGroup: PBXGroup, sourceTree: PBXSourceTree = .group, sourceRoot: Path) throws -> (reference: String, file: PBXFileReference) {
+        guard filePath.isFile else {
```
allu22: Why is it necessary to check if the file really exists?

Sources/xcproj/PBXProj+Helpers.swift
```diff
@@ -1,6 +1,28 @@
...
+
+    /// Returns root project.
+    public var rootProject: PBXProject? {
+        return objects.projects[rootObject]
+    }
+
+    /// Returns root project's root group.
+    public var rootGroup: PBXGroup {
+        guard let rootProject = self.rootProject else {
+            fatalError("Missing root project")
```
toshi0383: Is `fatalError` suitable here?
Maybe we can make `rootGroup` optional (`public var rootGroup: PBXGroup?`).

Sources/xcproj/PBXProjObjects+Helpers.swift
ilyapuchka: First to not make project buildable otherwise (it will not build if this file reference is then added to the build phase). Second to not allow paths to folders.

Sources/xcproj/PBXProj+Helpers.swift
ilyapuchka: I think it's more suitable than requiring to unwrap this property which will be nil only in specific situations when accessed before project is correctly setup. In my opinion it would be better if API would not allow to create objects in such invalid or incomplete states (what it does now as far as I understand), but not sure what it will take to change it in this direction.

pepibumur: I agree with @ilyapuchka here. If we use an optional attribute, the developer could wrongly access that attribute and get unexpected behaviours as a result. If we `fatalErrror` they would know that they are doing something wrong and it'd save them some debugging time.

pepibumur:
@allu22 could you have another look and tell us if you think it's ready to be merged?

pepibumur:
@ilyapuchka it seems that linting is failing, that's why Travis-CI is returning an error.

Sources/xcproj/PBXProj+Helpers.swift
toshi0383: Understood!👌

Sources/xcproj/PBXProjObjects+Helpers.swift
allu22: Actually you can non existing add file to build phase. This is common case with generated files. But as this is a helper function, I believe it is ok to make these kinds of checks here.

pepibumur:
Great work @ilyapuchka 👏

ilyapuchka:
🎉

Issue #204, opened 2017-12-23
Convert some UInt properties to Bool


yonaskolb:
There are a number of properties that can only ever be 0 or 1. These can be changed to Bools in the public interface. As this a breaking change it can go into the 2.0 release

pepibumur:
Good one. xcproj 2.0 will be a good release!

PR #218, opened 2018-01-22
Add ObjectReference

yonaskolb:
Now that `PBXObject` no longer contains a reference, the helper methods that return a PBXObject are hard to work with as you don't know the reference, for example `targets(named: String) -> PBXTarget`

This can also be evidenced by the new helper functions in #213 that all return a tuple of the reference and the object.

This PR adds a new genenric type `ObjectReference` that contains the reference and the object. This is also used in XcodeGen to ease the migration to the removal of reference from PBXObject

This is actually a breaking change. I was hoping to get this into 3.0 but I didn't know it was releasing early.

Another addition to the `Objects` class that could have also helped is the addition of a `getReference(object: PBXObject) -> String` that would return the reference for a given object using equatable or class identity, but I thought this would be a cleaner approach

This PR also makes `PBXObject` Equatable

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/218)

### DIFF
--- a/Sources/xcproj/PBXProjObjects+Helpers.swift
+++ b/Sources/xcproj/PBXProjObjects+Helpers.swift
@@ -9,12 +9,12 @@ public extension PBXProj.Objects {
     ///
     /// - Parameter name: target name.
     /// - Returns: all existing targets with the given name.
-    public func targets(named name: String) -> [PBXTarget] {
-        var targets: [PBXTarget] = []
-        targets.append(contentsOf: Array(nativeTargets.values) as [PBXTarget])
-        targets.append(contentsOf: Array(legacyTargets.values) as [PBXTarget])
-        targets.append(contentsOf: Array(aggregateTargets.values) as [PBXTarget])
-        return targets.filter { $0.name == name }
+    public func targets(named name: String) -> [ObjectReference<PBXTarget>] {
+        var targets: [ObjectReference<PBXTarget>] = []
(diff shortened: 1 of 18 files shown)


yonaskolb:
Sounds good.👍  
#204 should be included too

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -58,17 +58,17 @@ public extension PBXProj.Objects {
     ///   - toGroup: parent group
     ///   - options: additional options, default is empty set.
     /// - Returns: all new or existing groups in the path and their references.
-    public func addGroup(named groupName: String, to toGroup: PBXGroup, options: GroupAddingOptions = []) -> [(reference: String, group: PBXGroup)] {
+    public func addGroup(named groupName: String, to toGroup: PBXGroup, options: GroupAddingOptions = []) -> [ObjectReference<PBXGroup>] {
```
yonaskolb: Hmm, just thinking. This could either be `[String: PBXGroup]` or `[ObjectReference<PBXGroup>]`

ilyapuchka: I think array is better here, to make it is easier to inspect all created groups in order that will match their path order. Also I think with dictionary it will be impossible to get the last created group which is likely to be used in other methods as input parameter.

PR #224, opened 2018-01-30
Fixes PBXObject equality

### LINKED ISSUE #223: PBXObject Equatable is broken
#218 broke Equatable even more. Every `PBXObject` subclass now incorrectly returns true for every comparison, which also means many will always incorrectly pass.
It looks like the previous implementation was also not working as subclasses didn't check their parents and superclasses didn't call down to subclasses.
The fix for this should be the addition of Sourcery templates what work well with classes

yonaskolb:
Resolves #223

This makes PBXObject conform to `Equatable` which then calls out to an instance method that all subclasses override. When we move to Sourcery we can remove all of this

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/224)

### DIFF
--- a/Tests/xcprojTests/PBXRezBuildPhaseSpec.swift
+++ b/Tests/xcprojTests/PBXRezBuildPhaseSpec.swift
@@ -22,7 +22,7 @@ final class PBXRezBuildPhaseSpec: XCTestCase {
     }
 
     func test_equals_returnsTheCorrectValue() {
-        let another = PBXResourcesBuildPhase(files: ["123"],
+        let another = PBXRezBuildPhase(files: ["123"],
                                              runOnlyForDeploymentPostprocessing: false)
         XCTAssertEqual(subject, another)
     }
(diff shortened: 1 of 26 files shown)


yonaskolb:
@pepibumur do you know why the tests are failing?

pepibumur:
@yonaskolb, apparently there are some tests that are failing:
image

Tests/xcprojTests/PBXRezBuildPhaseSpec.swift
```diff
@@ -22,7 +22,7 @@ final class PBXRezBuildPhaseSpec: XCTestCase {
     }
 
     func test_equals_returnsTheCorrectValue() {
-        let another = PBXResourcesBuildPhase(files: ["123"],
+        let another = PBXRezBuildPhase(files: ["123"],
```
yonaskolb: This is an example of something that was incorrectly passing tests

pepibumur: :/ we need Sourcery

PR #234, opened 2018-02-12
Fix error thrown on adding file reference to framework

fuzza:
### Short description 📝
Resolves https://github.com/xcodeswift/xcproj/issues/230

### Solution 📦

As it was discussed in [original issue](https://github.com/xcodeswift/xcproj/issues/230), `addFile` method of `PBXProjObjects+Helpers.swift` now checks an existence of a file by given path.

In addition to this, there was added new check for supported file types based on PBXFileReference's
  `fileTypeHash`.

### Implementation 👩‍💻👨‍💻

- [x] Replace check `isFile` with `exists` for a filePath in `addFile` method
- [x] Add checks for supported files
- [x] Add some more test cases
- [ ] Update changelog

Misc:
- [x] Add XCTAssertThrowsSpecificError assert to check if specific error was
  thrown
- [x] Move fixture files to Files directory

### Review notes
`XCodeProjEditingError` enum was changed in this PR, please review if it should be considered as a breaking change.


fuzza:
🎉

Sources/xcproj/PBXProjObjects+Helpers.swift
```diff
@@ -99,10 +99,14 @@ public extension PBXProj.Objects {
         sourceTree: PBXSourceTree = .group,
         sourceRoot: Path) throws -> ObjectReference<PBXFileReference> {
 
-        guard filePath.isFile else {
-            throw XCodeProjEditingError.notAFile(path: filePath)
+        guard filePath.exists else {
+            throw XCodeProjEditingError.fileNotExists(path: filePath)
         }
 
+        guard let fileType = PBXFileReference.fileType(path: filePath) else {
```
yonaskolb: I don't think we should require that it's a file we know about. `PBXFileReference.fileType` isn't an exhaustive list of every possible file, and `explicitFileType` and `lastKnownFileType` are optional properties anyway that get filled in by Xcode when the project is opened.

fuzza:
@yonaskolb fixed

PR #238, opened 2018-03-02
Load Xcode Project-Relative XCConfig Files

briantkelley:
### Short description 📝
If an xcconfig file includes another xcconfig file with a relative path, Xcode will first attempt to load the included file relative to the including xcconfig's path. If that fails, Xcode will attempt to load the included xcconfig file relative to the Xcode project.

### Solution 📦
Add the Xcode Project Path as an optional parameter to the XCConfig initializer, so we can use that to resolve the include's relative path.

I went with making it an optional parameter instead of required to maintain source compatibility and to de-empahasize this less used feature. Open to making it required, however.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/238)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## next version
 
+### Changed
+- Support for `XCConfig` project-relative includes https://github.com/xcodeswift/xcproj/pull/238 by @briantkelley
+
 ## 4.1.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


toshi0383:
I didn’t know about this behavior.👀
Just wondering, but what if the xcodeproj is included from other xcodeproj from different directory?
Does it resolve to `$(SRCROOT)` or how does it work?

briantkelley:
@toshi0383 Yeah, it's similar to `$(SRCROOT)`. It resolves against the Xcode project using the xcconfig file that caused the relative include, so it's consistent regardless of whether an Xcode project is or isn't included in another Xcode project.

PR #239, opened 2018-03-02
Fix PBXObject.isEqual(to:) Implementations

briantkelley:
### Short description 📝
Due to a copy/paste error, super.isEqual(to:) was being called with self instead of the rhs object.

### Solution 📦
Update all implementations to pass the rhs object, not self. Surprisingly no tests were affected.

### Implementation 👩‍💻👨‍💻
While we’re here, implement a minor performance optimization by performing the downcast before the call to super, potentially short circuiting unnecessary comparisons.


yonaskolb:
Whoops 😳 , good catch @briantkelley!

PR #240, opened 2018-03-07
Fix Writing of showEnvVarsInLog

briantkelley:
### Short description 📝
Xcode writes showEnvVarsInLog to the pbxproj file if it’s set to false. If it’s true, it’s omitted.

### Solution 📦
Update xcproj to use the same behavior as Xcode to eliminate diffs when round tripping Xcode projects.

This requires defaulting to true when decoding the project file if the key is omitted. I also updated the default parameter in the initializer to true for good measure.

### Implementation 👩‍💻👨‍💻
- [x] Verify "Show environment variables in build log" is true when creating a new Run Script build phase
- [x] Verify Xcode writes the flag when false but never when true

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/240)

### DIFF
--- a/Sources/xcproj/PBXShellScriptBuildPhase.swift
+++ b/Sources/xcproj/PBXShellScriptBuildPhase.swift
@@ -46,7 +46,7 @@ final public class PBXShellScriptBuildPhase: PBXBuildPhase {
                 shellScript: String? = nil,
                 buildActionMask: UInt = defaultBuildActionMask,
                 runOnlyForDeploymentPostprocessing: Bool = false,
-                showEnvVarsInLog: Bool = false) {
+                showEnvVarsInLog: Bool = true) {
         self.name = name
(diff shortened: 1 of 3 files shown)


Sources/xcproj/PBXShellScriptBuildPhase.swift
```diff
@@ -117,7 +117,10 @@ extension PBXShellScriptBuildPhase: PlistSerializable {
         if let shellScript = shellScript {
             dictionary["shellScript"] = .string(CommentedString(shellScript))
         }
-        dictionary["showEnvVarsInLog"] = .string(CommentedString("\(showEnvVarsInLog.int)"))
+        if !showEnvVarsInLog {
```
pepibumur: What about adding a test that tests this behaviour?

PR #242, opened 2018-03-07
Add PBXProject Properties

briantkelley:
### Short description 📝
`PBXProject` previously loaded `projectReferences.count`. Now xcproj adds it to the plist archive. Add support for the `projectRoots` property.

### Solution 📦
`projectReferences` is a bit tricky to write to the plist without causing diffs. We need to look up the `PBXGroup` and `PBXFileReference` it refers to so we can grab the names for the comments.

Xcode supports multiple project roots. Deprecate `projectRoot` in favor of the new `projectRoots`. A web search for PBXProject.h indicates internally Xcode only stores an array, implying the plist key depends on the number of project roots. Reintroduce Deprecations.swift so we can maintain source compatibility with the `projectRoot` property and initializer parameter.

### Implementation 👩‍💻👨‍💻
- [x] Read and Write projects containing a projectReferences and verify there are no diffs.
- [x] Read and Write projects containing multiple project roots and verify there are no diffs.
- [x] Read and Write projects containing  a single project root and verify there are no diffs.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/242)

### DIFF
--- a/Sources/xcproj/PBXProject.swift
+++ b/Sources/xcproj/PBXProject.swift
@@ -187,6 +187,22 @@ extension PBXProject: PlistSerializable {
         }
         dictionary["projectDirPath"] = .string(CommentedString(projectDirPath))
         dictionary["projectRoot"] = .string(CommentedString(projectRoot))
+        if projectReferences.count > 0 {
+            dictionary["projectReferences"] = .array(projectReferences.flatMap { reference in
+                guard let productGroup = reference["ProductGroup"], let projectRef = reference["ProjectRef"] else {
+                    return nil
+                }
+
+                let groupName = proj.objects.groups.getReference(productGroup)?.name
+                let fileRef = proj.objects.fileReferences.getReference(projectRef)
+                let fileRefName = fileRef?.name ?? fileRef?.path
+
+                return [
+                    CommentedString("ProductGroup") : PlistValue.string(CommentedString(productGroup, comment: groupName)),
+                    CommentedString("ProjectRef") : PlistValue.string(CommentedString(projectRef, comment: fileRefName)),
+                ]
+            })
+        }
         dictionary["targets"] = PlistValue.array(targets
             .map { target in
                 let targetName = proj.objects.getTarget(reference: target)?.name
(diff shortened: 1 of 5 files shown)


Sources/xcproj/Deprecations.swift
```diff
@@ -0,0 +1,46 @@
+
+extension PBXProject {
+
+    @available(*, deprecated, message: "Use projectRoots instead")
```
pepibumur: Awesome 😛

Sources/xcproj/Deprecations.swift
```diff
@@ -0,0 +1,46 @@
+
+extension PBXProject {
```
pepibumur: What do you think about renaming this file to `PBXProject+Deprecations`?

Sources/xcproj/PBXProject.swift
```diff
@@ -121,7 +122,13 @@ final public class PBXProject: PBXObject {
         self.productRefGroup = try container.decodeIfPresent(.productRefGroup)
         self.projectDirPath = try container.decodeIfPresent(.projectDirPath) ?? ""
         self.projectReferences = (try container.decodeIfPresent(.projectReferences)) ?? []
-        self.projectRoot = try container.decodeIfPresent(.projectRoot) ?? ""
+        if let projectRoots: [String] = try container.decodeIfPresent(.projectRoots) {
```
pepibumur: 👏

Sources/xcproj/PBXProject.swift
```diff
@@ -186,7 +193,27 @@ extension PBXProject: PlistSerializable {
...
         }
         dictionary["projectDirPath"] = .string(CommentedString(projectDirPath))
-        dictionary["projectRoot"] = .string(CommentedString(projectRoot))
+        if projectRoots.count > 1 {
+            dictionary["projectRoots"] = projectRoots.plist()
+        } else {
+            dictionary["projectRoot"] = .string(CommentedString(projectRoots.first ?? ""))
+        }
+        if projectReferences.count > 0 {
+            dictionary["projectReferences"] = .array(projectReferences.flatMap { reference in
```
pepibumur: What do you think about extracting this into a separate method. Something like:

```swift
func projectReferencesPlistValue() -> PlistValue
```

Sources/xcproj/Deprecations.swift
briantkelley: Sounds good to me.

PR #243, opened 2018-03-07
Emit More Comments in Xcode Project Files

briantkelley:
### Short description 📝
Update xcproj to support the `comment` property and to fix diff noise when creating plist entires for `PBXVariantGroup`, `PBXProject`, and `PBXTarget`.

### Solution 📦
Add support for the `comment` property via a new `PBXContainerItem` class, which follows Xcode's design based on a Google class-dump header search. Update comment generation `PBXVariantGroup`, `PBXProject`, and `PBXTarget` to eliminate diff noise found when writing the 218 with xcproj.

### Implementation 👩‍💻👨‍💻
- [x] Add `PBXContainerItem` to support the `comment` property, matching the behavior from a class-dump search, and make it the super class of `PBXBuildPhase` and `PBXTarget`.
- [x] When creating the comment fo `PBXVariantGroup`, fall back to `path` if `name` is `nil`.
- [x] Similarly, when creating the comment for `PBXProject.mainGroup` and `PBXProject.productRefGroup`, fall back to `path` if `name` is `nil`.
- [x] `PBXTarget.buildRules` emits the `isa` as the comment, similar to the `dependencies` property.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/243)

### DIFF
--- a/Sources/xcproj/PBXTarget.swift
+++ b/Sources/xcproj/PBXTarget.swift
@@ -102,7 +102,7 @@ public class PBXTarget: PBXContainerItem {
 
         // Xcode doesn't write PBXAggregateTarget buildRules or empty PBXLegacyTarget buildRules
         if !(self is PBXAggregateTarget), !(self is PBXLegacyTarget) || !buildRules.isEmpty {
-            dictionary["buildRules"] = .array(buildRules.map {.string(CommentedString($0))})
+            dictionary["buildRules"] = .array(buildRules.map {.string(CommentedString($0, comment: PBXBuildRule.isa))})
         }
         
         dictionary["dependencies"] = .array(dependencies.map {.string(CommentedString($0, comment: PBXTargetDependency.isa))})
--- a/Sources/xcproj/PBXTarget.swift
+++ b/Sources/xcproj/PBXTarget.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// This element is an abstract parent for specialized targets.
-public class PBXTarget: PBXObject {
+public class PBXTarget: PBXContainerItem {
 
     /// Target build configuration list.
     public var buildConfigurationList: String?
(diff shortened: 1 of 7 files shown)


Sources/xcproj/PBXTarget.swift
```diff
@@ -88,7 +88,7 @@ public class PBXTarget: PBXObject {
     }
 
     func plistValues(proj: PBXProj, isa: String, reference: String) -> (key: CommentedString, value: PlistValue) {
-        var dictionary: [CommentedString: PlistValue] = [:]
+        var dictionary = plistValues(proj: proj, reference: reference)
```
pepibumur: Shouldn't this be `super.plistValues`?

briantkelley: Yeah, I can update that to make it more clear. This method has an `isa:` named argument so this code isn't infinitely recursive. After reading your comment I was like, wait, how *does* this work? :)

briantkelley:
I just searched around PBXTarget.h until I found something interesting. [PBXContainerItem.h](https://github.com/mmmulani/class-dump-o-tron/blob/master/Applications/Xcode.app/Contents/OtherFrameworks/DevToolsCore.framework/Versions/A/DevToolsCore/PBXContainerItem.h) has the `comments` property, which [PBXBuildPhase](https://github.com/mmmulani/class-dump-o-tron/blob/master/Applications/Xcode.app/Contents/OtherFrameworks/DevToolsCore.framework/Versions/A/DevToolsCore/PBXBuildPhase.h) and [PBXTarget](https://github.com/mmmulani/class-dump-o-tron/blob/master/Applications/Xcode.app/Contents/OtherFrameworks/DevToolsCore.framework/Versions/A/DevToolsCore/PBXTarget.h) inherit via the [PBXProjectItem](https://github.com/mmmulani/class-dump-o-tron/blob/master/Applications/Xcode.app/Contents/OtherFrameworks/DevToolsCore.framework/Versions/A/DevToolsCore/PBXProjectItem.h) class.

yonaskolb:
Where is the comment property set within the Xcode UI?

briantkelley:
As far as I know, the UI for editing comments was eliminated in Xcode 4, but we still have a lot of comments throughout our 218 Xcode projects, so even though it's seemingly an obsolete feature, I'd like to eliminate diffs.

PR #244, opened 2018-03-07
Add Properties PBXFileElement

briantkelley:
### Short description 📝
`PBXFileReference` and `PBXGroup` share many properties. Consolidate them in `PBXFileElement` and add additional unique properties to `PBXFileReference`.

### Solution 📦
Refactor the `PBXFileReference.includeInIndex`, `PBXGroup.usesTabs`, `PBXGroup.indentWidth`, `PBXGroup.tabWidth` properties so they're implemented in `PBXFileReference` so they can be shared by both subclasses. Add `PBXFileElement.wrapsLines`, `PBXFileReference.languageSpecificationIdentifier`, and `PBXFileReference.plistStructureDefinitionIdentifier`.

### Implementation 👩‍💻👨‍💻
- [x] Add `PBXFileElement.wrapsLines` property and parameters to the `PBXFileElement`, `PBXFileReference`, and `PBXGroup` initializers.
- [x] Move `includeInIndex` from `PBXFileReference` to `PBXFileElement` and add parameters to the `PBXFileElement` and `PBXGroup` initializers.
- [x] Move `usesTabs` from `PBXGroup` to `PBXFileElement` and add parameters to the `PBXFileElement` and `PBXFileReference` initializers.
- [x] Move `indentWidth` from `PBXGroup` to `PBXFileElement` and add parameters to the `PBXFileElement` and `PBXFileReference` initializers.
- [x] Move `tabWidth` from `PBXGroup` to `PBXFileElement` and add parameters to the `PBXFileElement` and `PBXFileReference` initializers.
- [x] Add `languageSpecificationIdentifier` and `plistStructureDefinitionIdentifier` to `PBXFileReference` and parameters to its initializer.
- [x] Simplify `PBXFileReference.isEqual(to:)` and `PBXGroup.isEqual(to:)` only compare properties the class implements.

This change does not remove any properties or initializer parameters on `PBXGroup` or `PBXFileElement`, so source compatibility should be maintained.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/244)

PR #245, opened 2018-03-07
Special Quote Cases for Commented Strings

briantkelley:
### Short description 📝
Xcode adds quotes around strings in the project file if they contain a triple underscore or double forward slash. Update xcproj to do the same.

### Solution 📦
Check for occurrences of the special sequences and add quotes if they occur.

### Implementation 👩‍💻👨‍💻
- [x] Add a check for a triple underscore after checking for invalid characters.
- [x] Add another check for the double forward slash
- [x] Add test cases for these strings.

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/245)


briantkelley:
Friendly ping ☺️. Are the updates in line with what you were thinking? #244 as well. Thanks!

PR #247, opened 2018-04-04
Target Xcode 9.3

LinusU:
**Opening this for early feedback**

There are probably some more things to do, and I need to fix the tests (would love some pointers on fixing the tests 😄)

This ties into yonaskolb/XcodeGen#284

### Implementation 👩‍💻👨‍💻
- [x] Changed the ordering of `codeCoverageEnabled` & `shouldUseLaunchSchemeArgsEnv`.
- [x] Modified `TestAction` and `LaunchAction` to only print `language` if set.

### GIF

---
This change is [Reviewable](https://reviewable.io/reviews/xcodeswift/xcproj/247)

### DIFF
--- a/Fixtures/WithoutWorkspace/WithoutWorkspace.xcodeproj/xcshareddata/xcschemes/WithoutWorkspace-Package.xcscheme
+++ b/Fixtures/WithoutWorkspace/WithoutWorkspace.xcodeproj/xcshareddata/xcschemes/WithoutWorkspace-Package.xcscheme
@@ -26,7 +26,6 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      language = ""
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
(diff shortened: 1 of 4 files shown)


pepibumur:
Good job @LinusU. You'd need to update the CHANGELOG to include your changes in the next version. Regarding the tests failing it seems that there's something wrong with the Danger script. It's a bit weird because it's not related to your changes.

My guess here is that danger-ruby-swiftlint [version has been bumped](https://github.com/ashfurrow/danger-ruby-swiftlint/blob/master/lib/version.rb) and it's causing that bug that you see in the logs. Could you try pinning that dependency to the version `0.11.1`?

```ruby
# Gemfile
pod "danger-swiftlint", "0.11.1"
```
If that doesn't work we can disable swiftlint from Danger and open an issue to fix it on another PR. This PR should be blocked by this 😛

LinusU:
> You'd need to update the CHANGELOG to include your changes in the next version. 

Will do!

> Regarding the tests failing [...]

The tests fail locally for me since part of the tests is cloning other repositories, opening, and then saving back the XcodeProj. Since these projects are still using Xcode 9.2, saving them will reorder the mentioned attributes, and remove the `language` property if it's empty. Therefore the tests "fail".

Potentially these projects would soon be updated, I could potentially go and submit a PR to each of them 😄

pepibumur:
Can't we make this change in a way it's backward compatible with those projects that haven't been updated yet? @LinusU

pepibumur:
@LinusU are you ok if I merge this one? I'll solve the issues on master and release a new version of the library.

LinusU:
Yes please! Sorry for the late response, currently on vacation 🏝🍹

yonaskolb:
Thanks @LinusU! Enjoy the vacation! 😄 🏝🍹

PR #250, opened 2018-04-11
Strip Trailing Semicolon in xcconfigs

briantkelley:
### Short description 📝
If a configuration setting ends in a semicolon, Xcode strips it from the value. Update xcproj to do the same.

### Solution 📦
Update the `settingRegex` regular expression to not capture a trailing semicolon (and trailing whitespace) but still capture all of the value portion of the string.

### DIFF
--- a/Sources/xcproj/XCConfig.swift
+++ b/Sources/xcproj/XCConfig.swift
@@ -105,7 +105,7 @@ final class XCConfigParser {
     // swiftlint:disable:next force_try
     private static var includeRegex = try! NSRegularExpression(pattern: "#include\\s+\"(.+\\.xcconfig)\"", options: .caseInsensitive)
     // swiftlint:disable:next force_try
-    private static var settingRegex = try! NSRegularExpression(pattern: "^([a-zA-Z0-9_\\[\\]=\\*~]+)\\s*=\\s*(\"?.*\"?)", options: [])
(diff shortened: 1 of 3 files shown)


Sources/xcproj/XCConfig.swift
```diff
@@ -93,8 +93,14 @@ final class XCConfigParser {
                                                             length: line.count))
             .compactMap { (match) -> (key: String, value: String)?  in
                 if match.numberOfRanges == 3 {
-                    let key: String = NSString(string: line).substring(with: match.range(at: 1))
-                    let value: String = NSString(string: line).substring(with: match.range(at: 2))
+                    let key: String = (line as NSString).substring(with: match.range(at: 1))
+                    var value: String = (line as NSString).substring(with: match.range(at: 2))
+
+                    // strip the trailing semicolon, if any
+                    if value.hasSuffix(";") {
```
pepibumur: What about using this regular expression? `.+(;\s*)$`? It should also catch the cases where there are spaces right after the semicolon.

tkrajacic: Any whitespace before the semicolon could also be stripped right?

pepibumur: We can then use `.+(\s*;\s*)$`

briantkelley: So the problem was if the semi-colon was optional then the `.*` would consume everything. So I updated that to `.*?` (dunno why I didn't think of that yesterday) and now it works as expected. Thanks!

PR #255, opened 2018-04-28
Fix fullPath for files that reside within a variant group.

ileitch:
Fix `fullPath` for files that reside in a `Base` variant group, e.g a `xib` that lives inside `Base.lproj`.


pepibumur:
Ignore the Danger error on CI. If you ran the tests locally and it passes feel free to merge it @ileitch. I'll fix the Danger issue on another PR.

PR #262, opened 2018-05-18
Add file reference identifier

daveanderson:
### Short description 📝

The file reference identifier may have been present in earlier versions of xcproj but was no longer exposed. This value is valuable for matching up files to specific references in other places in the project.

e.g. getting the variantGroup that includes localizable `Localizable.strings` and then iterating the `fileReferences` to match up the file that matches the `Localizable.strings` variant by reference identifier.

### Solution 📦

I could see that the codable container for included `reference` but that for some reason is was no longer being exposed by `PBXFileReference` or any of its superclasses.

### Implementation 👩‍💻👨‍💻

- [ ] Add the optional `reference` property to `PBXFileReference`, including codable conformance
- [ ] Include `reference` in the `PBXFileReferenceSpec`

### DIFF
--- a/Sources/xcproj/PBXFileReference.swift
+++ b/Sources/xcproj/PBXFileReference.swift
@@ -7,6 +7,9 @@ final public class PBXFileReference: PBXFileElement {
  
     // MARK: - Attributes
     
+    /// Element reference identifier
+    public var reference: String?
+    
     /// Element file encoding.
     public var fileEncoding: UInt?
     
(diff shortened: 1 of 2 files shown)


pepibumur:
@daveanderson this is not necessary anymore. The new version that is coming handles references in a better way.

- When you create an object it has a temporary reference that is not exposed publicly.
- When the object gets added to the project, you'll get a reference that you can use to reference that object from other objects.

You as a developer as responsible for keeping that reference and use it from wherever you want. References are not strings anymore, and instead, they are a reference type that can be passed around. When the project gets written a deterministic permanent value gets generated *(this is not implemented yet)*.

PR #273, opened 2018-06-20
Conform Equatable from XCScheme types

pepibumur:
### Short description 📝
XCScheme types don't conform the equatable protocol. This PR aims to add conformance to all the scheme-related types in the project.


Sources/xcodeproj/XCScheme+BuildableProductRunnable.swift
```diff
@@ -0,0 +1,42 @@
+import AEXML
```
yonaskolb: You could also potentially put these files in an `XCScheme` group and drop the `XCScheme+` from the filename. It would clean up the list of files a bit as there are so many now

Sources/xcodeproj/XCScheme+BuildableProductRunnable.swift
```diff
@@ -0,0 +1,42 @@
...
+            let element = AEXMLElement(name: "BuildableProductRunnable",
+                                       value: nil,
+                                       attributes: ["runnableDebuggingMode": runnableDebuggingMode])
+            element.addChild(buildableReference.xmlElement())
+            return element
+        }
+
+        // MARK: - Equatable
+
+        public static func == (lhs: BuildableProductRunnable, rhs: BuildableProductRunnable) -> Bool {
```
yonaskolb: Any particular reason these aren't using `Sourcery`?

pepibumur: It doesn't work well with nested types.

Sources/xcodeproj/XCScheme+BuildableProductRunnable.swift
pepibumur: 👍

PR #277, opened 2018-07-28
Fix crash on fullPath of fileRef.path

toshi0383:
### Short description 📝
`PBXFileElement.fullPath(sourceRoot:)` crashes by `precondition` when the `PBXFileReference` is generated using SwiftPM `generate-xcodeproj`.

### Solution 📦
Check if it's an absolute path and return as is if it is.

### DIFF
--- a/Sources/xcodeproj/PBXFileElement.swift
+++ b/Sources/xcodeproj/PBXFileElement.swift
@@ -152,7 +152,13 @@ public extension PBXFileElement {
             guard let group = projectObjects.groups.first(where: { $0.value.childrenReferences.contains(reference) }) else { return sourceRoot }
             guard let groupPath = try group.value.fullPath(sourceRoot: sourceRoot) else { return nil }
             guard let filePath = self is PBXVariantGroup ? try baseVariantGroupPath() : path else { return groupPath }
-            return groupPath.appending(RelativePath(filePath))
+            if filePath.hasPrefix("/") {
+                // Swift PM generated pbxproj can have AbsolutePath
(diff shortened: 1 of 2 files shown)


Sources/xcodeproj/PBXFileElement.swift
```diff
@@ -152,7 +152,13 @@ public extension PBXFileElement {
             guard let group = projectObjects.groups.first(where: { $0.value.childrenReferences.contains(reference) }) else { return sourceRoot }
             guard let groupPath = try group.value.fullPath(sourceRoot: sourceRoot) else { return nil }
             guard let filePath = self is PBXVariantGroup ? try baseVariantGroupPath() : path else { return groupPath }
-            return groupPath.appending(RelativePath(filePath))
+            if filePath.hasPrefix("/") {
```
pepibumur: I'd use [this constructor](https://github.com/apple/swift-package-manager/blob/master/Sources/Basic/Path.swift#L62) which is intended for this.

toshi0383: Done in c11c0d1 👌

toshi0383:
Added test in fba646b
Maybe I can improve coverage a lot more by using [SwiftCheck](https://github.com/typelift/SwiftCheck) but in different PR.

Sources/xcodeproj/PBXFileElement.swift
```diff
@@ -152,7 +152,10 @@ public extension PBXFileElement {
             guard let group = projectObjects.groups.first(where: { $0.value.childrenReferences.contains(reference) }) else { return sourceRoot }
             guard let groupPath = try group.value.fullPath(sourceRoot: sourceRoot) else { return nil }
             guard let filePath = self is PBXVariantGroup ? try baseVariantGroupPath() : path else { return groupPath }
-            return groupPath.appending(RelativePath(filePath))
+
+            // Swift PM generated pbxproj can have AbsolutePath
```
pepibumur: I wouldn't add comments in the code unless it's necessary. In this case, one can read from the last line of the method that the path can be either absolute or relative.

toshi0383:
Removed test's `do-catch` in 004bdd9 👌
Thanks for review!

Issue #285, opened 2018-08-19
Missing some PBXObjectReferences generation


yonaskolb:
xcodeproj 5.0 regression

The following references don't seem to be generated before writing:
- `PBXLegacyTarget.buildPhases`
- `PBXTargetDependency.targetDependency`
- `PBXAggregateTarget.buildPhases`
- `PBXAggregateTarget.dependencies`

PR #290, opened 2018-08-29
Fix targets not getting the reference generated

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/285

### Short description 📝
Due to a bug in the implementation, non-native targets didn't get the object reference generated.


yonaskolb:
Thank you for these PRs @pepibumur! 😄

Issue #288, opened 2018-08-19
No way to set TargetAttributes


yonaskolb:
`PBXProject.attributes` is a `[String: Any]`.
One of the important attributes is `TargetAttributes` which is a map of the target UUID to special properties including, but not limited to `TestTargetID`, `ProvisioningStyle` and `DevelopmentTeam`.

Now that `PBXTarget.reference` is a `PBXObjectReference` in 5.0 we can't create that map anymore. If `PBXObjectReference.value` was public it would be possible to get the actual reference, but then it would have to be `fixed` before being set on the attributes.

Another option would be to have a typed `targetAttributes: [PBXObjectReference: [String: Any]]` property, that gets resolved and then gets merged with the other `attributes` on write

pepibumur:
Good catch. Are the keys of `targetAttributes` always references to targets? If so, we can use `PBXObjectReference` as you suggested and update the writing logic to take the value. 

I wouldn't make the property `PBXObjectReference.value` because I think that's an internal detail the consumers of the API shouldn't worry about.

yonaskolb:
Yeah I don't think the keys can be anything other than PBX references

pepibumur:
Cool, then let's do it the way you suggested.

PR #295, opened 2018-09-07
Expose a public interface for setting/removing project target attributes

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/288

### Short description 📝
As @yonaskolb pointed out, after introducing object reference with xcodeproj 5, it was not possible to update the target attributes. The property depended on the object reference value being public, which we intentionally made internal.

### Solution 📦
- Make the `attributes` property internal.
- Expose three methods to set, remove, and read the attributes given a target.

### DIFF
--- a/Sources/xcodeproj/Objects/Project/PBXProject.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXProject.swift
@@ -126,6 +126,14 @@ public final class PBXProject: PBXObject {
         attributeReferences.removeValue(forKey: target.reference)
     }
 
+    /// Returns the attributes of a given target.
+    ///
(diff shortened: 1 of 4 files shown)


Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -108,7 +108,31 @@ public final class PBXProject: PBXObject {
     }
 
     /// Project attributes.
-    public var attributes: [String: Any]
+    var attributeReferences: [PBXObjectReference: [String: Any]]
```
pepibumur: I opted for making this one internal instead of using `PBXTarget` as a key. That would have required making `PBXTarget` conform the `Hashable` protocol.

PR #298, opened 2018-09-13
Some changes to support Xcode 10

pepibumur:
### Short description 📝
- Update constants.
- Change Xcode version in the CircleCI configuration file.

### DIFF
--- a/Sources/xcodeproj/Project/Xcode.swift
+++ b/Sources/xcodeproj/Project/Xcode.swift
@@ -7,41 +7,41 @@ public struct Xcode {
         /// Last known SDKs.
         public struct SDK {
             /// Last known SDK for iOS.
-            public static let ios: String = "11.3"
+            public static let ios: String = "12.0"
 
             /// Last known SDK for macOS.
-            public static let macos: String = "10.13"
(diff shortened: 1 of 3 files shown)


Sources/xcodeproj/Project/Xcode.swift
```diff
@@ -7,41 +7,41 @@ public struct Xcode {
...
 
     /// Default values.
     public struct Default {
         /// The default object version for Xcodeproj.
-        public static let objectVersion: UInt = 48
+        public static let objectVersion: UInt = 50
 
         /// Default compatibility version.
-        public static let compatibilityVersion: String = "Xcode 3.2"
+        public static let compatibilityVersion: String = "Xcode 9.3"
```
yonaskolb: Does this mean Xcode 10 projects are only openable by Xcode 9.3?

pepibumur: I guess so because that's the value Xcode sets when you create a project with Xcode 10

pepibumur:
CI is happy compiling it with Xcode 10. Any other concern @yonaskolb ?

Issue #304, opened 2018-09-18
Compile error in release builds


yonaskolb:
It seems Swift 4.2 is failing to build xcodeproj in release mode for some reason.
This can be seen by running:
```
swift build -c release
```

The following error occurs
```
PHI node operands are not the same type as the result!
  %159 = phi %Ts23_ContiguousArrayStorageCySSG* [ %336, %328 ], [ %143, %481 ], [ %143, %470 ], [ %143, %454 ], [ %143, %438 ], [ %143, %423 ], [ %143, %146 ]
<unknown>:0: error: fatal error encountered during compilation; please file a bug report with your project and the crash log
<unknown>:0: note: Broken function found, compilation aborted!
0  swift                    0x000000010e76664a PrintStackTraceSignalHandler(void*) + 42
1  swift                    0x000000010e765dfe SignalHandler(int) + 302
2  libsystem_platform.dylib 0x00007fff7908df5a _sigtramp + 26
3  libsystem_platform.dylib 0x000000011b583600 _sigtramp + 2723108544
4  libsystem_c.dylib        0x00007fff78e2b1ae abort + 127
5  swift                    0x000000010a9edaf7 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*)::$_1::__invoke(void*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool) + 519
6  swift                    0x000000010e71a278 llvm::report_fatal_error(llvm::Twine const&, bool) + 280
7  swift                    0x000000010e71a15b llvm::report_fatal_error(char const*, bool) + 43
8  swift                    0x000000010e6e17e6 (anonymous namespace)::VerifierLegacyPass::runOnFunction(llvm::Function&) + 54
9  swift                    0x000000010e68804e llvm::FPPassManager::runOnFunction(llvm::Function&) + 446
10 swift                    0x000000010e690363 llvm::legacy::FunctionPassManagerImpl::run(llvm::Function&) + 115
11 swift                    0x000000010e690258 llvm::legacy::FunctionPassManager::run(llvm::Function&) + 392
[...]

PR #307, opened 2018-09-18
Workaround Swift 4.2 compiler crash

keith:
Resolves https://github.com/tuist/xcodeproj/issues/304


keith:
Ah it looks like a fix has been pushed to master in the meantime. I think this one might be a little nicer though 😄

pepibumur:
Hahah interesting that we came up with a similar solution. Let's merge this one.

keith:
sg

PR #309, opened 2018-09-18
Workaround Swift 4.2 compiler crash

keith:
Resolves https://github.com/tuist/xcodeproj/issues/304

I'm not exactly sure where to submit this since we don't have a branch for the previous release, this branch is based on the tag


pepibumur:
Thanks for the fix @Keith. I'll create a branch on the repository with this change and tag it with a hot fix version number.

pepibumur:
Here's the release https://github.com/tuist/xcodeproj/releases/tag/4.3.1

keith:
Thanks!


--
Keith Smiley

PR #311, opened 2018-09-20
Adding customisable file sorting of groups and lists

drekka:
This PR adds options to sort the files with navigator groups, builds phases and the project file’s internal lists. This is extremely usefil in situations where multiple developers are working together as it drastically reduces the number of merge issues appearing on the project file. Particular where files are ordered by name as the uuid’s can change when multiple developers are involved.


Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -200,6 +205,9 @@ extension PBXProj: Equatable {
 extension PBXProj: Writable {
     public func write(path: AbsolutePath, override: Bool) throws {
         let encoder = PBXProjEncoder()
+        encoder.fileListSortOrder = fileListSortOrder
```
pepibumur: Instead of using an instance variable that is set to the encoder like this, I'd pass is as an argument to the `write` and to the `encode` methods. To avoid having to pass three variables, you can define a new type that contains the sorting order for each of those three elements:

```swift
struct PBXProjEncoderOrder {
  let files: FileListSortOrder
  let navigator: ProjectNavigatorSortOrder
  let buildPhase: BuildPhaseFileSortOrder
}
```
Then you can update the write method signature to be like:

```swift
// Pass a default order value
public func write(path: AbsolutePath, override: Bool, order: PBXProjEncoderOrder = PBXProjEncoderOrder()) throws
```

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
```diff
@@ -16,11 +56,20 @@ final class PBXProjEncoder {
...
+    public var navigatorGroupSortOrder: ProjectNavigatorSortOrder = .unsorted
+    public var buildPhaseFileSortOrder: BuildPhaseFileSortOrder = .unsorted
+    
     // swiftlint:disable function_body_length
     func encode(proj: PBXProj) throws -> String {
+        
         try referenceGenerator.generateReferences(proj: proj)
         guard let rootObject = proj.rootObjectReference else { throw PBXProjEncoderError.emptyProjectReference }
+        
+        // Sort sub objects in advance.
```
pepibumur: I would remove this comment since the code line is descriptive enough.

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
```diff
@@ -148,55 +212,96 @@ final class PBXProjEncoder {
...
     private func increaseIndent() {
         indent += 1
     }
-
+    
     private func decreaseIndent() {
         indent -= 1
     }
+    
+    private func sortFiles<T>(_ obj: T) where T: PBXObject {
```
pepibumur: Nitpick: I see too many empty lines here. Would you mind running `swift format .`?

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
```diff
@@ -16,11 +56,20 @@ final class PBXProjEncoder {
...
+    public var buildPhaseFileSortOrder: BuildPhaseFileSortOrder = .unsorted
+    
     // swiftlint:disable function_body_length
     func encode(proj: PBXProj) throws -> String {
+        
         try referenceGenerator.generateReferences(proj: proj)
         guard let rootObject = proj.rootObjectReference else { throw PBXProjEncoderError.emptyProjectReference }
+        
+        // Sort sub objects in advance.
+        proj.forEach(self.sortFiles)
```
pepibumur: Performance-wise, you are iterating through all the objects in the project, even if you don't plan to sort them. What I suggest is getting the objects that you plan to sort and execute your sorting logic on them. Note that [there are getters](https://github.com/tuist/xcodeproj/blob/master/Sources/xcodeproj/Objects/Project/PBXProj.swift#L88) for all different types of objects that a project has.

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
```diff
@@ -1,5 +1,45 @@
...
+            return lhs.0 < rhs.0
+        }
+    }
+}
+
+// Defines the sorting applied to groups with the project navigator and various build phases.
+public enum ProjectNavigatorSortOrder {
+    case unsorted
+    case byFilename
+    case byFilenameGroupsFirst
```
pepibumur: I like how you defined a function, `sort` in the `FileListSortOrder` type that you pass to the Foundation `sort` method for sorting. Is there any reason why you haven't followed a similar approach for `ProjectNavigatorSortOrder`, `BuildPhaseFileSortOrder`. If not, I'd define it here because we can easily test it without having to test the whole encoder.

drekka:
Ill do some updates tomorrow to do your suggestions.

Sources/xcodeproj/Objects/Project/PBXProj.swift
drekka: Nice. Will also build in a default for anyone which implements the defaults previously used.

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
drekka: Which formatter are you talking about. On my system there isn't any `swift format` command available. I've used https://github.com/nicklockwood/SwiftFormat in the past.

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
drekka: Yep. Good idea. I think originally I thought I'd have to deal with a much larger range of types. Refactoring.

Sources/xcodeproj/Objects/Project/PBXProjEncoder.swift
drekka: Have moved the other sort closures into the enums as well.

drekka:
Ok. I've gone through and addressed each of your comments ( I think :-).

In looking at it, I decided to step back a bit and wrap everything up in a `PBXOutputSettings` instance. This uses a less sort orientated language so if other options come along in the future they can easily be added to the struct without messing with the naming or content.

I've moved the sort closures into the enums, refactored the sort code to just address the objects that need sorting, removed the sort variables and added the output settings to all the functions that need to pass them along.

Hopefully everything looks a lot neater. Oh and ran SwiftFormat across it as well :-)

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
```diff
@@ -0,0 +1,112 @@
+//
```
pepibumur: Sorry for the nitpick, but would you mind removing the comments? 😬

pepibumur:
Thank you very much for addressing the comments @drekka. It's almost ready to be merged 😛. There's something that I forgot yesterday when I was doing the review. I'd be great if we could add a few unit tests to those sort functions. Once done, the PR is ready to be merged.

You'll become contributor of the organization so you'll be able to push directly to the repositories without having to for them.

Fantastic work 👏

drekka:
Cool. The lazy side of me was hoping to ignore the unit testing - ROFL, I'll get on it early next week.

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
drekka: Oops. Will do.

drekka:
Back again (been a bit busy :-). New bunch of commits doing the following:

* Merging in latest code from master branch.
* Clearing all linting issues from running `rake ci`.
* Swift formatted everything.
* Split `PBXProjWriterTests.swift` into `CommentedStringTests.swift` and `PlistValuesTests.swift` as it actually contained tests for those two things.
* Added `PBBProjEncoderTests.swift` and loaded it with tests to test the sorting.
* Refactored the sorting code after finding some issues when sorting groups and storyboards.

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
pepibumur: I think you missed this comment @drekka

pepibumur:
Great work @drekka. Your changes look :+1: to me. There are two more things that need to be done and we can merge the PR:
- Remove the headers from the files that you added.
- Unit test the code in [this file](https://github.com/tuist/xcodeproj/pull/311/files#diff-e009e99cefd3846a5cff7909fca03c8a)

After addressing those, we can hit the merge pull request button 🎉

drekka:
Okies. Should be good now.

pepibumur:
👏 🎉 Awesome! Congrats on your first contribution @drekka

Issue #314, opened 2018-09-23
XCBuildConfiguration.baseConfiguration is the wrong type


yonaskolb:
The `XCBuildConfiguration` `baseConfiguration` init param is supposed to be a `PBXFileReference` and not a `XCBuildConfiguration`

PR #316, opened 2018-09-23
Fix XCBuildConfiguration.buildConfiguration type

yonaskolb:
Resolves https://github.com/tuist/xcodeproj/issues/314


pepibumur:
Good catch! 👍

PR #318, opened 2018-09-23
Fix project attributes

yonaskolb:
This fixes `PBXProject` attributes after https://github.com/tuist/xcodeproj/pull/295

- Fixes target attributes not going into `TargetAttributes`
- Fixes top level, non target, attributes not being settable
- Exposes a `targetAttributes: [PBXTarget: [String: Any]]` by basing `PBXObject` hashable on `ObjectReference` (this is useful for other purposes in XcodeGen)


Sources/xcodeproj/Objects/Sourcery/Equality.generated.swift
```diff
@@ -175,7 +175,10 @@ extension PBXProject {
         if projectReferences != rhs.projectReferences { return false }
         if projectRoots != rhs.projectRoots { return false }
         if targetReferences != rhs.targetReferences { return false }
-        if !NSDictionary(dictionary: attributeReferences).isEqual(to: rhs.attributeReferences) { return false }
+        if !NSDictionary(dictionary: attributes).isEqual(to: rhs.attributes) { return false }
+        if targetAttributeReferences.keys != rhs.targetAttributeReferences.keys { return false }
+        // crashes for unknown reason
+        // if !NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhs.targetAttributeReferences) { return false }
```
pepibumur: Is this crash happening at runtime or during compilation?

pepibumur: It seems a legit crash. `NSDictionary` can't have a `PBXObjectReference` as a key if it's not subclass of `NSObject` and conforms the `NSCoding` protocol. I've pushed a fix for it.
If it looks good, we can merge this.

Issue #287, opened 2018-08-19
Missing blueprintIdentifier PBXObjectReference


yonaskolb:
In 5.0 `XCScheme.BuildableReference.blueprintIdentifier` isn't a `PBXObjectReference`, which means there's no way to set this to (for example) a `PBXNativeTarget.reference`

yonaskolb:
This change was reverted in 6.0

PR #320, opened 2018-09-25
Make BuildableReference.blueprintIdentifier a PBXObjectReference

yonaskolb:
Resolves https://github.com/tuist/xcodeproj/issues/287

### Short description 📝
Makes `BuildableReference.blueprintIdentifier` a `PBXObjectReference` 

### Solution 📦
The xml decoding doesn't have access to a context to get the references, so I just created a static var on `ProjectDecodingContext`, which isn't ideal but it works.


pepibumur:
I'm not sure about that approach though. If you don't decode the project before the workspace, you can't get the object from the workspace. That dependency is very implicit and not obvious. I'd rather keep the attribute in the scheme as a string and provide an API to lookup the object in the project:

```swift
extension PBXProj {
  fun object(uuid: String) -> PBXObject? {

  }
}
```

What do you think?

yonaskolb:
@pepibumur I've removed the shared decoder and just store the string on decoding now. Let me know what you think

yonaskolb:
@pepibumur are you able to take another look at this?

pepibumur:
That looks much better. This is ready to be merged 👍

Issue #322, opened 2018-09-26
Non deterministic scheme XML order


yonaskolb:
As Swift 4.2 has non deterministic hashing, the order of xml elements in Schemes are now random, so there are diffs on every generation. 
There was some work done in https://github.com/tuist/xcodeproj/pull/216 to give order to some properties that didn't match the Swift 4 default order. This needs to be expanded to all the properties.

yonaskolb:
Fixed by #325

PR #325, opened 2018-10-03
Sorting XML attributes

drekka:
Found that every times I saved a test project, that the order of the attributes in the schemes were changing. This resulted in Git thinking the files had changed when in fact they hadn't as the attribute values were the same. 

This PR sorts the XML attributes when writing XML files. This ensures that (after the first time), the attributes will always be in the same order and therefore the file will not be seen as changed unless an actual value changes.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -15,8 +15,8 @@
         "repositoryURL": "https://github.com/apple/swift-package-manager",
         "state": {
           "branch": null,
-          "revision": "6983434787dec4e543e9d398a0a9acf63ccd4da1",
-          "version": "0.2.1"
+          "revision": "235aacc514cb81a6881364b0fedcb3dd083228f3",
+          "version": "0.3.0"
         }
       }
     ]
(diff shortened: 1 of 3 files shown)


yonaskolb:
This could potentially fix #322. It needs to be tested though to see if Xcode rewrites the order when the schemes are edited.

drekka:
Probably as Xcode doesn't seem to care about order. The problem with this was using a dictionary. The order was pretty much changing every time it was written even though the values were the same, so Git saw a lot of false updates so to speak. 

If Xcode changes some values in the scheme then it doesn't matter if it's sorted.

pepibumur:
Thanks for fixing this @drekka. Did you check if Xcode follows the same sorting logic? If so, I think this is ready to be merged. I was thinking about adding an integration test that fails if Xcode changes that logic at any time, but I doubt Xcode will introduce such change because it'd cause a large git diff.

drekka:
Thanks for the feedback guys. I went back into the code and realised the problem was actually not from being unsorted. But due to the BuildAction not being specified in the `attributesOrder` at the top of the extension. I removed my sort code and added BuildAction to the `attributesOrder` with its attributes in the order that Apple uses. Then updated the tests to match.

yonaskolb:
Great! Perhaps it's still good to have sorted keys as a fallback. Then if new properties emerge, at least they will get deterministic sorting

Issue #302, opened 2018-09-18
Generate files with the right UTI


pepibumur:
## Context 🕵️‍♀️
From Apple:

> Uniform type identifiers (UTIs) provide a unified way to identify data handled within the system, such as documents, pasteboard data, and bundles. This chapter describes the concepts behind UTIs and shows how to specify them in your application bundles.

[Reference](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/understanding_utis/understand_utis_conc/understand_utis_conc.html#//apple_ref/doc/uid/TP40001319-CH202-CHDHIJDE)

## What 🌱
Make sure workspaces and projects are written with the right UTI. The UTI of a file can be obtained running `mdls path/to/file`. Here are the UTIs of the Xcode project files:

- Workspace: `com.apple.dt.document.workspace`
- Project: `com.apple.xcode.project`.

## Proposal 🎉
Change the writing logic to set the UTI attribute on the written files.

pepibumur:
![](https://github.trello.services/images/mini-trello-icon.png) [Generate files with the right UTI](https://trello.com/c/pYBfZ6bn/35-generate-files-with-the-right-uti)

IssuehuntBot:
@y-matsuwitter funded this issue with $40. [Visit this issue on Issuehunt](https://issuehunt.io/repos/93763457/issues/302)

PR #327, opened 2018-10-03
Less agressive overwriting of files - DO NOT MERGE

drekka:
I've reopened this for comments to decide whether it needs further work or should be dropped.

It started when I noticed that if I wrote the project files whilst having Xcode open, it would trigger an Xcode reload which in turn left Xcode in a state where it thought it was reloading and would not build or run anything. This is possibly due to a bug in Xcode. 

However an old Python script I've previous used for sorting the project file didn't cause this. Xcode just simply closed the Navigator tree and when I expanded it again all the files were in their sorted order. 

So I did some investigating to see what was going on and it appears that because xcodeproj delete the entire scheme and workspace directories before writing the new files out, it gives Xcode a hernia, leaving it unusable. 

So the change I've put into this PR are a first look at this. The idea is to update these directories rather than blow them away. This lets Xcode gracefully update and keep working. To this end I've modified the writing areas so that they avoid writing files unless their contents have actually changed, and I don't remove any directories during the write. 

As the project I'm working on reads the project, updates and writes it back out this should be fine. But I'm concerned as to whether this may not work for other people's code. Perhaps I should add the ability to switch between the original writing method and this one.

Thoughts? 

P.S. I removed the `override` argument to everything as it wasn't being used that I could see and was possible confusing things. Not sure if I was on the right track with that.


pepibumur:
> So I did some investigating to see what was going on and it appears that because xcodeproj delete the entire scheme and workspace directories before writing the new files out, it gives Xcode a hernia, leaving it unusable.

Thanks for doing this investigation. I've had similar issues with SPM. I think CocoaPods must have handled that scenario already but I'm not sure how they did it though.

> As the project I'm working on reads the project, updates and writes it back out this should be fine. But I'm concerned as to whether this may not work for other people's code.

Looking at your logic my only concern is that the writing logic becomes slower because it does extra 
IO operations. [XcodeGen](https://github.com/yonaskolb/xcodegen), which leverages xcodeproj, is used in large projects which are already hitting performance issues, so we should be careful here cc @yonaskolb. I'd avoid somehow the read that you do when writing the project. What about getting the hash of the content when the project is read, and compare it with the hash of the content that is going to be written?

Sources/xcodeproj/Extensions/AbsolutePath+Extras.swift
```diff
@@ -39,7 +39,19 @@ extension AbsolutePath {
     /// - Parameter content: content to be written.
     /// - Throws: an error if the writing fails.
     func write(_ content: String) throws {
-        try content.write(toFile: asString, atomically: true, encoding: .utf8)
+
+        // Check the current file
+        if exists {
+            // If the new content is the same as the old then don't overwrite.
+            if let oldContents = try? read(), oldContents == content {
```
pepibumur: This is the read I was referring to. If the project is large, this might be expensive.

Sources/xcodeproj/Project/XCBreakpointList.swift
```diff
@@ -401,7 +401,7 @@ public final class XCBreakpointList: Equatable, Writable {
 
     // MARK: - Writable
 
-    public func write(path: AbsolutePath, override: Bool) throws {
```
pepibumur: I'd keep the `override` attribute and discuss first whether it makes sense to keep it in the public API or get rid of it.

yonaskolb:
Yeah I'd be hesitant to have this in XcodeGen. 
Perhaps the best way is writing files to a temporary place and then comparing the file hashes. If they are different then copy the new file over.

pepibumur:
That's a good approach as well. What I'd do then is taking a large project and comparing the time it takes with those three approaches:

- Reading and comparing the content (current approach)
- Calculating the hash in memory.
- Writing and comparing the hash between the files.

Unless we are certain about this not compromising the performance we should make this an opt-in feature.

Sources/xcodeproj/Extensions/AbsolutePath+Extras.swift
drekka: Yep. I initially thought of using a MD5 checksum or similar, but haven't found an easy way to do it. Might have to chase it again as we don't care about what the contents are, just whether they have changed or not. The goal being to avoid writing the same content. However if it ends up being no improvement I'll just drop the idea and concentrate on the directory deletion part which I think is what was giving Xcode a problem.

Sources/xcodeproj/Project/XCBreakpointList.swift
drekka: Do we know if anyone is using it?

drekka:
One piece of code I did try created a `.new` file along side the original file and then used `FileManager.replaceItem(...)` to replace the original file. Strangely, Xcode then didn't respond to the new content at all as if it was notified of the change. Again could take another look at this idea.

Sources/xcodeproj/Project/XCBreakpointList.swift
pepibumur: It's hard to say because we don't have a list of all the projects that are using the project. I'd keep it as it was an open an issue explaining why you think we should deprecate and get rid of the argument. If we end up doing it, it's better to do it in a separate PR so that we can track it down easily.

pepibumur:
> Strangely, Xcode then didn't respond to the new content at all as if it was notified of the change

Does that mean that Xcode didn't reload the changes? Something that I think it might be worth considering is [this issue](https://github.com/tuist/xcodeproj/issues/302). Not sure if setting the right metadata on the file will have any impact on how Xcode reloads the project.

Looking at how CocoaPods does it, they do it by writing and then renaming the file:
- [This](https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project.rb#L359) is the line where they do the writing.
- [This](https://github.com/segiddins/atomos/blob/master/lib/atomos.rb) is the writing logic that does the write and renames.

And digging deeper on the git history, [here](https://github.com/CocoaPods/Xcodeproj/commit/df17a6db69494a65e17c4c1ba1ab2f1098f15a47#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR7) is why they introduced that change:

> Atomically write `project.pbxproj` files, so that Xcode will never see them in a half-written state.

drekka:
I'm going to have another crack at this when I get a spare second or two :-) I'm currently thinking I'll do it this way:

* As each object that represents a file is loaded from a file, I'll compute a hash of the contents and store it in the object. This is to save having to re-read the files later.
* If the developer sets a `merge` option in the `PBXOutputSettings`, do the following:
    * Don't delete the workspace and scheme directories.
    * When writing files:
        * Compute the hash of the new contents.
        * If the file does not have a stored hash then it's new so write it.
        * If the file has a stored hash then compare the hashes and if different, write the file.
        * read the contents of the directory and if there are files that have not been loaded, assume they are to be deleted and do so.

I think if this works, it should mean the minimum amount of file reading and writing when updating a project. It should not change the effort required to write out new projects.

I also want to re-check the `FileManager.replaceItem(...)` as a secure way of overwriting files.

pepibumur:
That idea looks good to me. I'd get some numbers before/after adding your changes to see how they are affecting the performance.

pepibumur:
I'm closing this one since it hasn't had any activity in the last few months. If you would like to continue this work feel free to re-open the PR.

yonaskolb:
Forgot about this one. Perhaps this would solve https://github.com/yonaskolb/XcodeGen/issues/483

drekka:
Yep. Close. I haven't had a chance to come back to this. :-(

PR #330, opened 2018-10-09
Add OSLogs to facilitate measuring times with Instruments

pepibumur:
### Short description 📝
I've added some logs using the new `os_signpost` to help track down performance bottlenecks.

### Solution 📦
Add a `OSLogger` class that logs only if the `XCODEPROJ_OSLOG` environment variable is defined , and the API is available in the OS.

I added some logs in the reading and writing functions.

### DIFF
--- a/Sources/xcodeproj/Project/XcodeProj.swift
+++ b/Sources/xcodeproj/Project/XcodeProj.swift
@@ -21,7 +21,7 @@ public final class XcodeProj: Equatable {
         var workspace: XCWorkspace!
         var sharedData: XCSharedData?
 
-        try OSLogger.instance.log(category: #file, name: "Write workspace", path.asString) {
+        try OSLogger.instance.log(name: "Write workspace", path.asString) {
             if !path.exists { throw XCodeProjError.notFound(path: path) }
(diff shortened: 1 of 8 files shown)


pepibumur:
Would you log anything else @yonaskolb ?

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -81,10 +91,18 @@ extension XcodeProj: Writable {
     ///   If false will throw error if project already exists at the given path.
     public func write(path: AbsolutePath, override: Bool = true, outputSettings: PBXOutputSettings) throws {
         try path.mkpath()
-        try writeWorkspace(path: path, override: override)
-        try writePBXProj(path: path, override: override, outputSettings: outputSettings)
-        try writeSchemes(path: path, override: override)
-        try writeBreakPoints(path: path, override: override)
+        try OSLogger.instance.log(category: String(describing: self), name: "Write workspace", path.asString) {
```
yonaskolb: You could get the category automatically using `#file`

pepibumur: Good idea. Let me update it.

PR #306, opened 2018-09-18
Add array helpers for PBXObjectReference

yonaskolb:
This just adds some extensions on Array to make the conversion between objects and references easier

PR #332, opened 2018-10-09
Reference updates

yonaskolb:
This changes a few things about object references.
This is best reviewed by commit, as each commit changes a single thing:

- reference array helpers to remove duplicated code - reimplements the lost commits of #306 (still not sure what happened to them)
- change `PBXObjectRefernce.object()` from a throwing function to an optional return as the error was usually discarded. This makes a lot of the code simpler. There is a new `PBXObjectRefernce.objectOrThrow()` for the rare cases the error is propagated.
- improve performance of object lookups by caching the PBXObject within the reference
- improve the performance of fixing references by removing the un-needed deletion and re-addition of objects
- add helper function for fixing references which removes duplicated code
- add class acronym prefix to the generated UUID
- add `TEMP` prefix to temporary UUIDs

### DIFF
--- a/Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
@@ -105,9 +105,16 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
 
     /// Returns the object the reference is referfing to.
     ///
+    /// - Returns: object the reference is referring to. Returns nil if the objects property has been released or the reference doesn't exist
+    func object<T>() -> T? {
+        return try? objectOrThrow()
+    }
+
+    /// Returns the object the reference is referfing to.
+    ///
     /// - Returns: object the reference is referring to.
     /// - Throws: an errof it the objects property has been released or the reference doesn't exist.
-    func object<T>() throws -> T {
+    func objectOrThrow<T>() throws -> T {
         guard let objects = objects else {
             throw PBXObjectError.objectsReleased
         }
@@ -128,6 +135,6 @@ extension Array where Element: PBXObject {
 extension Array where Element: PBXObjectReference {
 
     func objects<T: PBXObject>() -> [T] {
-        return compactMap { try? $0.object() }
+        return compactMap { $0.object() }
     }
 }
--- a/Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
@@ -12,6 +12,9 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
     /// Weak reference to the objects instance that contains the project objects.
     weak var objects: PBXObjects?
 
+    /// A weak reference to the object
+    weak var objectRef: PBXObject?
+
     /// Initializes a non-temporary reference.
     ///
     /// - Parameter reference: reference.
@@ -103,10 +106,17 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
         return lhs.value < rhs.value
     }
 
+    /// Sets the object so it can be retrieved quickly again later
+    ///
+    /// - Parameter object: The object
+    func setObject(_ object: PBXObject) {
+        objectRef = object
+    }
+
     /// Returns the object the reference is referfing to.
     ///
     /// - Returns: object the reference is referring to. Returns nil if the objects property has been released or the reference doesn't exist
-    func object<T>() -> T? {
+    func object<T: PBXObject>() -> T? {
         return try? objectOrThrow()
     }
 
@@ -114,13 +124,17 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
     ///
     /// - Returns: object the reference is referring to.
     /// - Throws: an errof it the objects property has been released or the reference doesn't exist.
-    func objectOrThrow<T>() throws -> T {
+    func objectOrThrow<T: PBXObject>() throws -> T {
+        if let object = objectRef as? T {
+            return object
+        }
         guard let objects = objects else {
             throw PBXObjectError.objectsReleased
         }
         guard let object = objects.get(reference: self) as? T else {
             throw PBXObjectError.objectNotFound(value)
         }
+        self.objectRef = object
         return object
     }
 }
--- a/Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
@@ -48,18 +48,11 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
     }
 
     /// Fixes its value making it permanent.
-    /// Since this object is used as a key in to refer objects from the PBXObjects instance, we need to delete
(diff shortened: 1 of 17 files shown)


Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
```diff
@@ -117,3 +117,17 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
...
+extension Array where Element: PBXObject {
+
+    func references() -> [PBXObjectReference] {
+        return map { $0.reference }
+    }
+}
+
+extension Array where Element: PBXObjectReference {
+
+    func objects<T: PBXObject>() -> [T] {
```
pepibumur: 👏 nice one.

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
```diff
@@ -12,6 +12,9 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
     /// Weak reference to the objects instance that contains the project objects.
     weak var objects: PBXObjects?
 
+    /// A weak reference to the object
+    weak var objectRef: PBXObject?
```
pepibumur: Nitpick: What about calling this object?

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
```diff
@@ -103,24 +106,35 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
...
+        if let object = objectRef as? T {
+            return object
+        }
         guard let objects = objects else {
             throw PBXObjectError.objectsReleased
         }
         guard let object = objects.get(reference: self) as? T else {
             throw PBXObjectError.objectNotFound(value)
         }
+        self.objectRef = object
```
pepibumur: Why do we need to set this attribute in the getter?

Sources/xcodeproj/Utils/ReferenceGenerator.swift
```diff
@@ -264,24 +241,26 @@ final class ReferenceGenerator: ReferenceGenerating {
...
-    /// make sure the generated reference doesn't collide with the existing one.
+extension PBXObject {
+
+    /// Given a list of identifiers, it generates a deterministic reference.
     ///
     /// - Parameter identifiers: list of identifiers used to generate the reference of the object.
     /// - Returns: object reference.
-    fileprivate func generate(identifiers: [String]) -> String {
-        return identifiers.joined(separator: "-").md5.uppercased()
+    func fixReference(identifiers: [String]) {
```
pepibumur: Much better 👍

Sources/xcodeproj/Utils/ReferenceGenerator.swift
```diff
@@ -258,9 +258,14 @@ extension PBXObject {
...
-            let value = identifiers.joined(separator: "-").md5.uppercased()
-            reference.fix(value)
+            let typeName = String(describing: type(of: self))
+            let acronym = typeName
+                .replacingOccurrences(of: "PBX", with: "")
+                .replacingOccurrences(of: "XC", with: "")
+                .filter { String($0).lowercased() != String($0) }
+
+            let id = ([typeName] + identifiers).joined(separator: "-").md5.uppercased()
+            reference.fix("\(acronym)_\(id)")
```
pepibumur: Why are we adding the acronym here? 
I'd add a comment with some examples of those references. It'll make it easier for someone reading the code in the future to understand the format.

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
```diff
@@ -26,7 +26,7 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
 
     /// Initializes a temporary reference
     init(objects: PBXObjects? = nil) {
-        value = String.random()
+        value = "TEMP_\(String.random())"
```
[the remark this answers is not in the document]
pepibumur: I'd add a test that tests that the value is the expected one when the reference has a temporary value.

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
```diff
@@ -103,17 +99,49 @@ class PBXObjectReference: NSObject, Comparable, NSCopying {
...
+    func object<T: PBXObject>() -> T? {
+        return try? objectOrThrow()
+    }
+
     /// Returns the object the reference is referfing to.
     ///
     /// - Returns: object the reference is referring to.
     /// - Throws: an errof it the objects property has been released or the reference doesn't exist.
-    func object<T>() throws -> T {
+    func objectOrThrow<T: PBXObject>() throws -> T {
```
pepibumur: What about calling this method `materialize`? If it sounds too ambiguous we can leave the one that you chose, it's fine.

Sources/xcodeproj/Utils/ReferenceGenerator.swift
pepibumur: I'd add some tests

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
yonaskolb: Yeah I wanted to but it clashed with `func object()`

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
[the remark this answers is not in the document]
yonaskolb: True. Should `invalidate` create a new value then?

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
pepibumur: I'd keep this one as object then and rename the other. What do you think?

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
pepibumur: Got it!

Sources/xcodeproj/Utils/ReferenceGenerator.swift
pepibumur: Cool, if people find it useful let's add it.

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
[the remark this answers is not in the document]
pepibumur: I'd say so.

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
yonaskolb: Yeah, was thinking rename `object()` to `getObject()`

Sources/xcodeproj/Objects/Project/PBXObjectReference.swift
pepibumur: I've seen the term being used in reactive libraries where a type wraps another type. I think it's a similar notion here, but in this case, we wrap it through a reference. Anyway, feel free to make the call here. This method not internal so we can rename it later if we are not quite happy with how it reads in the codebase.

yonaskolb:
Changes made 👍

PR #334, opened 2018-10-10
Remove SPM dependency

yonaskolb:
Goes someway to resolve https://github.com/tuist/xcodeproj/issues/320

This removes the `swift-package-manager` package and replaces it with `PathKit` and `SwiftShell`.

- `PathKit` replaces the uses of `AbsolutePath` and `RelativePath`
- `SwiftShell` replaces the uses of `Process.run`. This could be done with another dependency, or perhaps no dependency at all


pepibumur:
There seem to be some linting issues.

yonaskolb:
Fixed issues 👍

PR #336, opened 2018-10-11
Performance improvement: Set reference object after object decoding

yonaskolb:
This is a continuation of https://github.com/tuist/xcodeproj/pull/332

This sets the cached reference object when decoding a `PBXObject`


yonaskolb:
The changelog for this is in #338

PR #338, opened 2018-10-11
Performance improvement: use UUID instead of String.random

yonaskolb:
This shaves about another 10% off


yonaskolb:
Updated changelog 👍

PR #342, opened 2018-10-15
Resolve reference conflicts

yonaskolb:
### Short description 📝
This resolves cases where the reference generator generates the exact same value for multiple objects. This can happen when it doesn't have enough context

### Solution 📦
This PR solves the problem in 2 ways. 
- keeping track of what has already been generated, and differentiating duplicates by appending numbers.
- adding a `PBXObject.identifier` string that can be used to give more information about the object to differentiate it from equatable objects. This is used in XcodeGen for example to add a target name to cross platform product file references, which are otherwise all exactly the same as the product name is the same. Maybe this property could have a different name. `referenceContext` or `context` perhaps?

### DIFF
--- a/Sources/xcodeproj/Objects/Project/PBXObject.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXObject.swift
@@ -13,8 +13,16 @@ public class PBXObject: Hashable, Decodable, Equatable, AutoEquatable {
     /// The object reference in the project that contains it.
     let reference: PBXObjectReference
 
-    /// Used to differentiate this object from other equatable ones for the purpose of reference generation
-    public var identifier: String?
+    /**
+     Used to differentiate this object from other equatable ones for the purposes of uuid generation.
+
+     This shouldn't be required to be set in normal circumstances.
+     In some rare cases xcodeproj doesn't have enough context about otherwise equatable objects,
+     so it has to resolve automatic uuid conflicts by appending numbers.
+     This property can be used to provide more context to disambiguate these objects,
+     which will result in more deterministic uuids.
+    */
+    public var context: String?
 
     // MARK: - Init
 
--- a/Sources/xcodeproj/Objects/Project/PBXObject.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXObject.swift
@@ -13,6 +13,9 @@ public class PBXObject: Hashable, Decodable, Equatable, AutoEquatable {
     /// The object reference in the project that contains it.
     let reference: PBXObjectReference
 
+    /// Used to differentiate this object from other equatable ones for the purpose of reference generation
+    public var identifier: String?
+
     // MARK: - Init
 
     init() {
--- a/Sources/xcodeproj/Utils/ReferenceGenerator.swift
+++ b/Sources/xcodeproj/Utils/ReferenceGenerator.swift
@@ -16,6 +16,8 @@ final class ReferenceGenerator: ReferenceGenerating {
     /// Project pbxproj instance.
     var proj: PBXProj?
 
+    var references: Set<String> = []
+
     /// Generates the references of the objects of the given project.
     ///
     /// - Parameter proj: project whose objects references will be generated.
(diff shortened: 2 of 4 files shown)


Sources/xcodeproj/Objects/Project/PBXObject.swift
```diff
@@ -13,6 +13,9 @@ public class PBXObject: Hashable, Decodable, Equatable, AutoEquatable {
     /// The object reference in the project that contains it.
     let reference: PBXObjectReference
 
+    /// Used to differentiate this object from other equatable ones for the purpose of reference generation
+    public var identifier: String?
```
pepibumur: What about using the name that you suggested, `referenceContext`? Furthermore, I'd make the variable internal.

Sources/xcodeproj/Utils/ReferenceGenerator.swift
```diff
@@ -25,6 +27,16 @@ final class ReferenceGenerator: ReferenceGenerating {
         }
 
         self.proj = proj
+
+        // cache current reference values
+        var references: Set<String> = []
```
pepibumur: What about creating an extension for collections whose type is `PBXObject`? We could extract this logic there and simplify this code by doing:

```swift
self.references = proj.objects.fixedReferences
```


Moreover, what about renaming `references` to `cachedReferences`?

Sources/xcodeproj/Objects/Project/PBXObject.swift
pepibumur: I think it should be enough with the logic that was added to the `ReferenceGenerator`. Could you point me to a concrete example on XcodeGen to better understand why this might be necessary?

yonaskolb: It’s required in the linked XcodeGen PR https://github.com/yonaskolb/XcodeGen/pull/422/files#diff-2f3c621620623f55e7ff42cc32198a10R136
Without this then some file references would be duplicates and get the numbers appended, which also isn’t deterministic as they can’t be sorted in the group they are in, so would always cause diffs

yonaskolb: This context is something we lost the ability to add with the move to automatic reference generation. With saying that I think if this stays around it should be renamed to `context`

pepibumur: > Without this then some file references would be duplicates and get the numbers appended, which also isn’t deterministic as they can’t be sorted in the group they are in, so would always cause diffs

I see. I was wondering if we could somehow make that scenario, two targets with the same name, deterministic. For example, those targets have a fixed position in the project they belong to. What about using that?

I think we should aim for exposing as less as possible to make the API easy to use. If we add properties like this one, `identifier` or `context`, there's one more thing the consumer of the API needs to reason about.

Sources/xcodeproj/Utils/ReferenceGenerator.swift
pepibumur: > Objects isn’t a collection. Then we’d need a getObjects() function. Could add it, just trying to cut down on those huge methods that reference every type.

Fair enough :+1:

> Sure it could be renamed. This is all internal stuff so can be refactored. I just thought it’s more than a cache and is actually used in the logic

Let's keep it as `references`.

Sources/xcodeproj/Objects/Project/PBXObject.swift
yonaskolb: By position do you mean sort? Because they are equatable you can’t sort them. Or did you have another idea?

I totally agree about not exposing too much public api. This is a totally optional property though. I can add more documentation explaining that

pepibumur: > By position do you mean sort? Because they are equatable you can’t sort them. Or did you have another idea?

I mean that their position [in the project](https://github.com/tuist/xcodeproj/blob/master/Sources/xcodeproj/Objects/Project/PBXProject.swift#L93) can be used as a context. Do you think that'd work?

> I totally agree about not exposing too much public api. This is a totally optional property though. I can add more documentation explaining that

Worst case we need to have that property we should definitively docucument it 👍

yonaskolb:
I've renamed to `context` and added some more documentation

yonaskolb:
@pepibumur what do you think?
The only other automated way I can think of is to find the target names for every single file reference by checking it's build files -> build phases -> target -> name, which is very expensive

pepibumur:
> The only other automated way I can think of is to find the target names for every single file reference by checking it's build files -> build phases -> target -> name, which is very expensive

Let's keep the context. The documentation that you added explains very well what the attribute is for, and as you pointed out, it's not necessary in most of the cases.

Issue #341, opened 2018-10-15
Issue with adding file to PBXGroup


Vyeczorny:
Hi, I have an issue with adding a file to `PBXGroup`. I'm trying to add file to the group using method:
```
try group.addFile(at: Path(filePath), sourceTree: .group, sourceRoot: Path(projectPath))
```
where:
```
filePath = /Users/username/Documents/Praca/misc/TestApp/TestApp/TestApp/Connectors/TestConnector.swift
projectPath = /Users/username/Documents/Praca/misc/TestApp/TestApp
group.path! = "Connectors"
```

The file is added to the correct group, but the path in the .pbxproj file is wrong (path doesn't contain `Connectors` component):
```
FR_97E519B91B0000E0820899607156CA62 /* TestConnector.swift */ = {isa = PBXFileReference; explicitFileType = sourcecode.swift; lastKnownFileType = sourcecode.swift; name = TestConnector.swift; path = /Users/username/Documents/Praca/misc/TestApp/TestApp/TestApp/TestConnector.swift; sourceTree = "<group>"; };
```

I investigated that it is removed in line 173 of file `PBXGroup`, while creating path to save in pbxproj. Am I doing something wrong?

Thanks for your help,
Karol

pepibumur:
Hi @Vyeczorny. `filePath` should be relative to the group you are adding the file to. If you want to use absolute paths instead, which I don't recommend, you can use absolutePath as sourceTree.

Let me know if that helps :)

Vyeczorny:
I used absolute path only in development process, I will use relative path in final project. However, I think I found some bug, I created a [pull request](https://github.com/tuist/xcodeproj/pull/343), could you take a look on it?

pepibumur:
That's a legit bug. Thanks for adding a fix for it.

PR #343, opened 2018-10-15
Fixed not working PBXFileElement.fullPath(sourceRoot:) method

Vyeczorny:
Resolves https://github.com/tuist/xcodeproj/issues/341

### Short description 📝

I discovered that `fullPath(sourceRoot:)` method doesn't work properly - it removes some part the base path instead of concatenating element's path with its parents' paths and `sourceRoot`. 

In PR: https://github.com/tuist/xcodeproj/pull/334/files#diff-0e1bd5e5f3fdbfcaa7ef8aa4727d3895 `swiftPM` package was removed,`PathKit` was added and all occurrences of `AbsolutePath(filePath, relativeTo: groupPath)` initialized was changed to `Path(filePath).relative(to: groupPath)`, which is not the same (see implementation [here](https://github.com/apple/swift-package-manager/blob/0af02ed21df87bd5bcf169132d0ee90a27c47bdf/Sources/Basic/Path.swift#L62) and [here](https://github.com/tuist/xcodeproj/blob/7a32fa0ebf0c6120ed8a7de03a632c2a10bbe243/Sources/xcodeproj/Extensions/Path%2BExtras.swift#L43) ).

### Solution 📦
I used `+` operator for `Path` instead of `relative` method.


pepibumur:
Awesome work @Vyeczorny. This is ready to be merged.

Issue #344, opened 2018-10-19
Generate XCode style UUIDs


samskiter:
## Context 🕵️‍♀️

We are using [XCodeGen](https://github.com/yonaskolb/XcodeGen) to generate projects.
We are also using [the famous sort script from apple](https://github.com/mas-cli/mas/blob/master/script/sort.pl) to keep our project file sorted and reduce the chances of merge conflicts.

## What 🌱
After generating a project using XCodeGen, we ran the sort scripts and found it failed on line 171. The UUID was not found on a line.

[Tracking this down](https://github.com/tuist/xcodeproj/blob/b1f131b8614a1846b165baa02d7162a3c949ba21/Sources/xcodeproj/Utils/ReferenceGenerator.swift#L299), it looks like this is because XCodeProj generates UUIDs of the format `AAA_123456789123_1` (where AAA is an acronym for the portion of the project file the reference is being generated for and the final _1 is if a reference is already taken).

## Proposal 🎉
Please generate UUIDs in the same format that XCode does in order to keep other tools that make assumptions about the format of UUIDs working. Or provide an option to generate as XCode would.

To achieve this, I imagine you could perform a 24 or 32 byte hash operation on the `reference` variable, here: https://github.com/tuist/xcodeproj/blob/b1f131b8614a1846b165baa02d7162a3c949ba21/Sources/xcodeproj/Utils/ReferenceGenerator.swift#L299

samskiter:
Closing - completely our fault, we are on old XCodeGen and therefore old XCodeProj. Appologies for the noise!

samskiter:
Actually, updating XCodeGen (which uses the latest XCodeProj) we still are getting UUIDs that look like "AAA_123456789". Is that expected?

pepibumur:
cc @yonaskolb

pepibumur:
Closing this one in favor of https://github.com/tuist/xcodeproj/issues/392

PR #345, opened 2018-10-25
Generate XCode style UUID references

samskiter:

### Short description 📝
Addresses https://github.com/tuist/xcodeproj/issues/344

### Solution 📦
- Applies the MD5 to the whole reference ID
- Appends the counter to the ID *before* applying the MD5 hash
- Simplified the reference search by always appending the counter
- @keefmoon added tests for the format of the reference to make sure it is compatible with common XCode project tools (like sort.sh)

### DIFF
--- a/Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
@@ -120,6 +120,17 @@ public enum PBXBuildPhaseFileOrder {
     }
 }
 
+/// Defines the format of project file references
+public enum PBXReferenceFormat {
+    /// Adds prefix and suffix characters to the references.
+    /// The prefix characters identify the type of reference generated (e.g. BP for Build Phase).
+    /// The suffix number is only added for uniqueness if clashes occur.
+    case withPrefixAndSuffix
+    /// Standard 32 char format that XCode generates.
+    /// Note: Not guaranteed to be the same as XCode generates - only the format is the same.
+    case xcode
+}
+
 /// Struct of output settings passed to various methods.
 public struct PBXOutputSettings {
     /// The sorting order for the list of files in Xcode's project file.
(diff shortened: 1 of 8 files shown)


samskiter:
@pepibumur Would you mind having a look - not a big or controversial change I don't think

samskiter:
Just a friendly nudge - we need to update XCodeGen after this library is updated...

@Vyeczorny @yonaskolb @alvarhansen Not sure if any of you can approve?

Thanks!

samskiter:
Pushed latest change. PTAL.

I changed the contract of the ReferenceGenerator, so this is a breaking change. So I decided to make XCode the default mode.

Feel free to suggest better naming for the enum cases, I was struggling :)

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
```diff
@@ -120,6 +120,17 @@ public enum PBXBuildPhaseFileOrder {
     }
 }
 
+/// Defines the format of project file references
+public enum PBXReferenceFormat {
```
pepibumur: Nitpick: Is this the right format for documenting enums? IIRC the documentation of the cases goes in the enum documentation.

pepibumur:
Just a minor comment. Once addressed, we can go forward and merge it :shipit:

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
samskiter: This is consistent with the documentation of other enums in this class.

samskiter: I checked, and while the automatic doc shortcut does generate a single doc with bullet points for each of the case statements, the docs are not carried onto the case statements, meaning they effectively appear undoc'd if inspecting them. The way we have it here allows you to inspect them individually. My opinion is this format is better for consumers.

samskiter:
@yonaskolb  Any chance you could pull this change into XCodeGen please? Thanks!

samskiter:
& thanks for the reviews all!

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
pepibumur: Fair enough. Being able to inspect them individually is a great argument to keep the format that you proposed.

PR #347, opened 2018-11-10
Add .staticFramework to PBXProductType.

ileitch:
The [realm-cocoa](https://github.com/realm/realm-cocoa/) project contains a product type of `com.apple.product-type.framework.static`. It's a non-standard product type which appears to be created by https://github.com/kstenerud/iOS-Universal-Framework. Usually I wouldn't suggest to support non-standard types like this, but it's a small change and enables compatibility with a popular project.


Sources/xcodeproj/Objects/Targets/PBXProductType.swift
```diff
@@ -4,6 +4,7 @@ public enum PBXProductType: String, Decodable {
     case none = ""
     case application = "com.apple.product-type.application"
     case framework = "com.apple.product-type.framework"
+    case staticFramework = "com.apple.product-type.framework.static"
```
pepibumur: Would you mind adding a comment explaining why this type and a reference to the project you mentioned on the PR description? I think we should clearly state that this is not a standard product type.

keith:
This won't actually work correctly as a static framework without also changing the object type correct? I think because of that this shouldn't be added since it would mislead consumers who actually want static frameworks since they'd also have to set custom settings

pepibumur:
> This won't actually work correctly as a static framework without also changing the object type correct? 
I think so because is not a product type supported by Xcode. I wonder why Realm uses that one.
Let's wait to see @ileitch's thoughts on this but I agree with your misleading point @keith

keith:
Oh I must have misread? So this product type works (meaning it would set the setting I was talking about by default) but isn't documented? In that case I'm fully in support and we should file radars to make this public. Note static frameworks aren't supported by apple in general anywhere in their docs.

ileitch:
Here's an open radar about the issue: https://openradar.appspot.com/9403816

For my purposes I only needed to parse the project and identify targets, members files. I can't comment on the viability of the build frameworks, though I'm guessing it must work to some extent since Realm is still using it?

keith:
I've confirmed this does work, even on iOS, if you change the project.pbxproj to this type instead of the normal framework type. It correctly sets the object type to static

pepibumur:
> I've confirmed this does work, even on iOS, if you change the project.pbxproj to this type instead of the normal framework type. It correctly sets the object type to static

@keith does that mean that the compiler is able to link that framework statically? Can that framework be copied to the application bundle and have access to the resources in it?

keith:
The linker is able to link it statically yes. I didn’t try but I doubt it handles resources automatically, you likely have to add those to the necessary copy resources phase manually

ileitch:
Still want me to add comment for this? So it looks like it's just an undocumented type rather than non-standard?

pepibumur:
Yes, I'd still add the comment @ileitch. After that, we can merge this PR.

PR #353, opened 2018-12-07
Add .instrpkg as known PBXProductType

toshi0383:
### Short description 📝
xcodeproj fails to parse pbxproj which contains `*.instrpkg` file.

> dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "productType", intValue: nil)], debugDescription: "Cannot initialize PBXProductType from invalid String value com.apple.product-type.instruments-package", underlyingError: nil))

Replicable with this sample project. 
[sample.zip](https://github.com/tuist/xcodeproj/files/2655874/sample.zip)

### Implementation 👩‍💻👨‍💻

Added enum case for UTI `com.apple.product-type.instruments-package`.


pepibumur:
Good catch @toshi0383. The changes look good 👍

Issue #358, opened 2019-01-04
 Is it possible to specify UseModernBuildSystem - false?


johndpope:
## Context 🕵️‍♀️
> Provide information that helps with understanding your issue. For example your use case that the project doesn't cover, what you were doing when you found the bug... You can also provide the version of the library that you were using, how you integrated it with your project, the platform version...

related
https://github.com/yonaskolb/XcodeGen/issues/470

https://stackoverflow.com/questions/51205221/how-can-i-use-the-legacy-build-system-with-xcode-10s-xcodebuild

## What 🌱
n/a

## Proposal 🎉
n/a

johndpope:
related https://github.com/yonaskolb/XcodeGen/issues/470

PR #359, opened 2019-01-05
[WIP] Add support for WorkspaceSettings

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/358

### Short description 📝
Workspaces contain a `WorkspaceSettings.xcsettings` file which contains settings such as a the build system that Xcode should use. xcodeproj does not have support for that file (yet).

### Solution 📦
Add support for parsing and writing `WorkspaceSettings.xcsettings` files.

### Implementation 👩‍💻👨‍💻
- [x] Add `WorkspaceSettings` model.
- [ ] Change `buildSystem` to be of type `BuildSystem` instead of `String?`
- [ ] Add a method to write the file back to disk.
- [ ] Add a `workspaceSettings` attribute to the `XCSharedData` model.


Sources/xcodeproj/Project/WorkspaceSettings.swift
```diff
@@ -0,0 +1,34 @@
...
+
+public enum WorkspaceSettingsError: Error {
+    /// thrown when the settings file was not found.
+    case notFound(path: Path)
+}
+
+/// It represents the WorkspaceSettings.xcsettings file under a workspace data directory.
+public class WorkspaceSettings: Decodable {
+    /// Workspace build system.
+    public let buildSystem: String?
```
pepibumur: I plan to define a new type and use it instead of `String?`:
```swift
enum BuildSystem {
  case original
  case new
}
```

yonaskolb:
Awesome! 👏
Perhaps master could have swift format applied? It would make the changes here easier to see

PR #360, opened 2019-01-06
Improve documentation structure

pepibumur:
### Short description 📝
We haven't taken care of the documentation structure since we automated the documentation generation. As a result, the documentation is hard to navigate through due to the lack of structure.

### Solution 📦
Organize the classes, enums, extensions, and utils in multiple sections that are defined in the `.jazzy.yml` configuration file.


pepibumur:
The Jazzy documentation was removed from the repository so this is not necessary anymore.

Issue #352, opened 2018-12-07
Referencing targets in other projects


mxcl:
I am trying to add dependencies to targets in separate projects.

This doesn't appear to be possible. Specifically the `containerPortal` for the `PBXContainerItemProxy` needs to point to the `PBXFileReference` for the separate project. But `PBXContainerItemProxy` can only have its `containerPortal` set to a local `PBXProject` object.

My question is, is this possible and I’m just doing it wrong?

If it's not possible, I could probably submit a PR to add this functionality. Though I'd need guidance.

Thanks.

pepibumur:
Although it uses references internally the public interface requires you to pass a `PBXProject` instance. That means the project needs to [be part of the projects attribute](https://github.com/tuist/xcodeproj/blob/master/Sources/xcodeproj/Objects/Project/PBXObjects.swift#L6).

If the project is already there, just pass that reference, otherwise, you need to add it first.

mxcl:
Thanks, that got me further, I wasn't adding the project before, I'm still having trouble however, I have this:

```swift
let mainProject = try XcodeProj(path: projpath)
let cakeProject = try XcodeProj(path: rootpath + Path(".cake/Cake.xcodeproj"))

let mainPBXProject = try mainProject.pbxproj.rootProject()!
try mainPBXProject.mainGroup.addFile(at: rootpath + ".cake/Cake.xcodeproj", sourceRoot: rootpath)
    .explicitFileType = "wrapper.pb-project"

let cakePBXProject = try cakeProject.pbxproj.rootProject()!
mainProject.pbxproj.add(object: cakePBXProject)

let cakeTarget = cakeProject.pbxproj.nativeTargets.first(where: { $0.name == "Cake" })!

for target in mainProject.pbxproj.nativeTargets where target.productType == .application {
    let proxy = PBXContainerItemProxy(containerPortal: cakePBXProject, remoteGlobalID: cakeTarget, proxyType: .nativeTarget, remoteInfo: cakeTarget.name)
    let dep = PBXTargetDependency(name: cakeTarget.name, target: cakeTarget, targetProxy: proxy)
    target.dependencies.append(dep)
    // mainProject.pbxproj.add(object: proxy) //doing this makes Xcode crash (it adds the project as a full object)
    mainProject.pbxproj.add(object: dep)
}
```

[...]

PR #362, opened 2019-01-29
Fix remote target dependency

mxcl:
Resolves https://github.com/tuist/xcodeproj/issues/352

### Short description 📝

This PR allows creating fileref and target references that in other *embedded* projects and linking targets (in the primary project) to those target references (in the other project).

### Solution 📦

A small amount of changes to function parameter types and inheritances, but they are somewhat intrusive.

### Implementation 👩‍💻👨‍💻

1. `PBNativeTarget.addDependency` adapted to take `PBXTarget` rather than `PBXNativeTarget`
2. `PBContainerItemProxy`’s initializer modified to take `PBXObject` rather than only `PBXProject` because when adding references to another project’s targets you have to pass the target’s fileref.
3. In order to link to targets in other projects the `PBReferenceProxxy` had to made to derive `PBXFileElement`. AFAICT this is logical and an omission.

### Note

I also did the unrelated change: 9b0c6370ce751c120dc6e5f55a0105b4a3414794 which removes the `throws` type since the function doesn't in fact `throw`. This is API safe, but will show a warning for any end-users. Happy to remove this from this PR if desired.

### DIFF
--- a/Sources/xcodeproj/Objects/Targets/PBXReferenceProxy.swift
+++ b/Sources/xcodeproj/Objects/Targets/PBXReferenceProxy.swift
@@ -3,14 +3,11 @@ import Foundation
 /// A proxy for another object which might belong to another project
 /// contained in the same workspace of the document.
 /// This class is referenced by PBXTargetDependency.
-public final class PBXReferenceProxy: PBXObject {
+public final class PBXReferenceProxy: PBXFileElement {
     // MARK: - Attributes
 
     /// Element file type
     public var fileType: String?
-
-    /// Element path.
-    public var path: String?
 
     /// Element remote reference.
     var remoteReference: PBXObjectReference?
@@ -25,9 +22,6 @@ public final class PBXReferenceProxy: PBXObject {
         }
     }
 
-    /// Element source tree.
-    public var sourceTree: PBXSourceTree?
-
     // MARK: - Init
 
     /// Initializes the reference proxy.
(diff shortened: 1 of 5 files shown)


Sources/xcodeproj/Objects/Targets/PBXReferenceProxy.swift
```diff
@@ -44,8 +44,8 @@ public final class PBXReferenceProxy: PBXFileElement {
     // MARK: - Decodable
 
     fileprivate enum CodingKeys: String, CodingKey {
-        case remoteRef
         case fileType
+        case remoteRef
```
yonaskolb: This is next level commitment to git diffs 😄

mxcl:
K well, you reviewed it so I force pushed.

Notably the PBFileElement implementation for decoding/encoding the sourceTree is different, so this update will break decoding any persisted `Decodable`s for PBXReferenceProxies.

I don’t know if that's a big deal or not. Please advise.

Sources/xcodeproj/Objects/Targets/PBXReferenceProxy.swift
```diff
@@ -65,30 +56,20 @@ public final class PBXReferenceProxy: PBXObject {
...
-}
-
-// MARK: - PBXReferenceProxy
 
-extension PBXReferenceProxy: PlistSerializable {
-    func plistKeyAndValue(proj _: PBXProj, reference: String) -> (key: CommentedString, value: PlistValue) {
-        var dictionary: [CommentedString: PlistValue] = [:]
+    override func plistKeyAndValue(proj: PBXProj, reference: String) throws -> (key: CommentedString, value: PlistValue) {
+        guard case .dictionary(var dictionary) = try super.plistKeyAndValue(proj: proj, reference: reference).value else {
+            fatalError("super implementation changed and we didn’t realise!")
```
pepibumur: Instead of using `fatalError`, I'd add a test that fails if the super implementation changed. With that test in place, we can safely assume that the value that we get here is a dictionary.

mxcl: Do you have an example of how you’d like this to be done?

mxcl: Happy to do this in order to get this merged, just not sure how to go about it without more direction. Thanks.

mxcl: K I figured something out. Coming up.

mxcl: Although even with a test added to detect if the super changes, I’d still have to `fatalError` here unless you have another suggestion.

mxcl:
I added the test, LMK if it’s good. 👌

mxcl:
Tests passed locally.

pepibumur:
Released with [xcodeproj 6.5.0](https://github.com/tuist/xcodeproj/releases/tag/6.5.0)

PR #365, opened 2019-02-13
Optimize project file decoding

CognitiveDisson:
Hi.
I was going to use your framework to implement a remote cache. But I ran into a blocking problem - it takes my project file about 12 seconds to read from the disk and about 30 seconds to write back. The project is really big - about 13 megabytes, but even for this size such performance is extremely bad. 

I tried to find a issue about this, but did not find it. After that, I looked at the source code and found a lot of performance issues. Some of them I tried to fix with minimal changes. As a result, the project opening time was reduced from 12 to 3 seconds. There are still a few problems left in the code, but the main problems have been fixed. In general, I would recommend to refuse Codable and use dictionaries and Marshal - [benchmarks](https://github.com/bwhiteley/JSONShootout).

## Problem research

I used Xcode Instruments and my [utilitarian class](https://gist.github.com/CognitiveDisson/d04ce89c5ad38ddd46deb31efca2cece) to look for problems.

I made a demo project to check how long it takes to serialize a project file using a simple PropertyListSerialization.
```
    func readPlist(url: URL) {
        do {
            let plistXML = try Data(contentsOf: url)
            var plistData: [String: AnyObject] = [:]
            var propertyListForamt =  PropertyListSerialization.PropertyListFormat.xml
            do {
                plistData = try PropertyListSerialization.propertyList(from: plistXML, options: .mutableContainersAndLeaves, format: &propertyListForamt) as! [String : AnyObject]
                let objects = plistData["objects"] as! [String: AnyObject]
                let classes = plistData["classes"] as! [String: AnyObject]
                let firstObject = objects.first
                let firstClass = classes.first
            } catch {
                print("Error reading plist: \(error), format: \(propertyListForamt)")
            }
        } catch {
            print("error no upload")
        }
    }
```
It turned out that it takes less than a second (very quickly). I understood that the problem is in decoding and serialization.

### Concurrent decoding

First of all, I added concurrent decoding for objects in PBXObjects and implemented synchronization mechanisms for it.

Before: 
```
try objectsDictionaries.forEach { reference, dictionary in
    let object = try PBXObject.parse(reference: reference, dictionary: dictionary, userInfo: decoder.userInfo)
    objects.add(object: object)
}
```

After:
```
let dictionary = NSDictionary(dictionary: objectsDictionaries)
dictionary.enumerateKeysAndObjects(options: .concurrent) { (key, obj, _) in
    guard let reference = key as? String,
        let dictionary = obj as? Dictionary<String, Any>
        else { return }
    let userInfo = decoder.userInfo
    let parser = PBXObjectParser(
        reference: reference,
        userInfo: userInfo
    )
    guard let object = try? parser.parse(dictionary: dictionary)
        else { return }
    objects.add(object: object)
}
```

I chose the standard method `enumerateKeysAndObjects`, but I also tried to [manually select the number of threads](https://gist.github.com/CognitiveDisson/bb88b26262b49b3c78a2ad737024a373). There was no significant difference. This gave about 4 seconds. (  8 seconds instead of 12 ). Instruments showed excellent work parallelization.

### Blind decoding order

I also added very simple optimizations based on the order of parsing - since there are usually more `PBXBuildFile` entries in the project than `PBXTarget` ones, so we should start our blind decoding from them, thus reducing the number of checks. Similar with switches by type. This gave another 1-2 seconds. (6 seconds instead of 12)

### Eliminating unnecessary serialization

After that I found a time consuming decoding here:
```
let objectsDictionary: [String: Any] = try container.decodeIfPresent([String: Any].self, forKey: .objects)
```

I realized that in this case you can convert this project file to serialize into a dictionary and just take a value from it. To make this more accurate, I added a new property to the `ProjectDecodingContext` and read from it directly in the `KeyedDecodingContainer` extension. Currenly this optimization takes effect only during `PBXProj` decoding. For other objects there will be no key provided, but this extra check does not affect their decoding times. ( In the future, it is possible to improve this for nested objects. )

## Result

Finally, I got the cherished 3 seconds. Ideally, I would like about a 1-2 seconds, but this requires global changes (like replacing `Codable` with `Marshal`).

All tests pass (checked many times), and the project parsing is also correct (checked it in my demo project).

I also plan to speed up the project saving performance in the next PR, and I already know what to do.


keith:
I should have a chance to test this in the next few days, our pbxproj file is 32 mbs so it would be great to see an improvement. I'll have to locally work this into an XcodeGen build though

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -70,10 +70,22 @@ public final class PBXProj: Decodable {
...
         let objectsDictionary: [String: Any] = try container.decodeIfPresent([String: Any].self, forKey: .objects) ?? [:]
         let objectsDictionaries: [String: [String: Any]] = (objectsDictionary as? [String: [String: Any]]) ?? [:]
-        try objectsDictionaries.forEach { reference, dictionary in
-            let object = try PBXObject.parse(reference: reference, dictionary: dictionary, userInfo: decoder.userInfo)
+        
+        let dictionary = NSDictionary(dictionary: objectsDictionaries)
+        dictionary.enumerateKeysAndObjects(options: .concurrent) { (key, obj, _) in
+            guard let reference = key as? String,
+                let dictionary = obj as? Dictionary<String, Any>
+                else { return }
```
pepibumur: If the key is not a string, nor the value is a dictionary, we are skipping one entry here, which will result in a in-memory representation of the project that doesn't match the content in disk. What do you think about assuming that the types are the expected ones?

```swift
let reference = key as! String
let dictionary = obj as! Dictionary<String, Any>
```

Alternatively, we can throw an error that gives us more details about why it failed. The idea is being able to catch those as issues, instead of swallowing them.

Sources/xcodeproj/Extensions/Decodable+Dictionary.swift
```diff
@@ -8,8 +8,10 @@ extension Decodable {
     /// - Parameter jsonDictionary: json dictionary.
     /// - Throws: throws an error if the initialization fails.
     init(jsonDictionary: [String: Any]) throws {
```
pepibumur: What do you think about turning this into a static method that is more explicit about the type of decodable that is returning?

```swift
static func projectDecodable(dictionary: [String: Any]) throws -> Decodable
```

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -27,8 +27,8 @@ public final class XcodeProj: Equatable {
             if pbxprojPaths.count == 0 {
                 throw XCodeProjError.pbxprojNotFound(path: path)
             }
-            let pbxProjData = try Data(contentsOf: pbxprojPaths.first!.url)
-            let context = ProjectDecodingContext()
+            let (pbxProjData, jsonDictionary) = try XcodeProj.readPBXProj(url: pbxprojPaths.first!.url)
+            let context = ProjectDecodingContext(jsonDictionary: jsonDictionary)
```
pepibumur: I think this one can be optimised as well by passing a reference instead of a value type. If I'm not wrong, Swift creates a copy of the whole `jsonDictionary` when initialising the `ProjectDecodingContext` object.

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -69,6 +69,22 @@ public final class XcodeProj: Equatable {
             lhs.pbxproj == rhs.pbxproj &&
             lhs.sharedData == rhs.sharedData
     }
+    
+    // MARK: - Private
+    
+    private static func readPBXProj(url: URL) throws -> (Data, [String: Any]) {
+        let plistXML = try Data(contentsOf: url)
+        var propertyListForamt =  PropertyListSerialization.PropertyListFormat.xml
```
pepibumur: Typo: `propertyListForamt`

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -69,6 +69,22 @@ public final class XcodeProj: Equatable {
             lhs.pbxproj == rhs.pbxproj &&
             lhs.sharedData == rhs.sharedData
     }
+    
+    // MARK: - Private
+    
+    private static func readPBXProj(url: URL) throws -> (Data, [String: Any]) {
```
pepibumur: I'd rather pass a `Path` argument to be consistent with the rest of the code.

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -27,8 +27,8 @@ public final class XcodeProj: Equatable {
             if pbxprojPaths.count == 0 {
                 throw XCodeProjError.pbxprojNotFound(path: path)
             }
-            let pbxProjData = try Data(contentsOf: pbxprojPaths.first!.url)
-            let context = ProjectDecodingContext()
+            let (pbxProjData, jsonDictionary) = try XcodeProj.readPBXProj(url: pbxprojPaths.first!.url)
```
pepibumur: I'd rename `jsonDictionary` to `pbxProjDictionary`

Sources/xcodeproj/Project/XcodeProj.swift
pepibumur: Nitpick: I'd also rename the name of the argument to `pbxProjDictionary`

Sources/xcodeproj/Utils/Decoders.swift
```diff
@@ -12,6 +13,10 @@ class PBXObjectReferenceRepository {
     ///   - objects: objects.
     /// - Returns: object reference.
     func getOrCreate(reference: String, objects: PBXObjects) -> PBXObjectReference {
+        lock.lock()
```
pepibumur: What about using the extension for `NSRecursiveLock` that you introduced?

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -69,6 +69,22 @@ public final class XcodeProj: Equatable {
...
+    private static func readPBXProj(url: URL) throws -> (Data, [String: Any]) {
+        let plistXML = try Data(contentsOf: url)
+        var propertyListForamt =  PropertyListSerialization.PropertyListFormat.xml
+        let serialized = try PropertyListSerialization.propertyList(
+            from: plistXML,
+            options: .mutableContainersAndLeaves,
+            format: &propertyListForamt
+        )
+        guard let dictionary = serialized as? [String: Any] else {
+            return (plistXML, [:])
```
pepibumur: Am I wrong in thinking that if the serialised object is not a `[String: Any]` the performance improvements that you introduced will not be executed? I think it's safe to use `as!` here. I doubt that Xcode changes its format to have a key other than a string.

Sources/xcodeproj/Extensions/NSRecursiveLock+Sync.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension NSRecursiveLock {
+    func sync<T>(closure: () -> (T)) -> T {
```
pepibumur: Is this being used?

beefon: I saw `whileLocked()` naming across SPM in similar extensions. Perhaps, you might consider using it as well.

Sources/xcodeproj/Extensions/Decodable+Dictionary.swift
pepibumur: We can get rid of this extension. I searched for it in the project and is only used in a couple of tests, which can be refactored to not use it.

Sources/xcodeproj/Utils/Decoders.swift
```diff
@@ -28,10 +33,12 @@ class ProjectDecodingContext {
 
     /// Objects.
     let objects: PBXObjects
+    let jsonDictionary: [String: Any]
 
-    init() {
+    init(jsonDictionary: [String: Any] = [:]) {
```
beefon: Do you need to have a default value here?

Sources/xcodeproj/Objects/Project/PBXObjectParser.swift
```diff
@@ -0,0 +1,75 @@
...
+    let userInfo: [CodingUserInfoKey: Any]
+    
+    init(reference: String, userInfo: [CodingUserInfoKey: Any]) {
+        self.reference = reference
+        self.userInfo = userInfo
+    }
+    
+    // swiftlint:disable function_body_length
+    public func parse(dictionary: [String: Any]) throws -> PBXObject {
+        let decoder = XcodeprojJSONDecoder()
```
pepibumur: Looking at the code I think there's a bug here *(not introduced by the PR)*. The decoder, has a context object with a property `objectReferenceRepository` which is used by the objects when they are being initialised to create the references with other objects. When we create a decoder for each object being parsed, we are passing an empty reference repository.

We can fix it by moving the decoder to an instance variable of `PBXObjectParser`, and change the logic that iterates through the keys and values of the pbxproj to use the same parser instead of creating an instance in every step of the iteration.

Sources/xcodeproj/Objects/Project/PBXObjectParser.swift
```diff
@@ -0,0 +1,75 @@
...
+        decoder.userInfo = userInfo
+        var mutableDictionary = dictionary
+        mutableDictionary["reference"] = reference
+        let data = try JSONSerialization.data(withJSONObject: mutableDictionary, options: [])
```
pepibumur: This is an unnecessary step that was introduced to leverage the decoders but that we can get rid of in a follow-up PR and get another performance improvement.

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -70,10 +70,22 @@ public final class PBXProj: Decodable {
...
+            let userInfo = decoder.userInfo
+            let parser = PBXObjectParser(
```
pepibumur: I'd move this outside `dictionary.enumerateKeysAndObjects` and change its constructor to be:

```swift
let parser = PBXObjectParser(userInfo: userInfo)
```

Then you can call it for each object:

```swift
parser.parse(reference: reference, dictionary: dictionary)
```

Sources/xcodeproj/Extensions/NSRecursiveLock+Sync.swift
pepibumur: That's a good idea.

pepibumur:
> I should have a chance to test this in the next few days, our pbxproj file is 34 mbs so it would be great to see an improvement. I'll have to locally work this into an XcodeGen build though

It'd be great @Keith if you could give us some numbers from that project.

Sources/xcodeproj/Extensions/Decodable+Dictionary.swift
CognitiveDisson: I will move this file to the test folder - `Tests/xcodeprojTests / Extensions`

keith:
Looks like this saves ~4 seconds or ~10% on the largest version of our project 🎉 (I didn't test a ton)

Sources/xcodeproj/Project/XcodeProj.swift
CognitiveDisson: Fixed

Sources/xcodeproj/Utils/Decoders.swift
CognitiveDisson: Yes. This object is also widely used as a default parameter.

Sources/xcodeproj/Objects/Project/PBXObjectParser.swift
CognitiveDisson: In follow-up PR

CognitiveDisson:
@keith Strange, I expected a better result. Can you provide more information about the project file? What objects are there most?
If possible, I would like to debug on your project .

@pepibumur I corrected most of the comments. Can I merge?

I implemented using a custom decoder to work with a dictionary. This should take away the extra work. According to the results received a slight but still improved. (from 3 seconds to 2)
I'll try to check on other project files. And make a separate PR.

keith:
I can check in more detail but we have ~700 targets and like 15k files

Sources/xcodeproj/Errors/Errors.swift
```diff
@@ -21,6 +22,8 @@ public enum XCodeProjError: Error, CustomStringConvertible {
             return "The project doesn't contain a .pbxproj file at path: \(path.string)"
         case let .xcworkspaceNotFound(path):
             return "The project doesn't contain a .xcworkspace at path: \(path.string)"
+        case let .pbxprojIncorrectSerializationType(path, expectedType, serialized):
+            return "Can't serialize pbxproj plist at path: \(path.string) to: \(expectedType) but serialized result type is \(type(of: serialized)): \(serialized) "
```
pepibumur: Nitpick: I'd rephrase this one to:

```
Can't serialise \(serialized) into a \(expectedType) for project at path \(path)
```

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -70,10 +70,21 @@ public final class PBXProj: Decodable {
...
-            let object = try PBXObject.parse(reference: reference, dictionary: dictionary, userInfo: decoder.userInfo)
+        
+        let parser = PBXObjectParser(
+            userInfo: decoder.userInfo
+        )
+        let dictionary = NSDictionary(dictionary: objectsDictionaries)
+        dictionary.enumerateKeysAndObjects(options: .concurrent) { (key, obj, _) in
+            let reference = key as! String
+            let dictionary = obj as! Dictionary<String, Any>
+            guard let object = try? parser.parse(
```
pepibumur: I'd not use `try?` here and bubble up the error instead. By doing that we prevent the error from going unnoticed and help us detect if there's something to be supported/fixed.

Sources/xcodeproj/Project/XcodeProj.swift
```diff
@@ -69,6 +75,26 @@ public final class XcodeProj: Equatable {
...
+    
+    private static func readPBXProj(path: Path) throws -> (Data, [String: Any]) {
+        let plistXML = try Data(contentsOf: path.url)
+        var propertyListFormat =  PropertyListSerialization.PropertyListFormat.xml
+        let serialized = try PropertyListSerialization.propertyList(
+            from: plistXML,
+            options: .mutableContainersAndLeaves,
+            format: &propertyListFormat
+        )
+        guard let dictionary = serialized as? [String: Any] else {
```
pepibumur: You can safely cast the `serialised` object into a [String: Any] with `as!`. I don't see Xcode replacing the root format of the `pbxproj` file.

Sources/xcodeproj/Errors/Errors.swift
CognitiveDisson: Change throw on force cast and remove this error

Sources/xcodeproj/Project/XcodeProj.swift
CognitiveDisson: I had to add an extension. Method `enumerateKeysAndObjects` doesn't throw error.

pepibumur:
Awesome job @CognitiveDisson 🎉 . I'm merging it.
I've opened [an issue](https://github.com/tuist/xcodeproj/issues/369) to keep track of all the potential performance improvements that could be added to the project.

PR #367, opened 2019-02-15
Fix PBXTarget extension methods

danilsmakotin:
### Short description 📝

- `as` inside `compactMap` will cause exception if a first object into array has another type.
- `buildPhase` inside `resourcesBuildPhase` is compared with wrong type

### Solution 📦
- replace `as` to `as?`
- replace `.sources` to `.resources`

### DIFF
--- a/Sources/xcodeproj/Objects/Targets/PBXTarget.swift
+++ b/Sources/xcodeproj/Objects/Targets/PBXTarget.swift
@@ -200,7 +200,7 @@ public extension PBXTarget {
     /// - Throws: an error if the build phase cannot be obtained.
     public func sourcesBuildPhase() throws -> PBXSourcesBuildPhase? {
         return try buildPhaseReferences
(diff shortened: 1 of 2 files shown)


Sources/xcodeproj/Objects/Targets/PBXTarget.swift
```diff
@@ -212,8 +212,8 @@ public extension PBXTarget {
     /// - Throws: an error if the build phase cannot be obtained.
     public func resourcesBuildPhase() throws -> PBXResourcesBuildPhase? {
         return try buildPhaseReferences
-            .compactMap({ try $0.getThrowingObject() as PBXResourcesBuildPhase })
-            .filter({ $0.buildPhase == .sources })
+            .compactMap({ try $0.getThrowingObject() as? PBXResourcesBuildPhase })
+            .filter({ $0.buildPhase == .resources })
```
pepibumur: Good catch

PR #371, opened 2019-02-19
Project encoding performance improvements

CognitiveDisson:
## Improve project encoding performance 📝
For my test project: `xcodeproject.write(path: path)` was 57 seconds, and it became 8 seconds.

## Obtain PBXBuildFile from PBXBuildPhase instead searching PBXBuildPhase in PBXObjects
The main blocker was obtaining build phase name for PBXBuildFile encoding from PBXObjects -`plistKeyAndValue`. For each file, we ran through all the build phases and searched for a file with the same identifier. Of course, it took a lot of time.

Instead, I added an utility class that contains the file and build phase from which this file is. I am building this class from the build phase files.

This speed up for encoding on 46 seconds. ( 11 seconds instead of 57 ).

## Replaced the append in the string to append into the array + join.
Append into a typed array faster than into a string. This gave about 3 seconds. ( 8 seconds instead of 11 ).

Also this preparation for the subsequent changes. For an array, we can pre-set the size. This will allow non-allocation of memory several times. And we can prepare the data for recording in parallel.

Now this code slows down the most:
```
        elements.forEach { (element, multiline) in
            write(dictionaryKey: element.key, dictionaryValue: element.value, multiline: multiline)
        }
```
We can pass an inout array for each thread. I'm working on it now.

### DIFF
--- a/Sources/xcodeproj/Extensions/Dictionary+Enumerate.swift
+++ b/Sources/xcodeproj/Extensions/Dictionary+Enumerate.swift
@@ -7,7 +7,10 @@ extension Dictionary {
         using block: (Any, Any, UnsafeMutablePointer<ObjCBool>) throws -> Void) throws
     {
         var blockError: Error?
-        (self as NSDictionary).enumerateKeysAndObjects(options: opts) { (key, obj, stops) in
+        // For performance it is very important to create a separate dictionary instance.
+        // (self as NSDictionary).enumerateKeys... - works much slower
+        let dictionary = NSDictionary(dictionary: self)
+        dictionary.enumerateKeysAndObjects(options: opts) { (key, obj, stops) in
             do {
                 try block(key, obj, stops)
             } catch {
--- a/Sources/xcodeproj/Extensions/Dictionary+Enumerate.swift
+++ b/Sources/xcodeproj/Extensions/Dictionary+Enumerate.swift
@@ -15,6 +15,7 @@ extension Dictionary {
                 try block(key, obj, stops)
             } catch {
                 blockError = error
+                stops.pointee = true
             }
         }
         if let error = blockError {
--- a/Sources/xcodeproj/Objects/Project/PBXObjects.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXObjects.swift
@@ -339,6 +339,20 @@ extension PBXObjects {
         phases.merge(frameworksBuildPhases as [PBXObjectReference: PBXBuildPhase], uniquingKeysWith: { first, _ in return first })
         return phases
     }
+    
+    var buildPhaseFile: [PBXObjectReference: PBXBuildPhaseFile] {
+        let values: [[PBXBuildPhaseFile]] = buildPhases.values.map ({ buildPhase in
+            let files = buildPhase.files
+            let buildPhaseFile: [PBXBuildPhaseFile] = files.compactMap({ (file: PBXBuildFile) -> (PBXBuildPhaseFile) in
+                return PBXBuildPhaseFile(
+                    buildFile: file,
+                    buildPhase: buildPhase
+                )
+            })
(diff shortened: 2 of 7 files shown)


Sources/xcodeproj/Extensions/Dictionary+Enumerate.swift
```diff
@@ -7,11 +7,15 @@ extension Dictionary {
         using block: (Any, Any, UnsafeMutablePointer<ObjCBool>) throws -> Void) throws
     {
         var blockError: Error?
-        (self as NSDictionary).enumerateKeysAndObjects(options: opts) { (key, obj, stops) in
+        // For performance it is very important to create a separate dictionary instance.
+        // (self as NSDictionary).enumerateKeys... - works much slower
+        let dictionary = NSDictionary(dictionary: self)
```
CognitiveDisson: @thedavidharris  Should I add `#if os(macOS)` ? + else implementation

dharr419-ford: It's just the `ObjCBool` here that doesn't port, if there's an implementation of this that doesn't use that then it's fine.

thedavidharris: It's just the ObjCBool here that doesn't port, if there's an implementation of this that doesn't use it, that would be best

Sources/xcodeproj/Extensions/Dictionary+Enumerate.swift
```diff
@@ -7,11 +7,15 @@ extension Dictionary {
         using block: (Any, Any, UnsafeMutablePointer<ObjCBool>) throws -> Void) throws
     {
         var blockError: Error?
-        (self as NSDictionary).enumerateKeysAndObjects(options: opts) { (key, obj, stops) in
+        // For performance it is very important to create a separate dictionary instance.
+        // (self as NSDictionary).enumerateKeys... - works much slower
```
pepibumur: Good catch here

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
```diff
@@ -46,6 +46,15 @@ public enum PBXFileOrder {
             return lhs.0 < rhs.0
         }
     }
+    
+    internal func sort(lhs: (PBXObjectReference, PBXBuildPhaseFile), rhs: (PBXObjectReference, PBXBuildPhaseFile)) -> Bool {
```
pepibumur: I'd add a test for this one.

Sources/xcodeproj/Objects/Project/PBXObjects.swift
```diff
@@ -339,6 +339,20 @@ extension PBXObjects {
         phases.merge(frameworksBuildPhases as [PBXObjectReference: PBXBuildPhase], uniquingKeysWith: { first, _ in return first })
         return phases
     }
+    
+    var buildPhaseFile: [PBXObjectReference: PBXBuildPhaseFile] {
```
pepibumur: I'd add a comment here with some context on why this dictionary is necessary. Did you check if the creation of this dictionary has a significant impact on memory consumption?

CognitiveDisson: Yes, I checked, the effect is insignificant (I did not notice it at all), since these are calculated properties and all this data is already in memory.

Sources/xcodeproj/Objects/Project/PBXOutputSettings.swift
CognitiveDisson: Ready

CognitiveDisson:
@pepibumur I fixed all issues. And added one more optimization - concurrent decoding. Performance boost from 8 second to ~3.

PR #372, opened 2019-02-21
Optimized getting full path

CognitiveDisson:

I significantly speed up the full path obtaining , by storing the parent reference.

Was:
```
guard let group = projectObjects.groups.first(where: { $0.value.childrenReferences.contains(reference) }) else { return sourceRoot } // problem
```
Now: 
```
guard let group = parent else { return sourceRoot }
guard let groupPath = try group.fullPath(sourceRoot: sourceRoot) else { return nil }
```


Sources/xcodeproj/Objects/Files/PBXFileElement.swift
```diff
@@ -169,4 +172,13 @@ public extension PBXFileElement {
             .first(where: { $0.name == "Base" }) else { return nil }
         return baseReference.path
     }
+    
+    func assignParentForChildren() {
+        if let group = self as? PBXGroup {
```
pepibumur: Nitpick removing some indentation.

```swift
guard let group = self as? PBXGroup else { return }
group.children.each {  child in
  child.parent = self
  child.assignParentForChildren()
}
```

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
```diff
@@ -169,4 +172,13 @@ public extension PBXFileElement {
             .first(where: { $0.name == "Base" }) else { return nil }
         return baseReference.path
     }
+    
+    func assignParentForChildren() {
```
pepibumur: I'd add a comment here explaining why this is necessary. It'll help us in the future understand why it was added.

pepibumur: Also, I'd call the method `assignParentToChildren`

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
```diff
@@ -141,15 +145,14 @@ public extension PBXFileElement {
...
-        let projectObjects = try objects()
         switch sourceTree {
         case .absolute?:
             return path.flatMap { Path($0) }
         case .sourceRoot?:
             return path.flatMap { sourceRoot + $0 }
         case .group?:
-            guard let group = projectObjects.groups.first(where: { $0.value.childrenReferences.contains(reference) }) else { return sourceRoot }
-            guard let groupPath = try group.value.fullPath(sourceRoot: sourceRoot) else { return nil }
+            guard let group = parent else { return sourceRoot }
```
pepibumur: I have a concern with this. We are disregarding the fact that the user can add a child to a parent without calling the convenience methods where the call to `assignParentForChildren`. If that happens, this method would return an invalid result.

I'd rather fallback to the logic the we had before in case the file element doesn't have a parent.

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
```diff
@@ -141,15 +145,14 @@ public extension PBXFileElement {
...
+            guard let groupPath = try group.fullPath(sourceRoot: sourceRoot) else { return nil }
```
pepibumur: I'd test the following scenarios:
- A file element that has a parent. Test that we get the right path.
- A file element that doesn't have a parent. Test that we get the right path.

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
```diff
@@ -59,6 +62,7 @@ public class PBXFileElement: PBXContainerItem, PlistSerializable {
         self.tabWidth = tabWidth
         self.wrapsLines = wrapsLines
         super.init()
+        assignParentForChildren()
```
pepibumur: I'd add a test to make sure the children get the parent assigned when the group is initialised.

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
CognitiveDisson: i'm added fallback. But i think it is not necessary because in all public methods for adding child called this method.

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
CognitiveDisson: Sure. Renamed and commented.

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
CognitiveDisson: Ready

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
CognitiveDisson: Ready

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
CognitiveDisson: Because of the fallback, the profit decreases, because the root group is searched (it has no parent and it falls into the fallback) - which is not fast if there are a lot of groups.

For my test project, this is 1 minute 30 seconds versus 3 seconds

What do you think about that ?

pepibumur: > i'm added fallback. But i think it is not necessary because in all public methods for adding child called this method.

But it doesn't prevent us from adding a new method in the future and forgetting about this. It's good to add some safety here looking forward.

> Because of the fallback, the profit decreases, because the root group is searched (it has no parent and it falls into the fallback) - which is not fast if there are a lot of groups.

You could check whether the group is a root group and in that case return the `sourceRoot` path. What do you think?

CognitiveDisson: A good idea. But I do not understand how I can check it out?

pepibumur: Some pseudocode of how I think you could achieve it:

```swift
let objects = try objects()
objects.projects.first({ $0.mainGroup == group })
```
It should be fast because objects typically contain one project. Let me know if that works :)

Sources/xcodeproj/Objects/Files/PBXFileElement.swift
```diff
@@ -141,17 +145,33 @@ public extension PBXFileElement {
...
-            guard let filePath = self is PBXVariantGroup ? try baseVariantGroupPath() : path else { return groupPath }
-            return groupPath + filePath
+            let groupPath: Path?
+            
+            if let group = parent {
+                groupPath = try group.fullPath(sourceRoot: sourceRoot) ?? sourceRoot
+            } else {
+                let projectObjects = try objects()
+                let isThisElementRoot = projectObjects.projects.values.first(where: { $0.mainGroup == self }) != nil
+                if isThisElementRoot {
```
CognitiveDisson: @pepibumur  added check according to your advice

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    override func tearDown() {
+        super.tearDown()
+        try? tempDirectory.delete()
+    }
+    
+    func test_measureProjectRead() {
+        let duration = try! measureDuration {
+            _ = try XcodeProj(path: self.projectPath!)
+        }
+        XCTAssertLessThan(duration, 3)
```
CognitiveDisson: Values ​​with a large margin for independence from hardware. If something goes wrong, the values ​​will increase exponentially and it will be noticeable.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    func test_measureObtainFilesFullPath() {
+        let xcodeproj = try! XcodeProj(path: self.projectPath!)
+        var pathList = [Path]()
+        var count = 0
+        let duration = try! measureDuration {
+            for project in xcodeproj.pbxproj.projects {
+                for target in project.targets {
+                    for buildPhase in target.buildPhases {
+                        let files = buildPhase.files
+                        for file in files {
```
CognitiveDisson: @pepibumur Here is the problem. Maybe you can tell me how to solve it.
Array of files is empty. I found out that when I save a project file, an object is released for all the build files. ( `throw PBXObjectError.objectsReleased` ). But I did not understand the reason. Most likely some kind of error when generating.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+                            count += 1
+                        }
+                    }
+                }
+            }
+        }
+        XCTAssertLessThan(duration, 0.1)
+    }
+    
+    func test_measureProjGeneration() {
```
CognitiveDisson: @pepibumur And one more problem. Generation of the project is very long. (more than 10 minutes) 
The reason is another performance issue.
`PBXGroup func addFile(...)` This method takes a very long time. the more files the longer it takes.
This is due to this logic:
```
if let existingFileReference = try projectObjects.fileReferences.first(where: {
            try filePath == $0.value.fullPath(sourceRoot: sourceRoot)
 })
```
If I remove this logic, the generation will take less than half a minute.
A very long time to sort through all the files. Maybe there are some alternatives? Alternatively, we can create an utility dictionary where the key is the full path. What do you think about this option?

But perhaps this logic is not needed at all.
I do not fully understand why we change the path of the existing fileReference and why create a new fileReference if there was already an old fileReference.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+        XCTAssertLessThan(duration, 60)
+    }
+    
+    private func validate(xcodeproj: XcodeProj) {
+        let pbxproj = xcodeproj.pbxproj
+        let settings = XcodeProjGenerator.GenerateSetting()
+        let project = pbxproj.projects.first!
+        XCTAssertEqual(project.targets.count, settings.targetCount)
+        let expectedGroupCount = 1701
```
CognitiveDisson: I set the values, but they can be calculated. Can I leave it like this or better fix it?

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
+import Foundation
+import PathKit
+@testable import xcodeproj
+
+final class XcodeProjGenerator {
+    
+    struct GenerateSetting {
```
CognitiveDisson: These parameters allow us to generate a fairly large project - approximately 9 megabytes.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+        let targetCount: Int = 5
+        let groupCountInLevel: Int = 4
+        let fileCountInLevel: Int = 10
+        let maxGroupDepth: Int = 4
+    }
+    
+    @discardableResult
+    func generate(path: Path, sourceRoot: Path, setting: GenerateSetting = GenerateSetting()) throws -> XcodeProj {
```
CognitiveDisson: I think generation is much more flexible and better for performance tests than downloading a finished project.

CognitiveDisson:
@pepibumur I added some edits, as well as added performance tests. Is this a good idea?
But I ran into some problems (now these tests will not pass).
Maybe you can help me? I asked a few questions in the comments.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
pepibumur: Would you mind adding that as a comment?

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+            try xcodeproj.write(path: self.projectPath!)
+        }
+        XCTAssertLessThan(duration, 5)
+    }
+    
+    func test_measureObtainFilesFullPath() {
+        let xcodeproj = try! XcodeProj(path: self.projectPath!)
+        var pathList = [Path]()
+        var count = 0
+        let duration = try! measureDuration {
```
pepibumur: I'd flatten this a bit to facilitate readability. What about something like?

```swift
let files = xcodeproj.pbxproj.projects
  .flatMap({ $0.targets })
  .flatMap({ $0.buildPhases })
  .flatMap({ $0.files })
```

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+            try xcodeproj.write(path: self.projectPath!)
+        }
+        XCTAssertLessThan(duration, 5)
+    }
+    
+    func test_measureObtainFilesFullPath() {
+        let xcodeproj = try! XcodeProj(path: self.projectPath!)
+        var pathList = [Path]()
+        var count = 0
```
pepibumur: I think we can get rid of this one.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
pepibumur: The `addFile` method is intended to be used exceptionally when someone needs to add a single file to an existing group. It makes the addition of files convenient but at the cost of a not so good performance. Projects that want to have a high-performance project generation should not use those convenient methods and modify the objects directly.

I suggest changing the approach and adding a comment that states that the method shouldn't be used for adding a large list of files.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
pepibumur: This is a good idea. Thanks for bringing it into the project @CognitiveDisson

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
pepibumur: I agree with you here. It's more reliable as well since we don't need to pull something from the internet anymore.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+        let workspace = XCWorkspace()
+        let buildConfigurationList = XCConfigurationList()
+        let pbxproj = PBXProj(rootObject: nil, objectVersion: 0, archiveVersion: 0, classes: [:], objects: [])
```
pepibumur: I'd suggest creating the `buildConfigurationList` right after creating the `pbxproj` and then adding it to the `pbxproj` *(line 28)*

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
+import Foundation
+import PathKit
+import SwiftShell
+@testable import xcodeproj
+import XCTest
+
+final class MeasurementsTests: XCTestCase {
+    
+    let tempDirectory = Path.temporary
+    var projectPath: Path?
```
pepibumur: We can use explicitly unwrapped optionals here. It's safe to use them as long as we initialize them in the `setUp` method.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
+import Foundation
+import PathKit
+import SwiftShell
+@testable import xcodeproj
+import XCTest
+
+final class MeasurementsTests: XCTestCase {
+    
+    let tempDirectory = Path.temporary
```
pepibumur: I'd assign the value to this one in the `setUp` method.

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    var projectPath: Path?
+    var sourceRoot: Path?
+    
+    override func setUp() {
+        super.setUp()
+        try? tempDirectory.delete()
```
pepibumur: I'd rather create a unique temporary directory for each test:

```swift
let tempDirectory = Path.uniqueTemporary()
```

Tests/xcodeprojMeasurementsTests/MeasurementsTests.swift
```diff
@@ -0,0 +1,97 @@
...
+        self.sourceRoot = tempDirectory + Path("Sources")
+        let projectPath = self.projectPath!
+        let sourceRoot = self.sourceRoot!
+        let xcodeproj = try! generateXcodeProj(path: projectPath, sourceRoot: sourceRoot)
+        validate(xcodeproj: xcodeproj)
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        try? tempDirectory.delete()
```
pepibumur: I'd be more aggressive here with a `try tempDirectory.delete()`. If those temporary directories are not being deleted I prefer to be notified instead of leaving them in my local environment.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+    }
+    
+    private func fillProject(
+        pbxproj: PBXProj,
+        rootProject: PBXProject,
+        sourceRoot: Path,
+        mainGroup: PBXGroup,
+        setting: GenerateSetting) throws
+    {
+        try (0..<setting.targetCount).forEach { _ in
```
pepibumur:
```swift
for _ in 0..<setting.targetCount {

}
```

Which makes me think I should set up a linter in this project.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+            )
+            generateBuildPhase(
+                pbxproj: pbxproj,
+                target: target,
+                files: buildFiles
+            )
+        }
+    }
+    
+    private func generateTargets(pbxproj: PBXProj) -> PBXTarget {
```
pepibumur: `generateTarget` *(singular)*

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+        rootProject: PBXProject,
+        sourceRoot: Path,
+        mainGroup: PBXGroup,
+        setting: GenerateSetting) throws
+    {
+        try (0..<setting.targetCount).forEach { _ in
+            let target = generateTargets(pbxproj: pbxproj)
+            rootProject.targets.append(target)
+            var buildFiles = [PBXBuildFile]()
+            let currentPath = try mainGroup.fullPath(sourceRoot: sourceRoot)
```
pepibumur: I'd calculate the path based on how the project is being generated instead of using those convenient methods.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+            
+            let childGroupPath = currentPath + Path(groupName)
+            try childGroupPath.mkpath()
+            
+            try (0..<fileCount).forEach { _ in
+                let fileName = UUID().uuidString
+                let filePath = childGroupPath + Path(fileName)
+                try filePath.write(fileName)
+                let file = try group.addFile(at: filePath, sourceRoot: sourceRoot)
+                let buildFile = PBXBuildFile(file: file)
```
pepibumur: I'd always add the objects to the project immediately after they get created.

Tests/xcodeprojMeasurementsTests/XcodeProjGenerator.swift
```diff
@@ -0,0 +1,144 @@
...
+    
+    private func generateTargets(pbxproj: PBXProj) -> PBXTarget {
+        let target = PBXNativeTarget(name: UUID().uuidString)
+        pbxproj.add(object: target)
+        return target
+    }
+    
+    @discardableResult
+    private func generateBuildPhase(pbxproj: PBXProj, target: PBXTarget, files: [PBXBuildFile]) -> PBXSourcesBuildPhase {
+        let buildPhase = PBXSourcesBuildPhase(
```
pepibumur: Some of those attributes have default values so you can simplify this initialization a lot:

```swift
let buildPhase = PBXSourcesBuildPhase(files: files)
```

pepibumur:
Hey @CognitiveDisson 
Excellent work on adding those performance tests. The error that you were getting is most likely related to objects not being added to the `pbxproj`. My suggestion is to check where projects file are created and add them to to the project immediately after the initialization.

CognitiveDisson:
@pepibumur All checks have passed. Please, can you find out if new tests have been performed?
If yes, then it was the last edits.

I created a new class - BatchUpdater, it can be used for further improvements. For example, quickly add a specific file type to a project (without switch) - alternative to the `BPXObjects.addObject` method.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,84 @@
...
+                    return (fullPath, $0.key)
+                })
+            )
+            references = objectReferences
+        }
+        
+        if let existingObjectReference = objectReferences[filePath],
+            let existingFileReference = projectObjects.fileReferences[existingObjectReference] {
+            if !group.childrenReferences.contains(existingObjectReference) {
+                existingFileReference.path = groupPath.flatMap({ filePath.relative(to: $0) })?.string
```
pepibumur: I like the idea of the batch updater lazily creating a variable that allows us to get the object file reference given a path. However, I think this is not quite exactly the way Xcode adds files. Let's say we have the following scenario:

- Group: 
  - Name: `Sources`
  - Path: `/path/to/project/Sources`

- File we want to add:
  - Name: `SearchViewController.swift`
  - Path: `/path/to/project/Sources/Features/SearchViewController.swift`

If I call `addFile` passing the sources group and the absolute path to the `SearchViewController.swift` file. I'll end up getting the following on the Xcode project:

- Sources
   - SearchViewController.swift

When what I'd expect is the following:

- Sources
  - Features
    - Search
      - SearchViewController

pepibumur: I'd adjust the logic here a little bit to be like this *(continuing with the example from the comment above)*.

I'd take the absolute path of the file that I'd like to add, in this case `/path/to/project/Sources/Features/SearchViewController.swift` and I'dstart removing components from the right until we find a group that is part of the project. We'd find that `/path/to/project/Sources/` is already part of the project so we take that group and start creating the children down from that group. It'd result in the following:

- A `Features` group child of `Sources`.
- A `Search` group child of `Features`.
- A `SearchViewController.swift` file reference child of `Search`.

pepibumur:
Hey @CognitiveDisson. I really like the ideas that you are bringing into the project on this PR but I'm starting to feel that it's getting out of its original scope. What do you think if we take the performance tests and the batch updater out of this PR so we can merge this one, and continue the iterations on the tests and the batch updater on a different PR?

The batch updater is a very powerful idea for project generators because they wouldn't need to implement that logic themselves. For instance, I have to implement a similar logic for [Tuist](https://github.com/tuist/tuist).

CognitiveDisson:
Sure. I wanted this initially, but was too lazy to create a new PR.

PR #373, opened 2019-02-24
Change writing strategy

pepibumur:
Resolves https://github.com/yonaskolb/XcodeGen/issues/483
Resolves https://github.com/tuist/tuist/issues/243
Related https://github.com/tuist/xcodeproj/pull/327

### Short description 📝
Change the writing strategy to:
- Not write anything if the content hasn't changed.
- Replace the file instead of writing directly the content to the file.

### Solution 📦
- Add a new util that implements the new writing strategy.
- Update write functions to use the new util.

### DIFF
--- a/Sources/xcodeproj/Utils/FileWriter.swift
+++ b/Sources/xcodeproj/Utils/FileWriter.swift
@@ -3,7 +3,6 @@ import PathKit
 
 /// This protocol defines the interface to write project files to disk.
 protocol FileWriting {
-    
     /// Gracefully writes project data to disk:
     ///  1. If the content hasn't changed, it doesn't override the existing file.
     ///  2. If the content has changed, it overrides the existing file.
@@ -11,9 +10,22 @@ protocol FileWriting {
     /// - Parameters:
     ///   - data: Data to be writen.
     ///   - to: Path where the data will be written to.
-    /// - Throws: An errof if the write fails
+    /// - Throws: An error if the write fails
     /// - Returns: True if the file was writen.
+    @discardableResult
     func write(data: Data, to: Path) throws -> Bool
+
+    /// Gracefully writes project strings to disk:
+    ///  1. If the content hasn't changed, it doesn't override the existing file.
+    ///  2. If the content has changed, it overrides the existing file.
+    ///
+    /// - Parameters:
+    ///   - string: String to be writen.
+    ///   - to: Path where the data will be written to.
+    /// - Throws: An error if the write fails
+    /// - Returns: True if the file was writen.
+    @discardableResult
+    func write(string: String, to: Path) throws -> Bool
 }
 
 class FileWriter: FileWriting {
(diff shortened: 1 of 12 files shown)


Sources/xcodeproj/Utils/FileWriter.swift
```diff
@@ -0,0 +1,81 @@
...
+    ///
+    /// - Parameters:
+    ///   - data: Data to be writen.
+    ///   - to: Path where the data will be written to.
+    /// - Throws: An error if the write fails
+    /// - Returns: True if the file was writen.
+    @discardableResult
+    func write(data: Data, to: Path) throws -> Bool {
+        var existingData: Data?
+        if to.exists {
```
pepibumur: I'm not 100% sure about this one because we are loading the file into memory to do the comparison. I wonder how negative would be the impact on the memory footprint. Is there any chance you can try it with your project at Lyft @keith ?

The alternative that I considered for this one was writing the data to a temporary directory and compare the checksum of the two files. For that I'd need o some shell abstraction *(compatible with Linux)* to run the `md5` command. What do you think @yonaskolb?

keith: Yea in our case we wouldn't want this since we implement our own comparison checking and only regenerate the project if we need to in the first place. So whenever we run XcodeGen we know it needs to write it. We do this by diffing git ls-files and our project spec.

pepibumur: What about making this opt-in feature? @yonaskolb XcodeGen could have support for a flag for changing the writing strategy.

yonaskolb:
I agree that reading every single file that is written would be a bit heavy. For the problem this is trying to fix though maybe just doing that on the workspace would solve it.

Having said that I'm trying to reproduce the actual issue and show the alert in Xcode and I can't.
screen shot 2019-01-08 at 6 42 20 pm

It used to show this alert when generating. Now I can only shown an alert if trying to save changes after a regeneration
screen shot 2019-02-26 at 11 04 07 pm

It's unclear to me when Xcode actually shows a warning

pepibumur:
> I agree that reading every single file that is written would be a bit heavy

I'll keep the override flag and change the write logic to not make any comparison. I noticed that [CocoaPods' xcodeproj](https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project.rb#L359) moves the file (overriding) instead of deleting and writing the file. I think that should help with not getting Xcode confused about project changing in the file system.

> It's unclear to me when Xcode actually shows a warning

😓 I'll investigate it further

pepibumur:
Some observations that I made:
- When the project is replaced, Xcode can reload the project perfectly. XcodeGen [is already doing that](https://github.com/yonaskolb/XcodeGen/blob/master/Sources/XcodeGenKit/FileWriter.swift) so I'm assuming it's working fine.
- When the file that is replaced is a workspace, Xcode doesn't handle the reload gracefully.

I'm closing this PR leaving it up to the users of xcodeproj to define the writing strategy they'd like for their projects. If we see the concern is brought up often, we can reconsider this.

yonaskolb:
Thanks for the investigation @pepibumur

PR #375, opened 2019-02-27
Add Swiftformat

pepibumur:
### Short description 📝
Add swiftformat and set up Danger to run it in a strict manner so that warnings are reported as errors.


yonaskolb:
Not sure I understand the reason for removing SwiftLint? SwiftFormat is great, and that's a good addition, but auto-formatting doesn't replaces the value of linting. It would just replace something like `swiftlint autocorrect`.
I usually use both tools in concert.

pepibumur:
> Not sure I understand the reason for removing SwiftLint? SwiftFormat is great, and that's a good addition, but auto-formatting doesn't replaces the value of linting. It would just replace something like `swiftlint autocorrect`.

That's a good point. The Dangerfile has swiftformat set up to run as a linter, but I guess since it's not its primary goal, it doesn't provide the same level of linter that Swiftlint does. Let me check if I can configure both without conflicting with each other.

pepibumur:
Added Swiftlint back and fixed all the warnings and errors that we had in the project. Let mek now what you think @yonaskolb

Rakefile
```diff
@@ -4,9 +4,6 @@ require 'fileutils'
 require 'colorize'
 
 task :release_check do
-  puts "Linting code".colorize(:cyan)
-  system("swiftlint") || abort
```
yonaskolb: Do we want to keep this?

pepibumur: I think it's fine to get rid of it since we are running the linter on the PRs before they get merged.

PR #383, opened 2019-03-01
FileElement fullpath optimization

CognitiveDisson:
I significantly speed up the full path obtaining , by storing the parent reference.

Was:

```
guard let group = projectObjects.groups.first(where: { $0.value.childrenReferences.contains(reference) }) else { return sourceRoot } // problem
```
Now:

```
guard let group = parent else { return sourceRoot }
guard let groupPath = try group.fullPath(sourceRoot: sourceRoot) else { return nil }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## next version
 
+### Fixed
+- Fixed file full path performance issue https://github.com/tuist/xcodeproj/pull/372 by @CognitiveDisson.
+
 ## 6.6.0
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


CognitiveDisson:
@pepibumur These are changes from the previous PR about the full path.  After that I will create a new PR with BatchUpdater. ( It is based on these changes ) And at the end I will add new PR with performance tests (need a BatchUpdater)

pepibumur:
Awesome 🎉, changes merged. The `BatchUpdater` would be another great improvement that users `xcodeproj` would benefit a lot from. Glad to help you iterate on that feature on a follow-up PR.

PR #388, opened 2019-03-07
BatchUpdater

CognitiveDisson:
Implemented primary logic to update with batch. In the future it can be improved and expanded.

I agree with [your comment](https://github.com/tuist/xcodeproj/pull/372#discussion_r260750359), but I have not yet added functionality for creating groups, as this slows down performance. Instead, it added the creation of a file located directly inside the group folder. I think this is already quite convenient functionality. If such functionality is still needed, then I have to think how to add it.


CognitiveDisson:
@pepibumur everything is good with the PR? Maybe I need something to fix or add?

Sources/xcodeproj/Objects/Files/PBXGroup.swift
```diff
@@ -160,7 +160,15 @@ public extension PBXGroup {
         let groupPath = try fullPath(sourceRoot: sourceRoot)
 
         if let existingFileReference = try projectObjects.fileReferences.first(where: {
-            try filePath == $0.value.fullPath(sourceRoot: sourceRoot)
+            // Optimization: compare lastComponent before fullPath compare
+            guard let fileRefPath = $0.value.path else {
+                return try filePath == $0.value.fullPath(sourceRoot: sourceRoot)
+            }
+            let fileRefLPC = (fileRefPath as NSString).lastPathComponent
```
pepibumur: Nitpick: `fileRefLastPathComponent`. `LPC` is fairly ambiguous.

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -177,6 +177,20 @@ public extension PBXProj {
...
         objects.forEach(closure)
     }
+
+    /// This is a helper method for quickly adding a large number of files.
+    /// It is forbidden to add a file to a group one by one using the PBXGroup method addFile(...) while you are working with this class.
+    ///
+    /// - Parameters:
+    ///     - sourceRoot: source root path.
+    ///     - closure: closure in which you get the updater and can call the add method on it.
+    func batchUpdate(sourceRoot: Path, closure: (PBXBatchUpdater) throws -> Void) throws {
```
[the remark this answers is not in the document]
CognitiveDisson: Yea, for some reason, the linter remove public

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+    ///   - sourceRoot: path to project's source root.
+    /// - Returns: new or existing file and its reference.
+    @discardableResult
+    public func addFile(
+        to group: PBXGroup,
+        fileName: String,
+        sourceTree: PBXSourceTree = .group
+    ) throws -> PBXFileReference {
+        let groupPath = try group.fullPath(sourceRoot: sourceRoot)
+        let filePath = (groupPath ?? sourceRoot) + Path(fileName)
```
pepibumur: I think this fallback to `sourceRoot` is not obvious. I'd be strict and assume that `fullPath` returns a path. If it doesn't, that probably indicates that the project has been misconfigured and thus, it's something to be looked at at their end.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        guard filePath.exists else {
```
pepibumur: Aligned with @mxcl's suggestion, we can make the API throw less by assuming the developer knows that it's doing. I'd expect the user of the API to check whether the file exists or not. If we build on that assumption, I don't think we need the check here.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+
+// This is a helper class for quickly adding a large number of files.
+// It is forbidden to add a file to a group one by one using the PBXGroup method addFile(...) while you are working with this class.
+public final class PBXBatchUpdater {
+    private let projectObjects: PBXObjects
+    private let sourceRoot: Path
+    private var references: [Path: PBXObjectReference]?
+
+    init(projectObjects: PBXObjects, sourceRoot: Path) {
+        self.projectObjects = projectObjects
```
pepibumur: You can call it just `objects`.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        // Lazy initialization
+        let objectReferences: [Path: PBXObjectReference]
+        if let references = self.references {
+            objectReferences = references
+        } else {
+            objectReferences = Dictionary(uniqueKeysWithValues:
+                try projectObjects.fileReferences.compactMap({
+                    guard let fullPath = try $0.value.fullPath(sourceRoot: sourceRoot) else { return nil }
```
pepibumur: Similar comment here. What about using `!` and assuming that all the objects will have a full path?

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+        guard filePath.exists else {
+            throw XcodeprojEditingError.unexistingFile(filePath)
+        }
+
+        // Lazy initialization
+        let objectReferences: [Path: PBXObjectReference]
+        if let references = self.references {
+            objectReferences = references
+        } else {
+            objectReferences = Dictionary(uniqueKeysWithValues:
```
pepibumur: Have you measured how expensive this is for a large project?

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+    ) throws -> PBXFileReference {
+        let groupPath = try group.fullPath(sourceRoot: sourceRoot)
+        let filePath = (groupPath ?? sourceRoot) + Path(fileName)
+
+        guard filePath.exists else {
+            throw XcodeprojEditingError.unexistingFile(filePath)
+        }
+
+        // Lazy initialization
+        let objectReferences: [Path: PBXObjectReference]
```
pepibumur: Nitpick: Instead of adding a comment, we can have a method with that name:

```swift
fileprivate func lazilyInstantiateObjectReferences()
```

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+                    return (fullPath, $0.key)
+            }))
+            references = objectReferences
+        }
+
+        if let existingObjectReference = objectReferences[filePath],
+            let existingFileReference = projectObjects.fileReferences[existingObjectReference] {
+            if !group.childrenReferences.contains(existingObjectReference) {
+                existingFileReference.path = groupPath.flatMap({ filePath.relative(to: $0) })?.string
+                group.childrenReferences.append(existingObjectReference)
```
pepibumur: Do we want to return after this?

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+                    guard let fullPath = try $0.value.fullPath(sourceRoot: sourceRoot) else { return nil }
+                    return (fullPath, $0.key)
+            }))
+            references = objectReferences
+        }
+
+        if let existingObjectReference = objectReferences[filePath],
+            let existingFileReference = projectObjects.fileReferences[existingObjectReference] {
+            if !group.childrenReferences.contains(existingObjectReference) {
+                existingFileReference.path = groupPath.flatMap({ filePath.relative(to: $0) })?.string
```
pepibumur: Why would we need this?

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
```diff
@@ -0,0 +1,83 @@
...
+        case .group:
+            path = groupPath.map({ filePath.relative(to: $0) })?.string
+        case .sourceRoot:
+            path = filePath.relative(to: sourceRoot).string
+        case .absolute:
+            path = filePath.string
+        default:
+            path = nil
+        }
+        let fileReference = PBXFileReference(
```
pepibumur: I'd expect a different behavior than the one implemented here. With the following scenario:

- Existing group `Sources` at path `/path/to/project/Sources`.
- File to add: `/path/to/project/Sources/MyFeature/MyFeatureViewController.swift`

It'd result on a file reference child of the `Sources` group and with relative path `MyFeature/MyFeatureViewController.swift`. I'd expect though a `MyFeature` group child of `Sources` that has a child, `MyFeatureViewController.swift`

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
CognitiveDisson: Similarly `PBXGroup addFile`

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
CognitiveDisson: The method takes the file name, not the path. Therefore, you can add only a child file in this group.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
CognitiveDisson: I fully agree, this test should not be here. For Xcode, this is a valid case to have a reference to the file, but the file itself may not exist.

I also need to remove this check in the PBXGroup addFile method.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
pepibumur: If we add the reference to the group I don't think there's anything else to do.

pepibumur:
@CognitiveDisson [this comment](https://github.com/tuist/xcodeproj/pull/388/files#r264560491) regarding the purpose of the method made me think that we could make the method more useful and support adding a tree of files/groups.

In my experience adding files to a group, I think it makes sense for the consumer to keep track of the files that are being added. I can can append the file name to the group absolute path and add the file as a child.

I think it'd have far more value if we'd support adding a list of files from any sub-directory. For example, let's say we have a group at the path `/path/to/project/Sources`. And we'd like to add the following files:

- `/path/to/project/Sources/AppDelegate.swift`.
- `/path/to/project/Sources/Main/MainViewController.swift`
- `/path/to/project/Sources/Utilities/Keychain/KeychainController.swift`
- `//path/to/project/Sources/Storyboards/Main.storyboard`

If we created all the intermediate groups automatically we'd make adding files a very convenient process. I had to implement that for [tuist](https://github.com/tuist/tuist) and it was a lot of work. [Here](https://github.com/tuist/tuist/blob/master/Sources/TuistKit/Generator/ProjectFileElements.swift#L141) is the logic.

Sources/xcodeproj/Objects/Project/PBXProj.swift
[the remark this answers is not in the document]
CognitiveDisson: Oy. I did not notice, this is a public extension, there is no need to specify the public for the method.

Sources/xcodeproj/Utils/PBXBatchUpdater.swift
CognitiveDisson: I agree, so I do not understand why the `fullPath` method returns the optional value.

Sources/xcodeproj/Objects/Files/PBXGroup.swift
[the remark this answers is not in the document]
CognitiveDisson: I will change in force unwrap, because I try to follow the general style and rules of this repository.
But I would like to convey my point of view.

I believe that using force unwrap is a bad style for a user product (the auxiliary library is also intended for the user (another developer)). There are many articles, discussions and linter rules to prove this (for example: Swiftlint have this type of rule). Of course there is an alternative point of view.

I'll tell you why this is bad for me as a user of this library. I use this library as a dependency of my utility. My utility is a command line tool. If something went wrong, I want the user to get a distinct error. Therefore, I carefully check all parameters and extreme cases, pick up an error and display it. Most of the time I use exceptions for this. Since it provides a convenient mechanism for forwarding errors through the call stack. And they also allow you to add folbek, if this is an expected problem (ideally, of course, there should be no such problems). By the same principle, system methods work, for example methods of working with the file system.

When a force failure happens in some kind of dependency, I cannot catch it, I cannot display an error to the user. He will only see that the work was interrupted without giving a reason.
If this were an exception, then I would catch this situation and report that some part of the work could not be done, because your project file may be corrupted.

CognitiveDisson:
@pepibumur I implemented both approaches. And covered it with tests.

Sources/xcodeproj/Objects/Files/PBXGroup.swift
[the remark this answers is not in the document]
pepibumur: @CognitiveDisson that's a totally valid point of view. I agree with you that being explicit by raising errors helps the consumer of the library catch those errors and therefore show a nicer message to the users: "hey! this unexpected thing happened?"

However, I think depending of the type of error it should be treated differently. I can think of two scenarios when the last path component of the path can be nil:
- When there's a bug on xcodeproj's implementation.
- When something has be wrongly configured in the Xcode project.

If it's the first scenario, I don't think it's a good idea to expose the consumer of the API to those errors. We should instead have tests that can catch implementation bugs. If it's the second scenario, we shouldn't design an API assuming that the user might not know what they are doing. We can design the API to prevent a bad usage, but there's some responsibility that lays on the consumer's side.

I think Core Data is a good example of this. I've seen many developers complaining about Core Data not being thread safe and that resulting in their apps crashing. Core Data needs to be understood to be used properly. Try to use Core Data in unsafe ways... you'll come across assertions that crash your apps. Should they then expose those as errors instead? I don't think so.

Having said that, there's some work on our side to be done:
- We need more documentation that expalins users how things work.
- There's still room for improvement on making the API safer.

pepibumur:
That's great @CognitiveDisson. Also, thank you very much for bringing your thoughts on how the library should expose errors. It's very valuable having another point of view that helps improve the API.
I'll try to review the PR later today.

CognitiveDisson:
@pepibumur I have a little time to finish this. And these changes are important enough for me. Should I do something else so that it can be frozen?

pepibumur:
Just rebase from master and we are good to merge @CognitiveDisson

CognitiveDisson:
@pepibumur Hi. I merged master and fixed all problems. Pls merge it

Issue #385, opened 2019-03-05
Indent can be improved


mengxiangyue:
## Context 🕵️‍♀️
the sections content between `/* Begin XXX section */` and `/* End PBXBuildFile section */` dont have tab after modifying it to write into `project.pbxproj`. So causes many diffs in git.

## Proposal 🎉
I learn it, and find you create new `var tmpStateHolder = StateHolder()` and dont set the `indent` property in `PBXProjEncoder.swift` (line 323). It works well after i set the value 2 `tmpStateHolder.indent = 2`.
could you check it and add it into the project if i correct.

mengxiangyue:
you can try it by using `https://github.com/mengxiangyue/TestFramework`, then you can run the `command-tool` project, then you will get these below diffs:
![image](https://user-images.githubusercontent.com/1563329/53872436-39e2f700-4039-11e9-9e50-c28fac1f9bc3.png)

mxcl:
I’ve been seeing this too, sorry for the lack of a report, it didn’t occur to me that it was a bug for some reason.

Dschee:
I'm seeing this exact same issue in [this tool](https://github.com/JamitLabs/Accio) I'm working on right now. Would be great if this got fixed soon!

pepibumur:
Found the issue. I've started fixing it on [this PR](https://github.com/tuist/xcodeproj/pull/391). Thanks to the test that I added I found out that there are issues with schemes and workspaces as well. I'll fix all of them and then merge the branch.

pepibumur:
By the way, really cool idea @Dschee. Happy to see xcodeproj helping you making it real 😛

Dschee:
Yes, it was much easier to accomplish thanks to xcodeproj. But there are two more things I noticed and I‘m not sure if these are tracked in issues yet (I could open them if not):

1. The xcodeproj generated identifiers when I add new files or geoups in my projects are longer than the ones generated by Xcode itself (32 vs 24). Why is that?

2. Why doesn‘t the project adhere to the community standard of using uppercased camel casing for the project name? It looks really weird to me to „import xcodeproj“ instead of say „import XcodeProj“.

pepibumur:
> 1. The xcodeproj generated identifiers when I add new files or groups in my projects are longer than the ones generated by Xcode itself (32 vs 24). Why is that?

[...]

Issue #356, opened 2018-12-13
Editing the run phases affects schemes


kdbertel:
## Context 🕵️‍♀️
This is a continuation of #355. I am trying to write a Swift CLI that will add a build phase to run SwiftLint on an arbitrary target in an arbitrary project.

## What 🌱
I have been able to add the phase, and have that show up and run in Xcode. However, it appears that something in the save is not also saving the scheme correctly. In some instances, the scheme disappears entirely from the project; in other cases, it loses the ability to run tests.

My code for doing things (minus front matter for getting arguments) is:
```
let projectPath = Path(projectFile)

guard projectPath.exists else {
    print("Cannot find project file \("\(projectPath)".bold)", to: &stderr)
    exit(EXIT_FAILURE)
}

guard let project = try? XcodeProj(path: projectPath) else {
    print("Cannot read project file \("\(projectPath)".bold)", to: &stderr)
    exit(EXIT_FAILURE)
}

let pbxproj = project.pbxproj

guard let target = pbxproj.nativeTargets.first(where: { $0.name == targetName}) else {
    print("Unable to find target \("\(targetName)".bold)", to: &stderr)
    exit(EXIT_FAILURE)
}

let swiftLintName = "Run SwiftLint"

guard target.buildPhases.compactMap({ $0 as? PBXShellScriptBuildPhase }).filter({ $0.name == swiftLintName }).count == 0 else {
    print("Phase to run SwiftLint already exists. Ignoring.")
    exit(EXIT_SUCCESS)
}

let swiftLintScript = """
if which swiftlint >/dev/null; then
    swiftlint
else
    echo "warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint"
fi
"""

let runSwiftlintPhase = PBXShellScriptBuildPhase(name: swiftLintName, shellScript: swiftLintScript)

pbxproj.add(object: runSwiftlintPhase)
target.buildPhases.append(runSwiftlintPhase)

do {
    try project.write(path: projectPath, override: true)
} catch let error {
    print("Unable to write project: \("\(error)".bold)", to: &stderr)
    exit(EXIT_FAILURE)
}
```

Is there anything else I need to do to the project to guarantee that the scheme remains the same, and that all that's changing is the run phase?

(I have tried doing a diff between the SPM-generated project folder and the one touched by my script, but it ended up being 4300 lines long, the first 1000 or so removed from the `pbxproj` and the other 3000 added, or else I would paste it here)

[...]

PR #391, opened 2019-03-14
Fix diff issues

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/385
Resolves https://github.com/tuist/xcodeproj/issues/356

### Short description 📝
As it was reported [here](https://github.com/tuist/xcodeproj/issues/385) there was a bug in xcodeproj that caused some unexpected diffs after writing the project.

### Solution 📦
- [x] Fix the but that was introduced [here](https://github.com/tuist/xcodeproj/commit/6a96469cd2f909ba82358a2a15e3c89ba4141689#diff-6300a20d58094320a2e02f75bed1b086)
- [x] Remove integration test that pulls a project from the network.
- [x] Add a new integration test that uses a fixture project.


pepibumur:
There are issues with workspaces and schemes as well.

Issue #396, opened 2019-03-31
Version 7.0.0


pepibumur:
## Context 🕵️‍♀️
After several iterations of xcodeproj, there's still room for improvement in its API. Developers have shared issues and feedback that we can use to iterate on the 7th major version of the library. On this issue, I'm listing all the improvements/fixes that I think should make it into the next major version:

## What 🌱
- [x]  [**Change generated identifiers length to 24 (like Xcodes)**](https://github.com/tuist/xcodeproj/issues/392): Continuing with the effort of aligning the library closely to Xcode's behavior, we should use use the same length as Xcode for object unique identifiers.
  - PR: https://github.com/tuist/xcodeproj/pull/432
- [x]  **Swift 5 support**
  - Pull request: https://github.com/tuist/xcodeproj/pull/397/files
- [x]  [**Use upper camel case for project module name**](https://github.com/tuist/xcodeproj/issues/393): Follow the convention that most libraries stick to, using upper camel case for module names.
  - Pull request: https://github.com/tuist/xcodeproj/pull/398
- [x] [**Add support for RemoveRunnable**](https://github.com/tuist/xcodeproj/issues/276): Schemes can't specify a <RemoteRunnable> in the <LaunchAction>, and thus can't launch Sticker Packs.
  - PR: https://github.com/tuist/xcodeproj/pull/400
- [x] [**Auto generation of schemes**](https://github.com/tuist/xcodeproj/issues/115)
  - Pull request: https://github.com/tuist/xcodeproj/pull/399
- [x] Xcode 11 support

pepibumur:
When we publish this version we'll need to make sure that [we let CocoaPods know](https://github.com/CocoaPods/cocoapods-trunk/issues/31) about the deprecation of the `xcodeproj` in favor of `XcodeProj.

PR #403, opened 2019-04-07
Reduce the amount of throwing

pepibumur:
Related https://github.com/tuist/xcodeproj/issues/396

### Short description 📝
The current API was designed on the assumption that developers might misuse the API. To let them know about that, we used the Swift error throwing extensively. That resulted in many public methods throwing, some of which don't really need to.

### Solution 📦
I've gone through all the throws and remove those that are not strictly necessary.

Note that we are introducing a new assumption to the library, some methods should only be called if they are part of a project, otherwise the execution will abort. To prevent that from happening, the next step will be preventing developers from creating objects that don't belong to any project.


Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
```diff
@@ -10,7 +10,7 @@ public final class PBXBuildFile: PBXObject {
     /// Returns the file the build file refers to.
     public var file: PBXFileElement? {
         get {
-            return fileReference?.getObject()
+            return fileReference?.materialize()
```
pepibumur: I've taken the opportunity to rename this method. I've taken the idea from other libraries that you the same verb to describe the action of unwrapping a wrapped value.

Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
```diff
@@ -62,21 +62,19 @@ extension PBXBuildFile {
...
     ///
     /// - Returns: build phase type.
-    /// - Throws: an error if this method is called before the build file is added to any project.
-    func getBuildPhase() throws -> PBXBuildPhase? {
+    func getBuildPhase() -> PBXBuildPhase? {
         if let buildPhase = buildPhase {
             return buildPhase
         }
-        let projectObjects = try objects()
+        let projectObjects = objects()
```
pepibumur: Developers should not call these methods on objects that are not part of any project. If that happens, the execution will abort.

Sources/xcodeproj/Objects/Project/PBXObject.swift
```diff
@@ -71,12 +71,10 @@ public class PBXObject: Hashable, Decodable, Equatable, AutoEquatable {
...
     ///
     /// - Returns: objects the object belongs to.
-    /// - Throws: an error if this method is accessed before the object has been added to a project.
-    func objects() throws -> PBXObjects {
-        guard let objects = self.reference.objects else {
-            let objectType = String(describing: type(of: self))
-            throw PBXObjectError.orphaned(type: objectType, reference: reference.value)
-        }
-        return objects
+    func objects() -> PBXObjects {
```
pepibumur: We should not allow orphaned objects that don't belong to any project. We are not enforcing that from the API but such enforcement will land soon.

Sources/xcodeproj/Objects/Project/PBXObjectParser.swift
```diff
@@ -63,7 +63,7 @@ final class PBXObjectParser {
         case PBXBuildRule.isa:
             return try decoder.decode(PBXBuildRule.self, from: data)
         default:
-            throw PBXObjectError.unknownElement(isa)
+            preconditionFailure("unsupported object type")
```
pepibumur: Considering how often new objects are introduced to Xcode projects, I think it's safe to go with a `preconditionFailure` avoiding the need of having to catch errors.

Sources/xcodeproj/Project/WorkspaceSettings.swift
```diff
@@ -99,13 +97,12 @@ public class WorkspaceSettings: Codable, Equatable, Writable {
     ///
     /// - Parameter path: The path to write to
     /// - Parameter override: True if the content should be overriden if it already exists.
-    /// - Throws: writing error if something goes wrong.
-    public func write(path: Path, override: Bool) throws {
+    public func write(path: Path, override: Bool) {
         let encoder = PropertyListEncoder()
```
pepibumur: I've updated all the classes that can be publicly initialized with a path to assume that the developer has checked beforehand whether the path exists or not.

yonaskolb:
I have to say, I would disagree with a lot of the changes in here. While I understand the motivation of removing some of the friction in the public mutating apis, I think this goes too far. A program crashing if the developer made an obvious error is one thing, but crashing in the domain of reading or writing paths for example is often out of the developers control. The file system is fickle and all sort of issues can arise in regards to permissions, and the existence of certain files.

Plus with throwing methods gone we lose any chance of recovering from errors internally in different cases, the app will just crash. Other tools using this also have no way to recover from these errors, and you would force them to crash.

For an outside user is it so hard for a dev to type `try!` if they really don’t care about errors?

pepibumur:
> A program crashing if the developer made an obvious error is one thing, but crashing in the domain of reading or writing paths for example is often out of the developers control

This is understandable and I'll revert those.

> Plus with throwing methods gone we lose any chance of recovering from errors internally in different cases, the app will just crash. Other tools using this also have no way to recover from these errors, and you would force them to crash.

Throwing for reading/writing makes sense because the responses from the file system are out of the control of the library. However, I do disagree all methods should throw, including those that represent a bad usage of the library or a bug. If we take Core Data as an example, it immediately asserts the execution if developers badly lazily loaded relationships across threads. Developers can't recover from that, they need to read Core Data's documentation and use it properly. That's why besides this change, we plan to add more documentation and make the API safer.

mxcl:
My perspective when I suggested this project was a little “throwy” was to rethink some of the decisions. For example an xcodeproj is still valid if a path doesn't exist, this should not throw. If I (as a consumer of this library) want to enforce that a path doesn't exist I can check myself. Similarly if the file has already been added to a group, silently succeed, no need to throw, the desired result has already been achieved.

For other scenarios, eg. adding a link phase, ideally the type system should be used as much as possible to make it so errors just aren't possible. I think Vapor does this pretty well, `throw` exists in Vapor but most of the time it's just possible to write offending code due to the way generics are applied to the codebase.

mxcl:
Like if we examine this function:

```
public extension PBXBuildPhase {
    func add(file: PBXFileElement) throws -> PBXBuildFile {
        if let existing = files?.first(where: { $0.fileReference == file.reference }) {
            return existing
        }
        let projectObjects = try objects()
        let buildFile = PBXBuildFile(file: file)
        projectObjects.add(object: buildFile)
        files?.append(buildFile)
        return buildFile
    }
}
```

The `try objects()` is the only throwing portion, and to me (sorry if wrong) it seems like if this was to throw it would imply some sort of internal error inside `tuist/xcodeproj`, instead `try!` or much better: make it impossible for `objects()` to throw, inside that function (if not looked at it) we could instead ensure that the objects list exists if it doesn't, preventing error and invalid state.

Anyway, I have not read the patch in great detail, sorry for that, if I mentioned things already done I'm sorry for wasting any time. Just felt I should chime in.

pepibumur:
> For example an xcodeproj is still valid if a path doesn't exist, this should not throw. If I (as a consumer of this library) want to enforce that a path exists I can check myself or write an extension that checks before adding and throws an error I define.

That was my initial thought but @yonaskolb raised a good point there. Writes and reads can fail even after some basic checks from the user. 

> For other scenarios, eg. adding a link phase, ideally the type system should be used as much as possible to make it so errors just aren't possible. I think Vapor does this pretty well, throw exists in Vapor but most of the time it's just not possible to write code that would exist in an invalid state due to the way generics are applied to the codebase.

Haven't looked a Vapor but I think there's still a lot of room for improvement in the way we use types to make the API safer. One of the things that could be a source of errors is the facts that users can create objects that are not attached to any project. Some API methods are built on the assumption that objects belong to a project. When that's not met, it results on a crash *(with the refactor on this PR)*. Before rolling this out, I planned to make all constructors on individual objects internal, and expose them from the objects they have a relation with. For instance, if you want to add a new build phase, thi is how you'd currently do it:

```swift
// Before
let frameworksBuildPhase = PBXBuildPhase(....)
pbxproj.add(object: frameworksBuildPhase)
target.buildPhases.append(frameworksBuildPhase)

// After
let frameworksBuildPhase = target.addFrameworksBuildPhase(...)
```

The `try objects()` is a good example. We have that try there because users might try to use the method without adding the object to a project. With the change from above, that won't be possible and therefore we can reach that point and assume that objects will always be there.

It'd be great if you could go through the patch changes and add comments on the ones you are concerned about. We can discuss them individually.

pepibumur:
It'd be great if you could go through the patch changes and add comments on the ones you are concerned about so that  we can discuss them individually. This is an important change for the library so let's make sure we are all aligned with the direction we are taking.

Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
mxcl: I think this is the right choice if the choices are `throw` or `abort`, since throwing really should be for runtime errors and *not* programming errors (in my interpretation of how Apple does it). However I do agree with you that an ideal situation would not allow you to call this function at all if the object is not yet there, which could be done with a careful API.

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -212,16 +212,16 @@ extension PBXProj: Equatable {
...
 
 extension PBXProj: Writable {
-    public func write(path: Path, override: Bool) throws {
-        try write(path: path, override: override, outputSettings: PBXOutputSettings())
+    public func write(path: Path, override: Bool) {
+        write(path: path, override: override, outputSettings: PBXOutputSettings())
     }
 
-    public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) throws {
+    public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) {
```
mxcl: Probably, as said, this is *too* far, writes can definitely fail, and it is a “runtime error” rather than “programmer error” if that should happen.

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -212,16 +212,16 @@ extension PBXProj: Equatable {
...
         let encoder = PBXProjEncoder(outputSettings: outputSettings)
-        let output = try encoder.encode(proj: self)
+        let output = encoder.encode(proj: self)
         if override, path.exists {
-            try path.delete()
+            try! path.delete()
```
mxcl: Same with the bang here, if a delete failed I wouldn't want my app to crash. Deletes fail, it's the nature of filesystems, eg. permissions may be wrong.

Sources/xcodeproj/Project/WorkspaceSettings.swift
mxcl: Where the result is just a red-item in the open Xcodeproj, I think that's fine, though here I think (like before) writes can `throw` since they are runtime errors rather than programmer errors.

pepibumur:
Hey @mxcl, I've revisited all the interactions with the file system and made sure that they throw instead of failing the execution. Would you mind having another look?

pepibumur:
Would you mind having a look at this one @tuist/core 👀 ?

Sources/xcodeproj/Objects/Project/PBXProj.swift
```diff
@@ -218,7 +218,7 @@ extension PBXProj: Writable {
 
     public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) throws {
         let encoder = PBXProjEncoder(outputSettings: outputSettings)
-        let output = try encoder.encode(proj: self)
+        let output = encoder.encode(proj: self)
```
kwridan: For encoding / decoding throwing methods are expected I would have thought?

ollieatkinson: Since the function is still a `throws` function it makes sense to keep `try`

ollieatkinson: The try was removed because the throwing write function was removed - I'm thinking we need to revisit that to ensure it does still throw.

pepibumur: We are not exactly following the encoding/decoding approach that Apple follows with the decoders here. The custom decoder that we are using here just writes the Xcode-plist representation to a string that we end up writing to the disk. The only scenario that I can think of in which the encoder could error is if users refer to objects that don't belong to any project. 

To prevent that I'll make the API safer in a follow-up PR.

Is there any other scenario that you think I might be missing here?

Sources/xcodeproj/Objects/Files/PBXGroup.swift
```diff
@@ -157,10 +157,11 @@ public extension PBXGroup {
         override: Bool = true,
         validatePresence: Bool = true
     ) throws -> PBXFileReference {
-        let projectObjects = try objects()
+        let projectObjects = objects()
         if validatePresence, !filePath.exists {
-            throw XcodeprojEditingError.unexistingFile(filePath)
```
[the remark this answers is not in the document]
pepibumur: I removed the check entirely. I think that checking whether a file exists or not before adding it to the project should be done by the user.

pepibumur:
@tuist/core I've rebased from master and fixed a few tests that were failing. Would you mind giving it another look?

pepibumur:
I'm closing this PR. I slept over this idea and I think it'll result in a bad experience for the developer. 

XcodeProj opens misconfigured projects gracefully, meaning that even if there are broken relationships between objects, we are able to open the projects. However, writing those projects back will result in runtime crashes with the changes in this PR.

If there's any objection, please, don't hesitate to dump it on this PR. I'm happy to discuss this further.

@kwridan I'll fix that issue on a follow-up PR. It doesn't seem to be related to the changes in this PR.

Thanks to everyone involved in this PR, and sorry for not having recognized this sooner.

PR #407, opened 2019-04-22
Embed frameworks build phase

### LINKED ISSUE #406: Embed frameworks build phase is hard to access.
## Context 🕵️‍♀️
Trying to embed frameworks in an app target.
What I want to emulate is what happens when you drag and drop the frameworks to the `Xcode/Project/General/Embedded Binaries` section. Or when you click on the + in the same section.

## What 🌱
"Embed frameworks" is a target's build phase, precisely is a "Copy Files" build phase where the destination subfolder is Frameworks. It looks like Xcode fills the `XCode/General/Embedded Binaries` looking for all the build phases of this kind, i.e, "copy files" and "frameworks" destination. (I've tried to add more than one of this build phases, and XCode groups them in `XCode/General/Embedded Binaries`, no matter the name of the build phase.

## Proposal 🎉
Imitate what is done for the sources build phase or the resources build phase of a target, [here](https://github.com/tuist/xcodeproj/blob/12245cbba08c234c2d075a29ad2f42bd8a8d1572/Sources/xcodeproj/Objects/Targets/PBXTarget.swift#L201) and [here](https://github.com/tuist/xcodeproj/blob/12245cbba08c234c2d075a29ad2f42bd8a8d1572/Sources/xcodeproj/Objects/Targets/PBXTarget.swift#L213). We can add a method to that extension that retrieves the "Embed frameworks" build phase, something like `func embedFrameworksBuildPhase() -> [PBXSourcesBuildPhase]`. Its an array because you can have as many Copy Files + "frameworks" build phases as you want. No throws because embed frameworks are not mandatory (frameworks targets usually don't embed frameworks, apps targets may have no frameworks).

llinardos:
Resolves #406.

### Short description 📝
Allows fetching the Embed Frameworks build phases from a target.

### Solution 📦
Imitating the methods mentioned on the issue, I added a similar method to retrieve the embed frameworks build phases.

### Implementation 👩‍💻👨‍💻
- [ ] Given that embed framework build phases are copy file build phases with frameworks destination subfolder, the implementation is a filter over all the build phases: first filters the copy files build phases (`buildPhase == .copyFiles`), then cast them as `PBXCopyFilesBuildPhase`, finally filter for `dstSubfolderSpec == .frameworks`
- [ ] Add the tests for no build phases, no matching build phases, one matching build phase and more than one matching build phase.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,8 +17,8 @@ jobs:
       - run:
           name: Install Tools
           command: |
+            brew upgrade
             brew install swiftformat
-            brew install swiftlint
             bundle config build.sqlite3 --with-sqlite3-dir=/opt/local
       - run:
           name: Install dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - **Breaking** Swift 5 support https://github.com/tuist/xcodeproj/pull/397 by @pepibumur.
 - `WorkspaceSettings.autoCreateSchemes` attribute https://github.com/tuist/xcodeproj/pull/399 by @pepibumur
 - Additional Swift 5 fixes: https://github.com/tuist/xcodeproj/pull/402 by @samisuteria
+- Can access embed frameworks build phase for a target by @llinardos.
 
 ## 6.7.0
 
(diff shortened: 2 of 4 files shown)


pepibumur:
Could you update the CHANGELOG @llinardos ?

pepibumur:
@llinardos you can rebase from master and your build should pass.

Issue #411, opened 2019-04-29
unsync........?


gubikmic:
Hi, I haven't used this project in a while and see that there were some changes.

I think there should be a comment in PBXObjects.swift giving the rationale behind the unsync name. What's the reason behind this?
Also, why is it called 'unsync'? 'Unsync' is a verb, right? Shouldn't it be the adjective 'unsynced'?!

pepibumur:
Hey @gubikmic.

That's a totally valid concern. I missed this one when it was introduced, and I agree it's a bit ambiguous. All variables marked with `unsynced` are not thread safe. They are private and its access is serialized using locks.

I'm not quite sure about unsynced either. What about just underscoring the public name? I've seen that pattern in other places. I'm open to suggestions.

azav:
You are dealing with an abbreviation and “unsync” means unsynchronous (sp?) which is the nature of the action.  I am assuming it is being used as an adjective.  If it were an adverb, it would be unsynchronously. The verb of this is unclear to me, but synchronize is a verb.

PR #415, opened 2019-05-02
Rename some variables to have a more representative name

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/411

### Short description 📝
As @gubikmic pointed out [here](https://github.com/tuist/xcodeproj/issues/411), `unsync` is quite a vague and ambiguous prefix that makes it hard to understand what certain variables are for. 

### Solution 📦
Since variables prefixed by `unsync-` are not-thread-safe variables, I change the prefix to `notThreadSafe-` instead.


yonaskolb:
How about  just `_projects`?

PR #417, opened 2019-05-03
Add SWIFT_COMPILATION_MODE and CODE_SIGN_IDENTITY build settings

dangthaison91:
1. Add **SWIFT_COMPILATION_MODE**: 
- Debug: Incremental
- Release: Whole Module

2. Update **CODE_SIGN_IDENTITY**:
- Debug: iPhone Developer
- Release: iPhone Distribution

3. Fix **SWIFT_ACTIVE_COMPILATION_CONDITIONS** :
- Debug: "DEBUG"
- Actual Release: "DEBUG"
- Expected Release: "RELEASE"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Can access embed frameworks build phase for a target by @llinardos.
 - Added `com.apple.product-type.framework.static` to `PBXProductType`. https://github.com/tuist/xcodeproj/pull/347 by @ileitch.
 - Can add a not existing file to a group https://github.com/tuist/xcodeproj/pull/418 by @llinardos.
+- Add `SWIFT_COMPILATION_MODE` and `CODE_SIGN_IDENTITY` build settings, remove `DEBUG` flag for Release https://github.com/tuist/xcodeproj/pull/417 @dangthaison91 
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - Can access embed frameworks build phase for a target by @llinardos.
 - Added `com.apple.product-type.framework.static` to `PBXProductType`. https://github.com/tuist/xcodeproj/pull/347 by @ileitch.
 - Can add a not existing file to a group https://github.com/tuist/xcodeproj/pull/418 by @llinardos.
-- Add `SWIFT_COMPILATION_MODE` and `CODE_SIGN_IDENTITY` build settings, remove `DEBUG` flag for Release https://github.com/tuist/xcodeproj/pull/417 @dangthaison91 
+- **Breaking** Add `SWIFT_COMPILATION_MODE` and `CODE_SIGN_IDENTITY` build settings, remove `DEBUG` flag for Release https://github.com/tuist/xcodeproj/pull/417 @dangthaison91 
 
 ### Fixed
 
--- a/Sources/xcodeproj/Utils/BuildSettingsProvider.swift
+++ b/Sources/xcodeproj/Utils/BuildSettingsProvider.swift
@@ -45,8 +45,10 @@ public class BuildSettingsProvider {
         var buildSettings: [String: Any] = [:]
         if let platform = platform, platform == .iOS {
             buildSettings["SDKROOT"] = "iphoneos"
-            buildSettings["CODE_SIGN_IDENTITY"] = "iPhone Developer"
             buildSettings["LD_RUNPATH_SEARCH_PATHS"] = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks"
+            if let variant = variant {
+                buildSettings["CODE_SIGN_IDENTITY"] = variant == .debug ? "iPhone Developer" : "iPhone Distribution"
+            }
         }
         if let platform = platform, platform == .macOS {
             buildSettings["SDKROOT"] = "macosx"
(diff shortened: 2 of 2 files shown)


pepibumur:
@dangthaison91 could you update the changelog? We can merge afterwards 😛

kwridan:
good find @dangthaison91 ! Thanks for fixing it.

By default in Xcode when making a new project, the release configuration doesn't contain any `RELEASE` flags e.g.

Screen Shot 2019-05-03 at 7 36 40 AM

Is there a need to specify a build setting if it's the default specified by Xcode? e.g. 

In Debug `SWIFT_COMPILATION_MODE = singlefile` (Incremental) by default 

Screen Shot 2019-05-03 at 7 40 23 AM

dangthaison91:
@pepibumur @kwridan 
I just guess we want to add compile flag for Release also. It's OK to change **SWIFT_ACTIVE_COMPILATION_CONDITIONS** to empty for Release config?

For ** CODE_SIGN_IDENTITY**, IMHO "iPhone Distribution" for Release is fine because we often use it for archive.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Make build phase name public by @llinardos.
 - Can access embed frameworks build phase for a target by @llinardos.
 - Added `com.apple.product-type.framework.static` to `PBXProductType`. https://github.com/tuist/xcodeproj/pull/347 by @ileitch.
+- Add `SWIFT_COMPILATION_MODE` and `CODE_SIGN_IDENTITY` build settings, remove `DEBUG` flag for Release https://github.com/tuist/xcodeproj/pull/417 @dangthaison91
```
pedropinerashopify: Notice that this is a breaking change so I'd prefix the line with `**Breaking**`

dangthaison91: Moved it to **Breaking**

Sources/xcodeproj/Utils/BuildSettingsProvider.swift
```diff
@@ -67,10 +69,12 @@ public class BuildSettingsProvider {
         if let variant = variant, let swift = swift, variant == .debug, swift == true {
             buildSettings["SWIFT_OPTIMIZATION_LEVEL"] = "-Onone"
             buildSettings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = "DEBUG"
+            buildSettings["SWIFT_COMPILATION_MODE"] = "singlefile"
         }
         if let variant = variant, let swift = swift, variant == .release, swift == true {
             buildSettings["SWIFT_OPTIMIZATION_LEVEL"] = "-Owholemodule"
-            buildSettings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = "DEBUG"
+            buildSettings["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = ""
```
kwridan: thanks for the updates! 

we can skip setting this all together as `""` results in an empty string setting as opposed to not being set.
Screen Shot 2019-05-03 at 10 03 27 AM

dangthaison91: Removed as your suggestion @kwridan

kwridan:
Hi @dangthaison91
 
> For CODE_SIGN_IDENTITY , IMHO "iPhone Distribution" for Release is fine because we often use it for archive.

This is an interesting one - often Release is tied with archiving & distribution but isn't strictly necessary, for example when profiling the app on a device, you may choose to profile with the release configuration. 

Developers may not have the distribution certificate available to them locally, thus this would prevent them from profiling / archiving locally. This is often the case on bigger teams.

This setting can easily be overridden for each use case e.g. in [Tuist](https://github.com/tuist/tuist) each manifest can override this setting.

```swift
let release = Configuration(settings: ["CODE_SIGN_IDENTITY": "iPhone Distribution"])

let project = Project(name: "App",
                      targets: [
                          Target(name: "App",
                                 platform: .iOS,
                                 product: .app,
                                 bundleId: "io.tuist.App",
                                 infoPlist: "Info.plist",
                                 sources: "Sources/**",
                                 dependencies: [
                                    // ...
                                 ],
                                 settings: Settings(release: release)),
```

As such for `XcodeProj` which many tools are built on top of it may be best to keep the Xcode default of `iPhone Developer`. Open to hearing others thoughts on this who are using `XcodeProj` directly / indirectly in their tooling.

Thanks again for the updates and apologies for the back and forth.

pepibumur:
> As such for XcodeProj which many tools are built on top of it may be best to keep the Xcode default of iPhone Developer. Open to hearing others thoughts on this who are using XcodeProj directly / indirectly in their tooling.

I think it's fine defaulting to this one in this component because it's an opt-in feature of XcodeProj. If developers would like to use a different value for that setting, they can override or not use the settings provider.

Sources/xcodeproj/Utils/BuildSettingsProvider.swift
```diff
@@ -45,8 +45,10 @@ public class BuildSettingsProvider {
         var buildSettings: [String: Any] = [:]
         if let platform = platform, platform == .iOS {
             buildSettings["SDKROOT"] = "iphoneos"
-            buildSettings["CODE_SIGN_IDENTITY"] = "iPhone Developer"
             buildSettings["LD_RUNPATH_SEARCH_PATHS"] = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks"
+            if let variant = variant {
+                buildSettings["CODE_SIGN_IDENTITY"] = variant == .debug ? "iPhone Developer" : "iPhone Distribution"
```
pepibumur: Hey @dangthaison91, only one more little thing. The stance of XcodeProj when it comes to default values is staying as aligned as possible to Xcode's behavior. In the case of `CODE_SIGN_IDENTITY`, Xcode sets the value to `iPhone Developer` regardless of the build configuration.
```suggestion
                buildSettings["CODE_SIGN_IDENTITY"] = "iPhone Developer"
```

dangthaison91:
Hey @pepibumur @kwridan I restore `CODE_SIGN_IDENTITY` setting to `iPhone Developer` as it is Xcode default setting for both configuration.

pepibumur:
Awesome 🎉! PR merged

Issue #409, opened 2019-04-24
Can't add a (yet) not existent file to a group.


llinardos:
I'm trying to add a reference to a `.framework` that is not created yet. I'm sure the file will exist after compiling the target that generates it. But it doesn't exist yet.

The method [add file in PBXGroup](https://github.com/tuist/xcodeproj/blob/12245cbba08c234c2d075a29ad2f42bd8a8d1572/Sources/xcodeproj/Objects/Files/PBXGroup.swift#151) [check for existence of the file](https://github.com/tuist/xcodeproj/blob/12245cbba08c234c2d075a29ad2f42bd8a8d1572/Sources/xcodeproj/Objects/Files/PBXGroup.swift#L157), so it throws when I try to add the yet unexistent file.

I solve my issue adding the next method to PBXGroup:

```    
func addFileReference(_ fileReference: PBXFileReference) throws {
        let projectObjects = try objects()
        projectObjects.add(object: fileReference)
        fileReference.parent = self
        if !childrenReferences.contains(fileReference.reference) {
          childrenReferences.append(fileReference.reference)
        }
    }
```

I know what I'm doing, but I'm not sure if adding this method to the group API is ok. I mean, it allows to add unexistent files references. You guys know much more about the uses of the xcodeproj, what do you think? Is it worth to add it? If is it ok, I can make a PR for this feature.

pepibumur:
Hey @llinardos, that's a utterly valid use case.
A PR [was merged](https://github.com/tuist/xcodeproj/pull/410/files) recently modifying the `addFile` method to support overriding existing file. You could do add another argument to support adding non-existing files:

```swift
func addFile(validatePresence: Bool = true) // Default to true for backwards compatibility. 
```

llinardos:
Done @pepibumur! [Here](https://github.com/tuist/xcodeproj/pull/418) is the PR.

PR #418, opened 2019-05-03
Add not existent file to a group.

llinardos:
Adds a flag to validate or not file existance. Defaults to true for backwards compatibility.
Adds tests for not existing file with and without validate presence flag.

Resolves https://github.com/tuist/xcodeproj/issues/409


pepibumur:
Thank you very much @llinardos for this addition 🎉

llinardos:
I'm glad to contribute 😄

PR #425, opened 2019-05-20
Remote project support for PBXContainerItemProxy

damirdavletov:
### Short description 📝
PBXContainerItemProxy can reference objects defined either within the same .xcodeproj or within a remote one.

If the object is within the same _.xcodeproj_ as the proxy then **containerPortal** property contains **PBXProject** object and **remoteGlobalIDString** contains the ID of the object which _can_ be accessed through the normal object pool.
On the other hand if the referenced object is in remote _.xcodeproj_ then **containerPortal** property contains **PBXFileReference** of remote _.xcodeproj_ (where the object is located) and **remoteGlobalIDString** contains the ID of the object which _cannot_ be accessed through the normal object pool so has to accessed as a string.

Currently support for remote project is absent so I'm adding it in this pull request.

### Solution 📦
I added **containerPortalAsFileReference** and **remoteGlobalIDString** properties to be able to access the values.
Also **isContainerPortalFileReference** is introduced in order to check if **containerPortal** is a file reference or not.
Additionally I split the init method onto 2 convenience initialisers, one for local proxy and one for remote.

### DIFF
--- a/Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
+++ b/Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
@@ -1,5 +1,11 @@
 import Foundation
 
+public enum ContainerPortal {
+    case project(PBXProject)                /// Project where the proxied object is located in
+    case fileReference(PBXFileReference)    /// File reference to .xcodeproj where the proxied object is located
+    case unknownObject(PBXObject?)          /// This is used only for reading from corrupted projects. Don't use it.
+}
+
 /// This is the element to decorate a target item.
 public final class PBXContainerItemProxy: PBXObject {
     public enum ProxyType: UInt, Decodable {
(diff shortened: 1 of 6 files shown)


Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
```diff
@@ -8,10 +8,10 @@ public final class PBXContainerItemProxy: PBXObject {
         case other
     }
 
-    /// The object is a reference to a PBXProject element.
+    /// The object is a reference to a PBXProject element if proxy is for the object located in current .xcodeproj, otherwise PBXFileReference.
     var containerPortalReference: PBXObjectReference
 
-    /// Returns the project that contains the remote object.
+    /// Returns the project that contains the remote object. If container portal is a remote project this getter will fail. Use isContainerPortalFileReference to check if you can use the getter
```
pepibumur: Considering that we are in the middle of a major release, I'd take the opportunity to make the API nicer. What about something like:

```swift
enum ContainerPortal {
  case project(PBXProject)
  case fileReference(PBXFileReference)
}

// PBXContainerItemProxy
class PBXContainerItemProxy {
  public var containerPortal: ContainerPortal
}
```

What do you think @damirdavletov ?

Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
```diff
@@ -37,25 +51,64 @@ public final class PBXContainerItemProxy: PBXObject {
         }
     }
 
+    /// The value of remoteGlobalIDString property in pbxproj. Use this accessor if the object is in remote .xcodeproj.
+    public var remoteGlobalIDString: String? {
+        get {
+            return remoteGlobalIDReference?.value
+        }
+        set {
+            remoteGlobalIDReference = newValue.map { PBXObjectReference($0) }
```
pepibumur: I might be wrong because it's been a long time without playing with this project but I'd say the `remoteGlobalIDReference` is not necessary. We can just have `removeGlobalID` of type String.

Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
damirdavletov: @pepibumur I agree regarding making the API nicer. My objective was not introduce breaking changes, but if that's not problem I'll try to make it nicer. Thanks for the feedback!

pepibumur: Yeah, we avoid introducing breaking changes but this is a special case because we are working on a major release. Feel free to leverage Swift features to make the API nice :)

damirdavletov:
@pepibumur Thanks for the improvement suggestions! Please have a look at the amended implementation as per your comments.

Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
```diff
@@ -1,5 +1,11 @@
 import Foundation
 
+public enum ContainerPortal {
```
pepibumur: I'd move this one into `PBXContainerItemProxy`

pepibumur:
Only one more thing @damirdavletov, could you update the CHANGELOG? Make sure that entry that you add states that the change is a breaking change.

Sources/xcodeproj/Objects/Files/PBXContainerItemProxy.swift
damirdavletov: @pepibumur this is done

damirdavletov:
> Only one more thing @damirdavletov, could you update the CHANGELOG? Make sure that entry that you add states that the change is a breaking change.

This was already there. Let me know if you need anything else done. Thanks!

yonaskolb:
I think this may have broken things. Code outside this project can't get access to the `uuid` of an object, so there isn't a way to access a value to pass into `remoteGlobalIDString`

yonaskolb:
In addition as the `uuid` may be generated after everything is created, it's not possible to access it anyway. This may need to become an enum instead of a string
```swift
enum RemoteGlobalID {
	case reference(PBXReference)
	case uuid(String(
}
```

pepibumur:
Would you mind opening an issue with that @yonaskolb?

PR #428, opened 2019-05-21
Throw an error from fullPath if the path can't be resolved

damirdavletov:
### Short description 📝
ATM the following method:
```
public extension PBXFileElement {
    func fullPath(sourceRoot: Path) throws -> Path? 
}
```
returns sourceRoot value if the file element belongs to a group which doesn't have mainGroup in parent groups. 
In our setup we read an .xcodeproj then take all referenced .xcodeproj files recursively. Often when conflicts are resolved incorrectly there's duplicate entries of file elements with one being ignored by xcode as it's not part of group hierarchy.
 When we try get fullPath of such file element it returns sourceRoot and there is no way to distinguish if it is an error or correct path.

### Solution 📦
Throw and error in case fullPath can't be resolved correctly and let user handle it.

### Implementation 👩‍💻👨‍💻
Please have a look at the pull request. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - Added `com.apple.product-type.framework.static` to `PBXProductType`. https://github.com/tuist/xcodeproj/pull/347 by @ileitch.
 - Can add a not existing file to a group https://github.com/tuist/xcodeproj/pull/418 by @llinardos.
 - **Breaking** Add `SWIFT_COMPILATION_MODE` and `CODE_SIGN_IDENTITY` build settings, remove `DEBUG` flag for Release https://github.com/tuist/xcodeproj/pull/417 @dangthaison91 
-- Added throwing an error in case group path can't be resolved by @damirdavletov  
+- **Breaking** Added throwing an error in case group path can't be resolved by @damirdavletov  
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - Added `com.apple.product-type.framework.static` to `PBXProductType`. https://github.com/tuist/xcodeproj/pull/347 by @ileitch.
 - Can add a not existing file to a group https://github.com/tuist/xcodeproj/pull/418 by @llinardos.
 - **Breaking** Add `SWIFT_COMPILATION_MODE` and `CODE_SIGN_IDENTITY` build settings, remove `DEBUG` flag for Release https://github.com/tuist/xcodeproj/pull/417 @dangthaison91 
+- Added throwing an error in case group path can't be resolved by @damirdavletov
```
pepibumur: This is a breaking change so I'd annotate this line with `**Breaking**`

Tests/xcodeprojTests/Objects/Files/PBXFileElementTests.swift
```diff
@@ -73,8 +82,7 @@ final class PBXFileElementTests: XCTestCase {
         // Remove parent for fallback test
         fileref.parent = nil
 
-        let fullPath = try fileref.fullPath(sourceRoot: sourceRoot)
-        XCTAssertEqual(fullPath?.string, "/a/path")
+        XCTAssertThrowsError(try fileref.fullPath(sourceRoot: sourceRoot))
```
pepibumur: I'd assert that the thrown error is the expected one.

damirdavletov:
> Only a few minor things. Good job @damirdavletov

Fixed all of them. Thanks!

Issue #438, opened 2019-06-03
Add support for Xcode 11 / SwiftPM


ollieatkinson:
## Context 🕵️‍♀️

Xcode 11 releases swift packages integrated inside of the xcodeproj. This introduces some new types. 

I am creating this thread so that we may start collating together all the information needed to support these new types.

----

## Adding a dependency on a package from Xcode:

```
/* Begin PBXBuildFile section */
426666BC22A5DC4300E840E0 /* XcodeProj in Frameworks */ = {isa = PBXBuildFile; productRef = 426666BB22A5DC4300E840E0 /* XcodeProj */; };
/* End PBXBuildFile section */
```

```
/* Begin PBXFrameworksBuildPhase section */
		426666AD22A5DA9E00E840E0 /* Frameworks */ = {
			isa = PBXFrameworksBuildPhase;
			buildActionMask = 2147483647;
			files = (
				426666BC22A5DC4300E840E0 /* XcodeProj in Frameworks */,
			);
			runOnlyForDeploymentPostprocessing = 0;
		};
/* End PBXFrameworksBuildPhase section */
```

```
/* Begin PBXNativeTarget section */
		426666AF22A5DA9E00E840E0 /* Reader */ = {
			isa = PBXNativeTarget;
/* ... */
                      packageProductDependencies = (
                           426666BB22A5DC4300E840E0 /* XcodeProj */,
                       );
/* ... */
		};
/* End PBXNativeTarget section */
```

```
/* Begin PBXProject section */
		426666A822A5DA9E00E840E0 /* Project object */ = {
			isa = PBXProject;
/* ... */ 
                       packageReferences = (
                            426666BA22A5DC4300E840E0 /* XCRemoteSwiftPackageReference "xcodeproj" */,
                        );
/* ... */
		};
/* End PBXProject section */
```

```
/* Begin XCRemoteSwiftPackageReference section */
		426666BA22A5DC4300E840E0 /* XCRemoteSwiftPackageReference "xcodeproj" */ = {
			isa = XCRemoteSwiftPackageReference;
			repositoryURL = "https://github.com/tuist/xcodeproj.git";
			requirement = {
				branch = master;
				kind = branch;
			};
		};
/* End XCRemoteSwiftPackageReference section */

/* Begin XCSwiftPackageProductDependency section */
		426666BB22A5DC4300E840E0 /* XcodeProj */ = {
			isa = XCSwiftPackageProductDependency;
[...]

PR #439, opened 2019-06-04
Add support for the Swift Package Manager

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/438

### Short description 📝
With the introduction of Swift Package Manager support to Xcode projects, Apple has added new objects and attributes to existing models.

### Solution 📦
Add support for those changes from XcodeProj.

### Implementation 👩‍💻👨‍💻
- [x] Add `PBXProject.packages` attribute.
- [x] Add `PBXNativeTarget. packageProductDependencies` attribute.
- [x] Define `XCRemoteSwiftPackageReference` model.
- [x] Define `XCSwiftPackageProductDependency` model.
- [x] Add `PBXBuildFile.productRef` attribute.
- [x] Verify the optionality of the attributes.
- [x] Update the `ReferenceGenerator` logic to generate the reference for the new objects.
- [x] Add tests.
- [x] Add method to `PBXProject` to add a new Swift package to the project.

### DIFF
--- a/Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
+++ b/Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
@@ -18,7 +18,17 @@ public final class PBXBuildFile: PBXObject {
     }
     
     /// Product reference.
-    public var productReference: PBXObjectReference?
+    var productReference: PBXObjectReference?
+    
+    /// Product.
+    public var product: XCSwiftPackageProductDependency? {
+        get {
+            return productReference?.getObject()
+        }
+        set {
+            self.productReference = newValue?.reference
+        }
+    }
 
     /// Element settings
     public var settings: [String: Any]?
@@ -45,6 +55,7 @@ public final class PBXBuildFile: PBXObject {
     enum CodingKeys: String, CodingKey {
         case fileRef
         case settings
+        case productRef
     }
 
     public required init(from decoder: Decoder) throws {
@@ -53,6 +64,9 @@ public final class PBXBuildFile: PBXObject {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         if let fileRefString: String = try container.decodeIfPresent(.fileRef) {
             fileReference = objectReferenceRepository.getOrCreate(reference: fileRefString, objects: objects)
+        }
+        if let productRefString: String = try container.decodeIfPresent(.productRef) {
+            productReference = objectReferenceRepository.getOrCreate(reference: productRefString, objects: objects)
         }
         settings = try container.decodeIfPresent([String: Any].self, forKey: .settings)
         try super.init(from: decoder)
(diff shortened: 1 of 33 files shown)


kwridan:
👍 nice work!

Mind if I push a fixture that includes local and remote package dependencies? From what I gathered `Project.xcodeproj` has been configured with all use cases tested via `PBXProjIntegrationTests`, I can update it to include some packages too - or is it best I create a stand alone project for this - e.g. `ProjectWithSwiftPackageDependencies.xcodeproj`?

kwridan:
I was curious what happens when you open a project with Swift packages in Xcode 10.2.1 

Screen Shot 2019-06-06 at 7 15 26 PM

Looks like the `objectVersion` changes to `52` when adding packages to protect against opening the project with an older version of Xcode. 

```diff
 	classes = {
 	};
-	objectVersion = 50;
+	objectVersion = 52;
 	objects = {
```

Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
```diff
@@ -17,6 +17,19 @@ public final class PBXBuildFile: PBXObject {
         }
     }
 
+    /// Product reference.
+    var productReference: PBXObjectReference?
+
+    /// Product.
+    public var product: XCSwiftPackageProductDependency? {
```
yonaskolb: Would this just be a plain `PBXObject` or `PBXContainerItem`?

ollieatkinson: Looks like the `XCSwiftPackageProductDependency` is a reference for a `PBXTargetDependency`. As far as I am aware they introduced a new type `XCRemoteSwiftPackageReference` which contains things like the package url and versioning information. Checkout the issue where there's more information dumped: https://github.com/tuist/xcodeproj/issues/438

pepibumur:
> Mind if I push a fixture that includes local and remote package dependencies? From what I gathered Project.xcodeproj has been configured with all use cases tested via PBXProjIntegrationTests, I can update it to include some packages too - or is it best I create a stand alone project for this - e.g. ProjectWithSwiftPackageDependencies.xcodeproj?

Sure, go for it @kwridan

pepibumur:
> Looks like the objectVersion changes to 52 when adding packages to protect against opening the project with an older version of Xcode.

I've updated `Xcode.Default.objectVersion` with that value.

Sources/xcodeproj/Objects/BuildPhase/PBXBuildFile.swift
pepibumur: > Would this just be a plain PBXObject or PBXContainerItem?

Not sure @yonaskolb . Since the attribute has been introduced to refer to package product dependencies, I'd keep the type and make it more generic in the future if we see that Apple is using it for something else.

kwridan:
@pepibumur I've pushed an update to the existing fixture to a separate branch as it seems the `XcodeprojPropertyListDecoder` is failing to read the project but wanted to second sanity check incase I was doing something silly. 

 https://github.com/tuist/xcodeproj/tree/wip/spm-support-fixture

pepibumur:
I fixed the tests and pushed the changes in your branch to this branch.

PR #444, opened 2019-06-07
Swift Package fixes

yonaskolb:
- makes some properties in the new swift package classes public and in the inits so they can be used in other tools
- adds the `.swiftpm` directory to `.gitignore` for when opening the `Package.swift` in Xcode 11
- make `PBXProject.packages` and `PBXTarget.packageProductDependencies`
  - public
  - non optional
  - available in init

### DIFF
--- a/Sources/xcodeproj/Objects/Project/PBXProject.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXProject.swift
@@ -125,15 +125,15 @@ public final class PBXProject: PBXObject {
     }
 
     /// Package references.
-    var packageReferences: [PBXObjectReference]?
+    var packageReferences: [PBXObjectReference]
 
     /// Swift packages.
-    var packages: [XCRemoteSwiftPackageReference]? {
+    public var packages: [XCRemoteSwiftPackageReference] {
         set {
-            packageReferences = newValue?.map { $0.reference }
+            packageReferences = newValue.references()
         }
         get {
-            return packageReferences?.objects()
+            return packageReferences.objects()
         }
     }
 
@@ -175,8 +175,7 @@ public final class PBXProject: PBXObject {
         // Reference
         let reference = XCRemoteSwiftPackageReference(repositoryURL: repositoryURL, versionRules: versionRules)
         objects.add(object: reference)
-        if packages == nil { packages = [] }
-        packages?.append(reference)
+        packages.append(reference)
 
         // Product
         let productDependency = XCSwiftPackageProductDependency(productName: productName, package: reference)
@@ -223,6 +222,7 @@ public final class PBXProject: PBXObject {
                 projects: [[String: PBXFileElement]] = [],
                 projectRoots: [String] = [],
                 targets: [PBXTarget] = [],
+                packages: [XCRemoteSwiftPackageReference] = [],
                 attributes: [String: Any] = [:],
                 targetAttributes: [PBXTarget: [String: Any]] = [:]) {
         self.name = name
@@ -237,6 +237,7 @@ public final class PBXProject: PBXObject {
         projectReferences = projects.map { project in project.mapValues { $0.reference } }
         self.projectRoots = projectRoots
         targetReferences = targets.references()
+        packageReferences = packages.references()
         self.attributes = attributes
         targetAttributeReferences = [:]
         super.init()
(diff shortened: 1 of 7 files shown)


yonaskolb:
I don't like that Danger requires a changelog. Many things don't require an entry, and some PR's like this aren't new features they are just fixing what is in a previous one that has not been released yet

yonaskolb:
@pepibumur was there a reason that `XCRemoteSwiftPackageReference.VersionRules` isn't called `XCRemoteSwiftPackageReference.Requirement` like the actual xcodeproj property name?

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -125,15 +125,15 @@ public final class PBXProject: PBXObject {
     }
 
     /// Package references.
-    var packageReferences: [PBXObjectReference]?
```
pepibumur: What we've been doing for deciding whether a property should be optional or not is whether Xcode is able to open a project without that attribute. If I'm not wrong, Xcode is able to open a project that doesn't have this attribute. If that's the case, I'd keep this property as optional. It also ensures that we are not adding attributes to Xcode 10 projects that are are not readable by Xcode 10.

Sources/xcodeproj/Objects/SwiftPackage/XCRemoteSwiftPackageReference.swift
```diff
@@ -111,17 +111,17 @@ public class XCRemoteSwiftPackageReference: PBXContainerItem, PlistSerializable
     }
 
     /// Repository url.
-    var repositoryURL: String?
+    public var repositoryURL: String?
```
pepibumur: Ouch! Good catch

Sources/xcodeproj/Objects/Project/PBXProject.swift
yonaskolb: I just changed it to an empty array if there is nothing to decode, and only encode if there's something in it. I'm not sure I understand you, does this mess up a specific use case?

ollieatkinson: I'd say this change makes sense - optional arrays are often troubling when it comes to accessing their contents. I can't see a situation where this will cause a problem.

If you want to check the presence of a package inside the Xcodeproj it's possible to by calling `packageReferences.count > 0`.

ollieatkinson:
> I don't like that Danger requires a changelog. Many things don't require an entry, and some PR's like this aren't new features they are just fixing what is in a previous one that has not been released yet

As far as I'm aware the Danger rule is a warning for the CHANGELOG file - not a requirement for merge. Is that not the case?

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: Let's take the following scenario:
- The user edits a Xcode project that doesn't have the `packageReferences` attribute.
- We decode the attribute to an empty array.
- After saving the project, the user will see in the diff that a new attribute is added.

Xcode 10.x will ignore the attribute, but we are introducing unexpected modifications to the project. That's something we've always avoided, and I don't think it's solvable by not writing the array when it's empty, because that's another possible value for the attribute.

If I'm not wrong, I think Xcode 11 updates some attribute in the project to say that the project is compatible with version X of Xcode. What about using that value to decide whether the attribute should be written or not?

pepibumur:
That's why Danger is a warning and not erroring. We can't determine whether changes need documentation. The developer opening the PR is responsible for making the call here.

Sources/xcodeproj/Objects/Project/PBXProject.swift
yonaskolb: @pepibumur What do you mean by attribute? The packageReferences? If the array is empty we don't write it to the project, exactly as if it was nil

pepibumur:
> @pepibumur was there a reason that XCRemoteSwiftPackageReference.VersionRules isn't called XCRemoteSwiftPackageReference.Requirement like the actual encoded property name?
Having a plural as a type name is also a bit strange

I used the name that Xcode uses but you are right, it makes more sense to use the name of the property in the project. I'll update it.

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: Yes, let me put it on a table:

| Project in disk             | Read project                | Diff? |
|-----------------------------|-----------------------------|-------|
|  No packageReferences       | Empty packageReferences     | Yes   |
| Empty packageReferences     | Empty packageReferences     | No    |
| Non-empty packageReferences | Non-empty packageReferences | No    |

By turning nil into an empty array we'll get a diff after writing the project to disk in the first scenario.

ollieatkinson: I assume'd we wouldn't write `Empty packageReferences`. Is there a benefit to doing so? If we assume empty means we don't write it then we don't have to worry about the truth table above since we only handle two scenarios, empty and non-empty

pepibumur: There isn't any difference for Xcode. The point is that users don't want to see git diffs if the project hasn't been modified, and with the current implementation they will. I suggest the following modification:

```swift
// This one remains optional
var packageReferences: [PBXObjectReference]?

// We define this one as non-optional
public var packages: [XCRemoteSwiftPackageReference] {
  set {
    packageReferences = newValue.references()
  }
  get {
     return packageReferences.objects() ?? []
  }
}
```

Does that look good @ollieatkinson / @yonaskolb ?

ollieatkinson: That makes sense - 👍

PR #445, opened 2019-06-10
Fix PBXContainerItemProxy.remoteGlobalID

yonaskolb:
### Short description 📝
https://github.com/tuist/xcodeproj/pull/425 changed `PBXContainerItemProxy.remoteGlobalID` into a string to support references outside of the project. This created a problem because if setting the id from a PBXObject.uuid, this would changed upon reference resolution. Outsider users of xcodeproj also don't have access to the PBXReference

### Solution 📦
Create an enum that allows for setting an `.object` as before or a `.string` for remote projects


pepibumur:
I updated the CHANGELOG [here](https://github.com/tuist/xcodeproj/commit/bc8e2f5a80564e966534f71cc28a5e688c46cf46)

PR #458, opened 2019-07-19
Adding additional `lastKnownFileType`s

kwridan:
Part of: https://github.com/tuist/tuist/issues/425

### Short description 📝

- Generating projects that include `.c`, `.cpp`, `.mm` and `.hpp` files using Tuist had the `lastKnownFileType` missing for those files
- This causes Xcode to update the generated project to include those attributes
- In a workflow where the generated projects are checked-in, it can result in differences appearing even though no changes were made

### Solution 📦
- Update the list of known file types to include those files


pepibumur:
Just a CHANGELOG entry and we can merge :)

PR #462, opened 2019-07-28
Add variant group

timbaev:

### DIFF
--- a/Sources/xcodeproj/Objects/Files/PBXGroup.swift
+++ b/Sources/xcodeproj/Objects/Files/PBXGroup.swift
@@ -140,6 +140,25 @@ public extension PBXGroup {
         }
     }
 
+    /// Creates a variant group with the given name and returns it.
+    ///
+    /// - Parameters:
+    ///   - groupName: group name.
+    /// - Returns: created groups.
+    @discardableResult
+    func addVariantGroup(named groupName: String) throws -> [PBXVariantGroup] {
+        let objects = try self.objects()
+
+        return groupName.components(separatedBy: "/").reduce(into: [PBXVariantGroup]()) { groups, name in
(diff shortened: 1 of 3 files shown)


Sources/xcodeproj/Objects/Files/PBXGroup.swift
```diff
@@ -140,6 +140,25 @@ public extension PBXGroup {
         }
     }
 
+    /// Creates a variant group with the given name and returns it.
+    ///
+    /// - Parameters:
+    ///   - groupName: group name.
+    /// - Returns: created groups.
+    @discardableResult
+    func addVariantGroup(named groupName: String) throws -> [PBXVariantGroup] {
```
pepibumur: Would you mind adding a unit test that tests the logic in this method?

Tests/xcodeprojTests/Objects/Files/PBXGroupTests.swift
```diff
@@ -52,6 +52,26 @@ final class PBXGroupTests: XCTestCase {
...
+
+        let group = PBXGroup(children: [],
+                             sourceTree: .group,
+                             name: "group")
+
+        project.add(object: group)
+
+        let childVariantGroup = try? group.addVariantGroup(named: "child_variant_group").first
+
+        XCTAssertNotNil(childVariantGroup?.parent)
```
pepibumur: With this only assertion were are not testing most of the logic that is in the method.
- Are all the components being added as children?
- Are they created with the right name and source tree?
- Are all the expected variant groups returned?

pepibumur:
Tests are [failing on CI](https://circleci.com/gh/tuist/XcodeProj/507). Once fixed, we are good to merge the PR.

PR #463, opened 2019-07-30
Update BuildSettingProvider to return LD_RUNPATH_SEARCH_PATHS as Array<String>

marciniwanicki:
Part of https://github.com/tuist/tuist/pull/464.

### Short description 📝
There is a small inconsistent in the representation of multi-value settings returned from `BuildSettingsProvider`. 
`GCC_PREPROCESSOR_DEFINITIONS` is defined as an array of strings, whereas `LD_RUNPATH_SEARCH_PATHS` is a single string.

The problem appears once the project is serialized by Xcode (tested with Xcode 10.2.1). `LD_RUNPATH_SEARCH_PATHS` are reformatted to arrays which can result in an unexpected diff.

### Solution 📦
Use `Array<String>` to represent `LD_RUNPATH_SEARCH_PATHS` setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## next version
 
+### Changed
+
+- Update `BuildSettingProvider` to return `LD_RUNPATH_SEARCH_PATHS` as `Array<String>` https://github.com/tuist/xcodeproj/pull/463 by @marciniwanicki
+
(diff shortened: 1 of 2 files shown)


pepibumur:
Could you add a note to the CHANGELOG @marciniwanicki? We can merge afterwards.

PR #468, opened 2019-09-08
Local package

fortmarek:
Adding support for adding local Swift package needed for `tuist` ad https://github.com/tuist/tuist/pull/394

### Short description 📝
Adding public func for adding local swift package

### Solution 📦
Taking already implemented `addSwiftPackage` and adjusting it for local. Parameter `addFileReference` adds the posibility to choose where to add `PBXFileReference` by the user. 

### Implementation 👩‍💻👨‍💻

- [x] Adding public-facing method
- [x] Adding tests

### Questions

I have not been able to find any tests for `addSwiftPackage`, so I am not sure how to test this, if at all.

### DIFF
--- a/Sources/xcodeproj/Errors/Errors.swift
+++ b/Sources/xcodeproj/Errors/Errors.swift
@@ -173,9 +173,9 @@ enum PBXProjError: Error, CustomStringConvertible {
         case let .targetNotFound(targetName):
             return "Could not find target with \(targetName)"
         case let .frameworksBuildPhaseNotFound(targetName):
-            return "Could not find frameworks build phase for \(targetName) target."
+            return "Could not find frameworks build phase for target \(targetName)."
         case let .sourcesBuildPhaseNotFound(targetName):
-            return "Could not find sources build phase for \(targetName) target."
+            return "Could not find sources build phase for target \(targetName)."
         }
     }
 }
--- a/Sources/xcodeproj/Errors/Errors.swift
+++ b/Sources/xcodeproj/Errors/Errors.swift
@@ -163,6 +163,7 @@ enum PBXProjError: Error, CustomStringConvertible {
     case invalidGroupPath(sourceRoot: Path, elementPath: String?)
     case targetNotFound(targetName: String)
     case frameworksBuildPhaseNotFound(targetName: String)
+    case sourcesBuildPhaseNotFound(targetName: String)
     var description: String {
         switch self {
         case let .notFound(path):
@@ -173,6 +174,8 @@ enum PBXProjError: Error, CustomStringConvertible {
             return "Could not find target with \(targetName)"
         case let .frameworksBuildPhaseNotFound(targetName):
             return "Could not find frameworks build phase for \(targetName) target."
+        case let .sourcesBuildPhaseNotFound(targetName):
+            return "Could not find sources build phase for \(targetName) target."
         }
     }
 }
--- a/Sources/xcodeproj/Errors/Errors.swift
+++ b/Sources/xcodeproj/Errors/Errors.swift
@@ -164,6 +164,7 @@ enum PBXProjError: Error, CustomStringConvertible {
     case targetNotFound(targetName: String)
     case frameworksBuildPhaseNotFound(targetName: String)
     case sourcesBuildPhaseNotFound(targetName: String)
+    case pathIsAbsolute(Path)
     var description: String {
         switch self {
         case let .notFound(path):
@@ -173,9 +174,11 @@ enum PBXProjError: Error, CustomStringConvertible {
         case let .targetNotFound(targetName):
             return "Could not find target with \(targetName)"
         case let .frameworksBuildPhaseNotFound(targetName):
-            return "Could not find frameworks build phase for target \(targetName)."
+            return "Could not find frameworks build phase for target \(targetName)"
         case let .sourcesBuildPhaseNotFound(targetName):
-            return "Could not find sources build phase for target \(targetName)."
+            return "Could not find sources build phase for target \(targetName)"
+        case let .pathIsAbsolute(path):
+            return "Path must be relative, but path \(path.string) is absolute"
         }
     }
 }
--- a/Sources/xcodeproj/Errors/Errors.swift
+++ b/Sources/xcodeproj/Errors/Errors.swift
@@ -161,12 +161,18 @@ enum PBXProjEncoderError: Error, CustomStringConvertible {
 enum PBXProjError: Error, CustomStringConvertible {
     case notFound(path: Path)
     case invalidGroupPath(sourceRoot: Path, elementPath: String?)
+    case targetNotFound(targetName: String)
+    case frameworksBuildPhaseNotFound(targetName: String)
     var description: String {
         switch self {
         case let .notFound(path):
             return ".pbxproj not found at path \(path.string)"
         case let .invalidGroupPath(sourceRoot, elementPath):
             return "Cannot calculate full path for file element \"\(elementPath ?? "")\" in source root: \"\(sourceRoot)\""
+        case let .targetNotFound(targetName):
+            return "Could not find target with \(targetName)"
+        case let .frameworksBuildPhaseNotFound(targetName):
+            return "Could not find frameworks build phase for \(targetName) target."
         }
     }
 }
--- a/Sources/xcodeproj/Errors/Errors.swift
+++ b/Sources/xcodeproj/Errors/Errors.swift
@@ -158,7 +158,7 @@ enum PBXProjEncoderError: Error, CustomStringConvertible {
 /// PBXProj error.
 ///
 /// - notFound: the .pbxproj cannot be found at the given path.
-enum PBXProjError: Error, CustomStringConvertible {
+enum PBXProjError: Error, CustomStringConvertible, Equatable {
     case notFound(path: Path)
     case invalidGroupPath(sourceRoot: Path, elementPath: String?)
     case targetNotFound(targetName: String)
--- a/Sources/xcodeproj/Objects/Project/PBXProject.swift
+++ b/Sources/xcodeproj/Objects/Project/PBXProject.swift
@@ -203,7 +203,6 @@ public final class PBXProject: PBXObject {
         return reference
     }
     
-    
     /// Adds a local swift package
     ///
     /// - Parameters:
@@ -231,7 +230,9 @@ public final class PBXProject: PBXObject {
         objects.add(object: buildFile)
         
         // Link the product
-        guard let frameworksBuildPhase = try target.frameworksBuildPhase() else { throw PBXProjError.frameworksBuildPhaseNotFound(targetName: targetName) }
+        guard let frameworksBuildPhase = try target.frameworksBuildPhase() else {
+            throw PBXProjError.frameworksBuildPhaseNotFound(targetName: targetName)
+        }
         
         frameworksBuildPhase.files?.append(buildFile)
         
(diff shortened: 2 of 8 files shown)


Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
...
     }
+    
+    
+    /// Adds a local swift package
+    ///
+    /// - Parameter path: Relative path to the swift package
+    /// - Parameter productName: The product to depend on
+    /// - Parameter target: Target to link package product to
+    /// - Parameter addFileReference: Include a file reference to the package (defaults to main group)
+    public func addLocalSwiftPackage(path: String,
```
pepibumur: Nitpick: I'd turn the `path` type into `Path`.

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
...
+                                     productName: String,
+                                     target: PBXTarget? = nil,
+                                     addFileReference: Bool = true) -> XCSwiftPackageProductDependency {
+        let objects = try! self.objects()
```
pepibumur: I'd add a `throws` to the method signature and bubble this up instead of resorting to `try!`.

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
...
+    
+    /// Adds a local swift package
+    ///
+    /// - Parameter path: Relative path to the swift package
+    /// - Parameter productName: The product to depend on
+    /// - Parameter target: Target to link package product to
+    /// - Parameter addFileReference: Include a file reference to the package (defaults to main group)
+    public func addLocalSwiftPackage(path: String,
+                                     productName: String,
+                                     target: PBXTarget? = nil,
```
pepibumur: I think it's safe here to assume that the developer wants to add the package to an existing target, and therefore I'd make this a non-optional attribute. If developers are only interested in adding this to the project, they can implement it themselves.

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
...
+        let productDependency = XCSwiftPackageProductDependency(productName: productName)
+        objects.add(object: productDependency)
+        target?.packageProductDependencies.append(productDependency)
+
+        // Build file
+        let buildFile = PBXBuildFile(product: productDependency)
+        objects.add(object: buildFile)
+        
+        // Link the product
+        try? target?.frameworksBuildPhase()?.files?.append(buildFile)
```
pepibumur: I'd not swallow the errors thrown here.

pepibumur: Moreover, if a frameworks build phase doesn't exist, this method doesn't do what's supposed to do. I'd rather throw an error and tell the user what's going on.

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
...
+        
+        // File reference
+        if addFileReference {
+            let reference = PBXFileReference(sourceTree: .group,
+                                             name: productName,
+                                             lastKnownFileType: "folder",
```
pepibumur: Is this necessary?

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
...
+
+        // Build file
+        let buildFile = PBXBuildFile(product: productDependency)
+        objects.add(object: buildFile)
+        
+        // Link the product
+        try? target?.frameworksBuildPhase()?.files?.append(buildFile)
+        
+        // File reference
+        if addFileReference {
```
pepibumur: Why would you add a package without adding its reference? Similar to my comment about the optionality of the target argument, I think it's safe to default to true here and remove this argument from the method signature.

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: Considering you can do:

```swift
self.targets.first(where: { $0.name == targetName})
```
I'd turn the target argument into `targetName: String`.

pepibumur:
> I have not been able to find any tests for addSwiftPackage, so I am not sure how to test this, if at all.

I'd add a unit test that calls the method on a project with a `PBXNativeTarget` and validates that the project has been modified accordingly to add the Xcode-specific elements:
- Did we add the build file?
- Was the build file added to the `target. packageProductDependencies`?
- Did we add the file reference to the project?

Sources/xcodeproj/Objects/Project/PBXProject.swift
```diff
@@ -191,6 +191,44 @@ public final class PBXProject: PBXObject {
 
         return reference
     }
+    
+    
+    /// Adds a local swift package
+    ///
+    /// - Parameter path: Relative path to the swift package
+    /// - Parameter productName: The product to depend on
```
pepibumur: Does it require the extension to be in the name/

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: Should not be that the same for `addSwiftPackage` then?

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: Good point 👍

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: OK, similarly should not be that the same for `addSwiftPackage`? Otherwise, agree, just tried to be consistent 🙂

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: Do you mean `addFileReference` boolean? If so, that's something we (along with @kwridan ) wanted to take care of in Tuist - local packages' references are added to the top of the children in the main group, so in `tuist` we wanted to achieve consistency with that and that was not possible without adding the reference ourselves before calling this method.

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: Not sure I follow, could you elaborate?

pepibumur: If the product name is for example `Alamofire.framework`, do I have to pass `Alamofire` or the name with the extension? I'd include that in the line that documents the argument:
```swift
/// - Parameter productName: The product to depend on with/without the extension
```

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: I see what you mean. What about adding a comment in the line `220` explaining this scenario? I think it's not obvious why someone might want to add a Swift package without adding the reference.

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: We should be consistent yeah, but thinking about it again, I think it makes sense to bubble them up.

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: It's just that it should be relative, so maybe make assertion for `path.isRelative = true` and add it to the documentation? Or would it be better to accept absolute path, too and derive the relative from the project's root folder?

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: Yeah, it should be the same. I disregarded that when I implemented `addSwiftPackage`. My bad 😬

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: That's a good point, it should be relative. If we get the absolute path, we can't turn it into a relative because the project's absolute path is only known when the project is written. I suggest two things:
- Turn it into a `Path` type and throw if the path is not relative.
- Mention this in the line that documents the argument.

What do you think?

fortmarek: Like that increased safety, will do 👍

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: Now, is not that the same for `addSwiftPackage`? If so, probably would make sense to change it to that, too.

Sources/xcodeproj/Objects/Project/PBXProject.swift
fortmarek: Without the extension, added it 👍

Sources/xcodeproj/Objects/Project/PBXProject.swift
pepibumur: Yes, I'd change it too

Tests/xcodeprojTests/Objects/Project/PBXProjectTests.swift
```diff
@@ -30,4 +30,121 @@ final class PBXProjectTests: XCTestCase {
...
+                                             productName: "Product",
+                                             targetName: target.name)
+        }
+        // Then
+        catch let error {
+            XCTAssertEqual(PBXProjError.frameworksBuildPhaseNotFound(targetName: target.name).localizedDescription, error.localizedDescription)
+        }
+    }
+    
+    func test_addSwiftPackage() throws {
```
pepibumur: Thanks for adding a test for this one.

Tests/xcodeprojTests/Objects/Project/PBXProjectTests.swift
```diff
@@ -30,4 +30,121 @@ final class PBXProjectTests: XCTestCase {
...
+                                 buildConfigurationList: configurationList,
+                                 compatibilityVersion: "0",
+                                 mainGroup: mainGroup,
+                                 targets: [target])
+        
+        objects.add(object: project)
+        
+        // When
+        do {
+            _ = try project.addLocalSwiftPackage(path: "Product",
```
pepibumur: Nitpick: You can use `XCTAssertThrowsError` here

pepibumur:
```swift
XCTAssertThrowsError(try methodCall()) {
  XCTAssertEqual(error as? PBXProjError, PBXProjError.frameworksBuildPhaseNotFound(targetName: target.name))
}
```
Notice that `PBXProjError` needs to conform the `Equatable` protocol.

Sources/xcodeproj/Errors/Errors.swift
```diff
@@ -161,12 +161,24 @@ enum PBXProjEncoderError: Error, CustomStringConvertible {
 enum PBXProjError: Error, CustomStringConvertible {
```
pepibumur: I'd conform `Equatable` here so that we can compare the errors from the tests.

fortmarek: Good point, done 👍

PR #473, opened 2019-09-15
Add onlyGenerateCoverageForSpecifiedTargets parameter to TestAction

kateinoigakukun:
### Short description 📝

This change allows generation of schemes where TestAction has `onlyGenerateCoverageForSpecifiedTargets` enabled. This parameter is necessary to gather code coverage for specific targets.

### Solution 📦
Add `onlyGenerateCoverageForSpecifiedTargets` to TestAction properties.

### Implementation 👩‍💻👨‍💻

- [x]  Added property to `onlyGenerateCoverageForSpecifiedTargets` to TestAction
- [x] Added test case for this


pepibumur:
Thanks for bringing this to the project @kateinoigakukun

PR #477, opened 2019-09-17
Fix the reference length

pepibumur:
### Short description 📝
[This commit](https://github.com/tuist/XcodeProj/commit/3b909ba75da84e92c9b747540f8b9f4f8715ae09) introduced a regression that caused the tests to fail because the generated object reference is 25 characters long instead of 24.

This PR fixes it.


giginet:
This issue caused by #474 😞 

We should use ClosedRange instead of changing the test case.

```swift
String(reference[..<reference.index(reference.startIndex, offsetBy: 24)])
```

pepibumur:
> We should use ClosedRange instead of changing the test case.

Feel free to open a PR with that suggestion 🤗

PR #488, opened 2019-10-10
System library added to a group has empty path

damirdavletov:
### Short description 📝
Adding system library to Frameworks group using PBXGroup.addFile() method causes BPFFileReference to have empty path

### Solution 📦
Edit PBXGroup.addFile method to correctly create a file with path that was passed with sourceTree type sdkRoot and developerDir


damirdavletov:
Thanks for the quick review @pepibumur ! I've updated the changelog as well.

pepibumur:
Thank you for fixing it! We can merge it as soon as CI is happy.

PR #489, opened 2019-10-10
Fix Products group serialisation with temporary ids

damirdavletov:
### Short description 📝
When adding remote xcodeproj bundle reference to the project, remote product references are serialised with temporary ids.

### Solution 📦
Fix serialisation of **Products** group of the remote project references


damirdavletov:
Thanks @pepibumur !

PR #496, opened 2019-10-29
Test Plan Consistent Serialisation

adamkhazi:
### Short description 📝
Test plans were recently made optional (c15034948a2a132bf559f14d3c6b4d1b73749663). When reading from a file, test plans were wrongly defaulting to an empty array even when a `nil` test plan was set on the `TestAction` that was written to file originally.

### Solution 📦
Remove `?? []`and leave `testPlans` as `nil`.

### Implementation 👩‍💻👨‍💻
- [X] Remove default test plan behaviour
- [X] Write a test for this case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## Next
 
+### Fixed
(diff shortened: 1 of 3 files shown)


pepibumur:
I'm releasing a new patch release with this. Thanks for looking into it @adamkhazi

Issue #498, opened 2019-11-05
XcodeProj's dependency on swift-package-manager breaks dependency resolution for consumers of XcodeProj


elliottwilliams:
## Context 🕵️‍♀️
In #492, @pepibumur added SwiftPM as a dependency of XcodeProj. In Package.resolved:

```swift
.package(url: "https://github.com/apple/swift-package-manager", .branch("swift-5.0-RELEASE")),
```

## What 🌱
Unfortunately, this breaks downstream Swift packages that depend on XcodeProj. There's a restriction in SwiftPM that [version-based packages _cannot_ have branch-based dependencies][2].

>  Note that packages which use branch-based dependency requirements cannot be depended-upon by packages which use version-based dependency requirements; you should remove branch-based dependency requirements before publishing a version of your package.

So if I create a package that declares a version requirement on XcodeProj:

```swift
// swift-tools-version:5.1

import PackageDescription

let package = Package(
  name: "repro",
  dependencies: [
    .package(url: "https://github.com/tuist/XcodeProj.git", from: "7.0.0"),
  ]
)
```

it'll fail to resolve:
```
$ swift package resolve
Updating https://github.com/tuist/XcodeProj.git
error: the package xcodeproj[https://github.com/tuist/XcodeProj.git] @ 7.2.2 contains incompatible dependencies:
    swift-package-manager[https://github.com/apple/swift-package-manager] @ swift-5.0-RELEASE
```

This is using the Swift 5.1 toolchain that ships with Xcode 11.0:

```
$ swift -version 
Apple Swift version 5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7)
Target: x86_64-apple-darwin18.7.0
$ xcodebuild -version
Xcode 11.0
Build version 11A420a
```

## Proposal 🎉

Not sure what a good workaround would be, since SwiftPM doesn't tag versioned releases and instead publishes a branch for each Swift version 😬 

FWIW, [Carthage recently dropped their dependency on SwiftPM][1] due to difficulties supporting multiple Swift versions and working around bugs in SwiftPM itself.

[1]: https://github.com/Carthage/Carthage/pull/2863
[2]: https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescription.md#package-dependency-requirement

pepibumur:
> Unfortunately, this breaks downstream Swift packages that depend on XcodeProj. There's a restriction in SwiftPM that version-based packages cannot have branch-based dependencies.

[...]

PR #499, opened 2019-11-06
Remove SwiftPM dependency and use Foundation.Process in tests

elliottwilliams:
Resolves https://github.com/tuist/xcodeproj/issues/498

### Short description 📝
SwiftPM's `Basic` has a useful higher-level subprocess API, but depending on SwiftPM itself is heavy and breaks resolution for consumers of XcodeProj (#498). This change removes the SwiftPM dependency and uses Foundation's `Process` to execute subprocesses in XcodeProj's tests.

### Solution 📦
`Process` is marked "complete" in [swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation/blob/master/Docs/Status.md), so the API should be considered portable across Windows, Linux, and macOS.

I've added a private function to `PBXProjIntegrationTests` that provides a similarly-high-level function for executing a command and getting its output. I'd be happy to move it somewhere more reusable if you anticipate needing it from other tests in the future 🙂

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -27,24 +27,6 @@
           "revision": "f14ff47f45642aa5703900980b014c2e9394b6e5",
           "version": "0.9.0"
         }
-      },
-      {
-        "package": "llbuild",
-        "repositoryURL": "https://github.com/apple/swift-llbuild.git",
-        "state": {
-          "branch": "master",
-          "revision": "0a778ca0c51025ffec95de881e59eb35c92f9944",
-          "version": null
-        }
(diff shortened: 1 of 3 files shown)


pepibumur:
Thanks for tackling this @elliottwilliams. Keeping that as a private function in the file is ok. If the need for reusing it in the arises at some point, we can move it somewhere else.

TIL that Process is portable to other platforms. That's great! I'll merge the PR and release a minor version so that you can point the project to this version.

PR #511, opened 2019-12-12
Add enable previews build settings

fortmarek:
### Short description 📝
When creating a new project with SwiftUI checkbox, Xcode sets `ENABLE_PREVIEWS=YES`. It does not do so when setting storyboard as a default, but I have not found any drawbacks when it is set to yes, even if SwiftUI previews are not being used.

### Solution 📦
Add `ENABLE_PREVIEWS=YES` to build settings `.all`

Linked to https://github.com/tuist/tuist/pull/785


kwridan:
Thanks for looking into this @fortmarek 

Perhaps instead of setting it globally at the project level (which will be applied to all targets and projects) it can be set for applications only? Xcode seems to apply that setting to the application target only by default when creating new projects.

There are a few different methods in that file that specify defaults for different product types / platform combos:
(e.g. https://github.com/tuist/XcodeProj/blob/master/Sources/XcodeProj/Utils/BuildSettingsProvider.swift#L191)

That said, it's worth noting that this change will mean all Tuist generated projects will include that setting (existing and new). This may be ok seeing there are similar style defaults pre-defined here and as long as we keep it part of the "recommended" not the "essential" default settings in Tuist, users would still able to opt out if needed.

The alternate to that would be to have generated template's manifest include `Settings(base: ["ENABLE_PREVIEWS": "YES"])` for the application target.

fortmarek:
I have not noticed that it sets it just for the application, my bad, good point @kwridan. Would definitely change it to that if we decide to go down this path.

I realize this change will be for all tuist projects who upgrade the tuist version - but since it defaults to this option for SwiftUI projects and does not have any drawbacks, I figured it might be fine (as long as we keep it in `.recommended`, definitely).

I thought about generating it in the manifest, too, but in my opinion it's better to keep it away from the users, but definitely worth having a discussion about it.

pepibumur:
I agree with @kwridan here. If this is an application-specific build setting I'd limit the scope of the setting to just targets that are apps. Moreover, considering this is an Xcode 11 feature, I'd add it to the project only if the project is being generated with Xcode 11.

That being said, we can either do that in Tuist instead of here, or pass the version of Xcode for which we are returning the default settings and use that information to determine whether the setting should be included in the returned dictionary. What do you think?

fortmarek:
Thanks for the feedback - the way I see it, I think it'd be best to move it to application build settings since I think this option should be available from `xcodeproj`. And then do the filtering based on Xcode in `tuist` to preserve backwards compatibility.

fortmarek:
The setting has been moved to application target, the filtering will be implemented [here](https://github.com/tuist/tuist/pull/814).
The tests will not work until this has been merged, so will wait for that 🙂

yonaskolb:
I'm interested, does anyone know what this build setting actually does? Previews work fine without it

fortmarek:
I'd love to know that, unfortunately, I have not been able to find anything. I'd think it does make some optimizations, as you have mentioned @yonaskolb, but I am not sure if there is any way for us to find out at the moment.

jcampbell05:
It enables Xcode Previews which Swift UI uses to show you the UI you are building but it isn't exclusive to Swift UI you can use it with your own code, the reason it's behind a flag is it isn't supported for older iOS versions so some projects won't want to set it if they are for iOS 12 or earlier.

However it is possible to just use macros so not sure why apple implement this flag, I guess may be for performance reasons for old projects.

Here is more info:
https://www.avanderlee.com/xcode/xcode-previews/

fortmarek:
You're right that it is something that can be used without SwiftUI - but I am still not sure if this explains what the flag actually does; the project is still compilable when you enable it on projects that support versions prior to 13.0 and previews work without it being set to `YES`.

jcampbell05:
Most likely a way of Xcode detecting when it's not enabled so when user manually enables it in Xcode's UI it knows to add certain files.

Perhaps we should try doing this to see ?

fortmarek:
I have tried, but have not been able to find anything, unfortunately. If you come up with some findings, please do post them here, though! 🙂 

But still, since this is a default behavior for Xcode with SwiftUI option, I still think it belongs to the application target and would like to move forward with this [pr](https://github.com/tuist/tuist/pull/814), so do you think we are good to merge @pepibumur?

pepibumur:
> But still, since this is a default behavior for Xcode with SwiftUI option, I still think it belongs to the application target and would like to move forward with this pr, so do you think we are good to merge @pepibumur?

Agree. Fee free to merge it @fortmarek

fortmarek:
Thanks! Will do, but only after https://github.com/tuist/tuist/pull/814 is close to being ready to preserve the backwards compatibility (but I have tough three weeks coming up, so I hope that there is no rush to push this)

PR #516, opened 2020-01-23
Added useTestSelectionWhitelist attribute to XCScheme.TestableReference

basvankuijck:
### Short description 📝
Add the `useTestSelectionWhitelist` attribute to scheme creation so it can either automatically include (or exclude) new tests

### DIFF
--- a/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
@@ -8,6 +8,7 @@ extension XCScheme {
         public var skipped: Bool
         public var parallelizable: Bool
         public var randomExecutionOrdering: Bool
+        public var useTestSelectionWhitelist: Bool
         public var buildableReference: BuildableReference
         public var skippedTests: [SkippedTest]
 
@@ -17,17 +18,20 @@ extension XCScheme {
                     parallelizable: Bool = false,
                     randomExecutionOrdering: Bool = false,
                     buildableReference: BuildableReference,
-                    skippedTests: [SkippedTest] = []) {
+                    skippedTests: [SkippedTest] = [],
+                    useTestSelectionWhitelist: Bool = false) {
             self.skipped = skipped
             self.parallelizable = parallelizable
             self.randomExecutionOrdering = randomExecutionOrdering
             self.buildableReference = buildableReference
             self.skippedTests = skippedTests
+            self.useTestSelectionWhitelist = useTestSelectionWhitelist
         }
 
         init(element: AEXMLElement) throws {
             skipped = element.attributes["skipped"] == "YES"
             parallelizable = element.attributes["parallelizable"] == "YES"
+            useTestSelectionWhitelist = element.attributes["useTestSelectionWhitelist"] == "YES"
             randomExecutionOrdering = element.attributes["testExecutionOrdering"] == "random"
             buildableReference = try BuildableReference(element: element["BuildableReference"])
             if let skippedTests = element["SkippedTests"]["Test"].all, !skippedTests.isEmpty {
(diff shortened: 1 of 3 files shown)


Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -8,6 +8,7 @@ extension XCScheme {
         public var skipped: Bool
         public var parallelizable: Bool
         public var randomExecutionOrdering: Bool
+        public var useTestSelectionWhitelist: Bool
```
pepibumur: Hi @basvankuijck. Thanks for opening this PR. Would you mind making this attribute optional? Since Xcode is able to open projects where this attribute is not defined that means that the attribute is optional and we should reflect it as such.

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -42,6 +46,7 @@ extension XCScheme {
         func xmlElement() -> AEXMLElement {
             var attributes: [String: String] = ["skipped": skipped.xmlString]
             attributes["parallelizable"] = parallelizable ? parallelizable.xmlString : nil
+            attributes["useTestSelectionWhitelist"] = useTestSelectionWhitelist ? useTestSelectionWhitelist.xmlString : nil
```
pepibumur: After making the attribute optional, we should not set it here unless it has a value.

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
yonaskolb: Hi @pepibumur, I don’t understand your reasoning? If this property is not in a project it will be set to false when reading. When writing it will only be set if true. This is the same for the other booleans in this class and many others

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
basvankuijck: I'm not quite sure what to do here.
How is `useTestSelectionWhitelist` different from the `parallelizable` or `testExecutionOrdering` attribute in terms of reading / setting?

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
pepibumur: If a read project has that property as nil, we should keep it as nil to write it back as nil. If we turn nil into a default boolean, we’ll end up creating an attribute when we shouldn’t. That causes diff issues that users will end up reporting an issues.

It’s not an issue when generating projects, but it is when reading and writing existing projects.

pepibumur: In other words, nil != false

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
pepibumur: Sorry. We don’t have to do anything here.

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
yonaskolb: If a project is read where this is nil, then the property can be set to false. When the project is written again, the value won't be written if the value is false, meaning it will end up the same. This causes 0 diffs.

The only theoretical diff is if a project had this explicitly set to false, then reading and writing it would remove the property. This wouldn't happen though as Xcode removes the property when you set it to false in the UI, which is what we are proposing XcodeProj do as well

yonaskolb: This is also the same logic that the other bools in this class have been using for years

pepibumur: This wouldn't happen though as Xcode removes the property when you set it to false in the UI, which is what we are proposing XcodeProj do as well

Xcode writes are optimized for minimizing the diffs after writing the project. If we read, we write nil and as a user of the tool there are no unexpected diffs. We might have missed this in some other properties.

pepibumur:
The changes should be on [master](https://github.com/tuist/XcodeProj/pull/549) now. I'm releasing a new version of XcodeProj that includes the new attribute. Thanks for adding this feature @basvankuijck

PR #517, opened 2020-01-23
Make target attributes optional

pepibumur:
Related https://github.com/tuist/tuist/issues/904

The `PBXProject` `targetAttributes` is not written when the dictionary is empty. That's causing some [issues](https://github.com/tuist/tuist/issues/904) because Fastlane expects the dictionary to be defined in the .pbxproj. Since this is an optional attribute, I'm defining it as such and writing it even if the dictionary is empty.

### DIFF
--- a/Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
+++ b/Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
@@ -176,7 +176,9 @@ extension PBXProject {
         if projectRoots != rhs.projectRoots { return false }
         if targetReferences != rhs.targetReferences { return false }
         if !NSDictionary(dictionary: attributes).isEqual(to: rhs.attributes) { return false }
-        if !NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhs.targetAttributeReferences) { return false }
+        guard let targetAttributeReferences = targetAttributeReferences,
+            let rhsTargetAttributeReferences = rhs.targetAttributeReferences,
+            NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhsTargetAttributeReferences) else { return false }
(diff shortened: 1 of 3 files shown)


Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
```diff
@@ -176,7 +176,9 @@ extension PBXProject {
         if projectRoots != rhs.projectRoots { return false }
         if targetReferences != rhs.targetReferences { return false }
         if !NSDictionary(dictionary: attributes).isEqual(to: rhs.attributes) { return false }
-        if !NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhs.targetAttributeReferences) { return false }
+        guard let targetAttributeReferences = targetAttributeReferences,
+            let rhsTargetAttributeReferences = rhs.targetAttributeReferences,
+            NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhsTargetAttributeReferences) else { return false }
```
yonaskolb: This will return false if both are nil

yonaskolb:
Hi @pepibumur. I don’t understand this change. Wasn’t the only required thing to write the attributes even if empty? Did it become optional for a reason?

pepibumur:
> Hi @pepibumur. I don’t understand this change. Wasn’t the only required thing to write the attributes even if empty? Did it become optional for a reason?

We need to write the dictionary if it's empty because some Fastalane features [depened on it](https://github.com/tuist/tuist/issues/904) being defined. I took the opportunity to make it optional because the attribute is optional for Xcode. Same as [here](https://github.com/tuist/XcodeProj/pull/516), empty dictionary is not the same as not-defined dictionary and I'm just making that distinction.

PR #518, opened 2020-01-27
Generate project references before targets

evandcoleman:
### Short description 📝
Previously, targets were generated before project references. This means that any build files in your main project that reference build files in another project will be generated with non-deterministic UUIDs.

### Solution 📦
Generate project references before generating targets

### Implementation 👩‍💻👨‍💻
This bug was surfaced to me while implementing a new feature in XcodeGen (https://github.com/yonaskolb/XcodeGen/pull/701) which allows you to link with targets in external Xcode projects. Every time the project was generated, the UUIDs of the `PBXBuildFile`s were different. I realized that this is because the `PBXProductReferenceProxy` object still had a `TEMP` UUID by the time it got to generating build phases for the main project. After changing the order of generating, the problem went away because the reference proxy no longer had the temporary UUID.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+### Fixed
+- Ensure references to products in external projects are generated with deterministic UUIDs https://github.com/tuist/XcodeProj/pull/518 by @evandcoleman
+
 ## 7.6.0
 
 ### Changed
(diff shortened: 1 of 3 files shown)


evandcoleman:
@pepibumur done!

yonaskolb:
Thanks @evandcoleman! 
@pepibumur looks like the changelog check is failing, and tests aren’t running

pepibumur:
Thanks a lot @evandcoleman. You can ignore that, it's a bit flakey so I'll have to look into it.

evandcoleman:
@pepibumur Thanks! When can we expect this to be included in a release? Or is that something that I'm able to do and open a PR with the changes?

pepibumur:
> @pepibumur Thanks! When can we expect this to be included in a release? Or is that something that I'm able to do and open a PR with the changes?

I just released 7.7.0 so self-serve.

evandcoleman:
Awesome, thank you!!

PR #519, opened 2020-02-01
Fix target attribute optionality

yonaskolb:
https://github.com/tuist/XcodeProj/pull/517 made a breaking change to make `targetAttributes` optional.

This PR fixes:
- project equality always returning false if `targetAttributes` were optional
- makes `targetAttributes` non optional, as they are always set in both the initializers. The attributes are always still written which solves the aim of https://github.com/tuist/XcodeProj/pull/517

### DIFF
--- a/Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
+++ b/Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
@@ -176,9 +176,7 @@ extension PBXProject {
         if projectRoots != rhs.projectRoots { return false }
         if targetReferences != rhs.targetReferences { return false }
         if !NSDictionary(dictionary: attributes).isEqual(to: rhs.attributes) { return false }
-        guard let targetAttributeReferences = targetAttributeReferences,
-            let rhsTargetAttributeReferences = rhs.targetAttributeReferences,
-            NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhsTargetAttributeReferences) else { return false }
+        if !NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhs.targetAttributeReferences) { return false }
         return super.isEqual(to: rhs)
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+### Fixes
+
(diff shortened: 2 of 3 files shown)


Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
```diff
@@ -176,9 +176,7 @@ extension PBXProject {
         if projectRoots != rhs.projectRoots { return false }
         if targetReferences != rhs.targetReferences { return false }
         if !NSDictionary(dictionary: attributes).isEqual(to: rhs.attributes) { return false }
-        guard let targetAttributeReferences = targetAttributeReferences,
-            let rhsTargetAttributeReferences = rhs.targetAttributeReferences,
-            NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhsTargetAttributeReferences) else { return false }
+        if !NSDictionary(dictionary: targetAttributeReferences).isEqual(to: rhs.targetAttributeReferences) { return false }
```
yonaskolb: Fixes incorrect case were both were nil

pepibumur:
> project equality always returning false if targetAttributes were optional

Awesome! Thanks for tackling this.

> makes targetAttributes non optional, as they are always set in both the initializers. The attributes are always still written which solves the aim of #517

The rule we have been following to decide whether an attribute should be or not optional is that the attribute should be optional if Xcode is able to open a project with the attribute not define. That's the case for `targetAttributes` and therefore should remain as optional. It's an unfortunate breaking change.

yonaskolb:
> The rule we have been following to decide whether an attribute should be or not optional is that the attribute should be optional if Xcode is able to open a project with the attribute not define. That's the case for targetAttributes and therefore should remain as optional. It's an unfortunate breaking change.

As per #517 though, this property is always written so that Fastlane can open it. There's no room for it to be optional

pepibumur:
> As per #517 though, this property is always written so that Fastlane can open it. There's no room for it to be optional

Although I disagree with going down that path, it's not a big deal so let's merge this and revisit it before the next major version to avoid breaking changes.

pepibumur:
I'll release a new version right after the merge

yonaskolb:
Thank you @pepibumur! :)

PR #529, opened 2020-03-11
Optimize bottlenecks

michaeleisel:
### Solution 📦
Replace high-level Swift ways of doing things, which in these cases cause significant overhead, with fast and low-level ways of doing things


CExts.c
```diff
@@ -0,0 +1,59 @@
+//
```
pepibumur: Could you remove the headers from the file to be consistent with the rest of the codebase?

Sources/XcodeProj/Extensions/String+md5.swift
```diff
@@ -21,28 +21,29 @@
...
-                Array(UnsafeBufferPointer(
-                    start: bufferPointer.baseAddress?.assumingMemoryBound(to: UInt8.self),
-                    count: data.count
-                ))
-            }
-
-            let MD5Calculator = MD5(message)
-            let MD5Data = MD5Calculator.calculate()
+    private func toHex(_ char: UInt8) -> UInt8 {
+        if char < 10 {
```
pepibumur: Would you mind adding a comment explaining the rationale behind comparing char with 10?

CExts.c
```diff
@@ -0,0 +1,59 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stddef.h>
+#include <ctype.h>
+#include <string.h>
+#include <stdbool.h>
+
+const char *PATEscapedString(const char *string) {
```
pepibumur: Would you mind documenting this method? What does `PAT` stand for?

CExts.c
```diff
@@ -0,0 +1,59 @@
+//
+//  CExts.cpp
```
pepibumur: Has this file been taken from another open source project? If so, I'd add a reference to it.

Sources/XcodeProj/Extensions/String+md5.swift
```diff
@@ -21,28 +21,29 @@
...
+        } else {
+            return (char - 10) + 65
+        }
+    }
 
-            var MD5String = String()
-            for c in MD5Data {
-                MD5String += String(format: "%02x", c)
+    var md5: String {
+        guard let data = data(using: .utf8, allowLossyConversion: true) else { abort() }
```
pepibumur: Aborting with no message is a terrible user experience for the user. What about passing a string with a explanatory comment of what happened?

Sources/XcodeProj/Extensions/String+md5.swift
```diff
@@ -21,28 +21,29 @@
...
+        return data.withUnsafeBytes { bufferPointer in
+            let buffer = malloc(Int(CC_MD5_DIGEST_LENGTH))!.assumingMemoryBound(to: UInt8.self)
```
pepibumur: The logic here is hard to read. To simplify the understanding and maintenance of it in the future, I'd add a brief explanation of what it does.

pepibumur:
Also, would you mind adding an entry to the CHANGELOG?

CExts.c
michaeleisel: PAT is an arbitrary 3-letter prefix for namespacing. What would you prefer, XCP?

CExts.c
michaeleisel: Sure

michaeleisel:
The big question I'd like to discuss is, how to deal with the `.c` file? SwiftPM doesn't support mixed-language projects, and other package managers have issues as well. This is the first non-Swift file, right?

Here are some options:
- Translate the C to Swift where it makes extensive use of unsafe pointers. This may incur some performance overhead
- Create a separate package, e.g. `XcodeProjCExt`, whose version is synced with that of `XcodeProj`. This is what I do for https://github.com/michaeleisel/ZippyJSON, which puts its non-Swift code in https://github.com/michaeleisel/ZippyJSONCFamily . This is a pain, but also a good long-term investment if it will be good to have more code like this in the future. XcodeProj is used in large projects where it has a large workload, so I think it could be worth it. Just as a thought experiment, if the whole project were written in C++ (which I am not actually suggesting), it could easily be 10x faster.
- ?

pepibumur:
I lean more towards the second option, having an independent package with those extensions. I added you to the contributors team, which has access to this [repository](https://github.com/tuist/XcodeProjCExt) that I just created to host that package. Feel free to push the extension there and the `Package.swift` to define the dependency from this PR.

Thanks a lot @michaeleisel for bringing this improvement. Performance is something we've disregarded for a long time and it's great having your experience with C and your contributions to make XcodeProj as fast as possible.

michaeleisel:
🎉

CExts.c
michaeleisel: no, this is my own

michaeleisel:
just pushed the package. note that the first branch i pushed was not master, it was `me-init`. so, that's now the default branch, which you may want to change in the settings

michaeleisel:
also, now that we're using C and unsafe Swift, in case we're not already, it would be good to run unit tests pre-release for:
- address sanitizer
- thread sanitizer
- memory leaks
- guard malloc

Sources/XcodeProj/Extensions/String+md5.swift
```diff
@@ -21,28 +21,29 @@
...
+            return char + 48
+        } else {
+            return (char - 10) + 65
+        }
+    }
 
-            var MD5String = String()
-            for c in MD5Data {
-                MD5String += String(format: "%02x", c)
+    var md5: String {
```
michaeleisel: i can rewrite the meat of this function in C as well

michaeleisel: it would be faster (swift is not very good at stack allocation of primitive arrays

pepibumur: Agree. If you think it'll be faster let's go for it!

pepibumur:
image

😛Just changed it to be `master` the default. Would you mind pushing your changes to master?

pepibumur:
> also, now that we're using C and unsafe Swift, in case we're not already, it would be good to run unit tests pre-release with:

Haven't done it before. Is there any flag that we can pass to the SPM to run the tests with those things enabled? Feel free to adjust the GitHub Actions workflow as needed.

michaeleisel:
ptal. fyi, i'm having issues with doing `.package(url: "https://github.com/tuist/XcodeProjCExt", .exact("7.8.0"))` (although `.package(path: "../XcodeProjCExt")` works). i get "invalidAccess". any ideas?

michaeleisel:
it looks like the github workflow doesn't do any unit tests? are unit tests run in some other way, pre-release?

michaeleisel:
Got the `invalidAccess` thing fixed (just a cache issue). For swiftpm you can add `--sanitize=<sanitizer>`, e.g. `swift test --sanitize=address`

pepibumur:
Are we good to merge @michaeleisel ?

michaeleisel:
i have one more change i want to make, but i want to make sure it doesn't cause any perf regressions. are there are any perf tests, preferably end-to-end?

pepibumur:
> i have one more change i want to make, but i want to make sure it doesn't cause any perf regressions. are there are any perf tests, preferably end-to-end?

We don't have at the moment. Do you want to add them as part of this PR?

michaeleisel:
OK it looks good to me now. No perf tests, but that could be for another day

michaeleisel:
Are we good to merge?

PR #555, opened 2020-07-04
Add RemotePath to RemoteRunnable

kwridan:
### Short description 📝
Some schemes (e.g. Watch apps) have a `RemotePath` attribute set on the `RemoteRunnable` element.

e.g.

```xml
      <RemoteRunnable
         runnableDebuggingMode = "2"
         BundleIdentifier = "com.apple.Carousel"
         RemotePath = "/AppWithExtensions">
         <BuildableReference
            BuildableIdentifier = "primary"
            BlueprintIdentifier = "42BF876824B0EA0700C4B605"
            BuildableName = "WatchApp.app"
            BlueprintName = "WatchApp"
            ReferencedContainer = "container:AppWithExtensions.xcodeproj">
         </BuildableReference>
      </RemoteRunnable>
```

### Solution 📦
Add the missing attribute.

### Implementation 👩‍💻👨‍💻

- [x] Add `RemotePath`
- [x] Fix how `Equatable` works for runnable (due to there being a class hierarchy)
- [x] Add fixture to test out the new attribute


kwridan:
The fixture turned out to be quite a big one with all the default template files added - if needed I can omit it and only a handful of the `.xcscheme` files as those are the only ones needed for the test.

pepibumur:
@kwridan thanks for adding this. Just a small question: where are we setting that attribute?

kwridan:
Thanks for reviewing @pepibumur - I don't believe this is expose in the Xcode UI / Scheme Editor, however we can see it int he raw `.xcscheme` file when we manually create new targets (e.g. Adding a Watch App / Extension), and Xcode auto generates a corresponding scheme for it.

e.g. 

https://github.com/tuist/XcodeProj/pull/555/files#diff-59648d792327482305a8d10b32beb63cR60

pepibumur:
My bad, when I was reviewing the PR I thought for some reason that it was in the Tuist repository. Changes look good then!

Issue #562, opened 2020-07-23
`.bundle` explicitFileType Wrong in `.pbxproj`


adamkhazi:
## Context / What 🕵️‍♀️
Adding a bundle in Xcode 12 Beta 2 and Xcode 11.5 will add a `.bundle` reference under the `Products` folder. This reference in the `.pbxproj` will have `explicitFileType = wrapper.cfbundle`. Xcodeproj sets this to `explicitFileType = "wrapper.plug-in"` which is incorrect. We can see the `git diff` of what Tuist might generate and what Xcode 11.5 might create below (from the fixture https://github.com/tuist/tuist/tree/master/fixtures/ios_app_with_framework_and_resources):

```diff
-               A16B291CD4C238C56FCD434A /* StaticFramework2Resources.bundle */ = {isa = PBXFileReference; explicitFileType = "wrapper.plug-in"; includeInIndex = 0; path = StaticFramework2Resources.bundle; sourceTree = BUILT_PRODUCTS_DIR; };
+               A16B291CD4C238C56FCD434A /* StaticFramework2Resources.bundle */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = StaticFramework2Resources.bundle; sourceTree = BUILT_PRODUCTS_DIR; };
```

## Proposal 🎉
Change the value of `bundle` inside `Xcode.swift` to `wrapper.cfbundle`.

PR #563, opened 2020-07-23
.bundle `explicitFileType` corrected

adamkhazi:
Resolves https://github.com/tuist/XcodeProj/issues/562

### Short description 📝
Adding a bundle in Xcode 12 Beta 2 and Xcode 11.5 will add a .bundle reference under the Products folder. This reference in the `.pbxproj` will have `explicitFileType = wrapper.cfbundle`. Tuist will use `Xcode.swift` and set the value of `explicitFileType` to the wrong value.

### Solution 📦
Correct the value for `"bundle"` in `Xcode.swift` to use wrapper.cfbundle which mimics the behaviour of Xcode 12 Beta 2 and Xcode 11.5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+### Fixed
+- explicitFileType corrected for .bundle https://github.com/tuist/XcodeProj/pull/563 by @adamkhazi
+
 ## 7.14.0
 
 ### Fixed
--- a/Sources/XcodeProj/Project/Xcode.swift
+++ b/Sources/XcodeProj/Project/Xcode.swift
@@ -102,7 +102,7 @@ public struct Xcode {
         "avi": "video.avi",
         "bin": "archive.macbinary",
         "bmp": "image.bmp",
-        "bundle": "wrapper.plug-in",
+        "bundle": "wrapper.cfbundle",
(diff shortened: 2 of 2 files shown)


adamkhazi:
Hey @pepibumur could I get some clarification as to why this is a breaking change? I am happy to make the changes you suggested.

pepibumur:
> Hey @pepibumur could I get some clarification as to why this is a breaking change? I am happy to make the changes you suggested.

Sure, if I'm not wrong, the `explicitFileType` for bundles has changed in Xcode 12. If we merge this change, and someone uses the latest version of the library to lets say, generate projects, the generated projects might not be compatible with Xcode 11 because bundles will have the `explicitFileType` that Xcode 12 expects.

Correct me if I'm wrong, I might be missing some details.

adamkhazi:
Thanks for the reply @pepibumur 

The `.bundle` references in the `.pbxproj` from Xcode 11.1, 11.5, 12 Beta 2, 12 Beta 3 in my tests have had `explicitFileType = wrapper.cfbundle` when I added a bundle manually through the Xcode UI. 

I also tested adding a bundle in Xcode 12 Beta 3 and opening it in Xcode 11.5 and saw no problems.

pepibumur:
@all-contributors add @adamkhazi for maintenance

PR #565, opened 2020-07-28
Add XCSchemeManagement struct

pepibumur:
### Short description 📝
I'm adding a new entity to the project, `XCSchemeManagement`, which models Xcode's `xcschememanagement.plist` file. Xcode users that files to attach metadata to schemes such as the order of the schemes and whether a scheme is shared or not.

### Implementation 👩‍💻👨‍💻
- Add `XCSchemeManagement ` and some associated entities.
- Add the encoding/decoding logic. 
- Add tests that test the encoding/decoding logic using a fixture.
- Add a test that ensures that the writes produce no git diffs.


wendyliga (approved):
Thanks @pepibumur

wendyliga (approved):

PR #567, opened 2020-08-25
Faster path operations

michaeleisel:
There are inefficiencies in PathKit that I tried to fix in master (rough sketch at https://github.com/kylef/PathKit/pull/72), but that project is basically no longer maintained. So, here we just use my fork directly. I also want to move Xcodeproj to using this version as well. Not only does it deliver nice perf wins, but I will be maintaining my fork for others to make changes for the foreseeable future. It adds a new dep, PathKitCExt, for C/C++-level changes. It's worked well for us at Spotify for the past several months. This is the diff (including one commit that was made after 1.0.0 was released) - https://github.com/kylef/PathKit/compare/73f8e9dca9b7a3078cb79128217dc8f2e585a511...michaeleisel:master

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,7 +9,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/tadija/AEXML", .upToNextMinor(from: "4.4.0")),
-        .package(url: "https://github.com/kylef/PathKit", .upToNextMinor(from: "1.0.0")),
+        .package(url: "https://github.com/michaeleisel/PathKit", .branch("master")),
     ],
     targets: [
         .target(name: "XcodeProj",
--- a/Package.resolved
+++ b/Package.resolved
@@ -12,11 +12,20 @@
       },
       {
         "package": "PathKit",
-        "repositoryURL": "https://github.com/kylef/PathKit",
+        "repositoryURL": "https://github.com/michaeleisel/PathKit",
         "state": {
-          "branch": null,
-          "revision": "73f8e9dca9b7a3078cb79128217dc8f2e585a511",
-          "version": "1.0.0"
+          "branch": "master",
+          "revision": "f5f8bc3c31a787e639b545205309c661c894d8e0",
+          "version": null
+        }
+      },
+      {
+        "package": "PathKitCExt",
+        "repositoryURL": "https://github.com/michaeleisel/PathKitCExt.git",
+        "state": {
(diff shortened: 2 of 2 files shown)


Package.swift
```diff
@@ -9,7 +9,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/tadija/AEXML", .upToNextMinor(from: "4.4.0")),
-        .package(url: "https://github.com/kylef/PathKit", .upToNextMinor(from: "1.0.0")),
+        .package(url: "https://github.com/michaeleisel/PathKit", .branch("master")),
```
dflems, line 12: Feel like this may cause issues with people who depend on `XcodeProj` _and_ `PathKit` (independently) as it may cause conflicts. Maybe spin off a new library with a new name instead of `PathKit` to avoid conflicts? Either that or try and convince the maintainer to hand it over 😄

dflems: Also best to pin to a particular release or commit I think, rather than `master`

mxcl: This will indeed cause conflicts, if this is to happen the fork needs a separate module name.

As an aside, since PathKit is not maintained (and is rubbish IMO) may I suggest [my pathing library](https://github.com/mxcl/Path.swift)? It's better.

michaeleisel: how does its speed compare to this fork?

mxcl: I'm sure the speed isn’t as good as your fork. Though I would want benchmarks on how the pathing library impacts the speed of this project in use personally. Ultimately I am not involved enough in this project to deserve a say on this PR.

michaeleisel: i made specific changes to PathKit to fix bottlenecks for my usage of XcodeGen. i don't know if PathKit, with or without my changes, is faster or slower than other libraries. personally, and i'm not one of the main contributors either, but i'm happy to use anything that is as fast. i will go ahead with this fork unless you want to do perf tests

pepibumur: > Maybe spin off a new library with a new name instead of PathKit to avoid conflicts? Either that or try and convince the maintainer to hand it over

I agree with @dflems here. What about giving it a new name to avoid conflicts? 

In regards to using another pathing library, I'd avoid it unless there're clear advantages over the PathKit's fork.

pepibumur:
I'm closing this one due to inactivity. If there's still interest in making this happen, feel free to reopen it.

Issue #512, opened 2019-12-15
PBXBuildPhase type redudancy


marciniwanicki:
## What 🌱
I spotted `PBXBuildPhase` type is exposed publicly twice in the `PBXBuildPhase.swift`:

- First
```swift
    /// The build phase type of the build phase
    public var buildPhase: BuildPhase {
        fatalError("This property must be overriden")
    }
```

- Second
```swift
    /// Returns the build phase type.
    ///
    /// - Returns: build phase type.
    func type() -> BuildPhase? {
        return buildPhase
    }
```

I wonder if there is a reason behind that duplication?

## Proposal 🎉
We could deprecate and later remove the `type()` from the extension.

pepibumur:
Ouch, good catch. I don't see a reason to keep both. Feel free to open a PR adding a deprecation notice to the one that you think is worth deprecating.

PR #571, opened 2020-10-04
Deprecate PBXBuildPhase.type() function

marciniwanicki:
Resolves https://github.com/tuist/XcodeProj/issues/512

Deprecate `func type() -> BuildPhase?` of `PBXBuildPhase`. The method is redundant and might be removed in the future as the underlying `buildPhase ` property is `public`.


pepibumur (approved):

pepibumur:
@all-contributors add @marciniwanicki for maintenance

PR #585, opened 2020-11-07
Linux support

elliottwilliams:
I've been interested in getting XcodeProj compiling and running on Linux. It turns out that it's possible!

Linux support is obviously a little esoteric, but I think it's valuable for users that want to integrate this tool into automated systems which are linux-based. For example: at Yelp, we have a ton of Linux CI capacity and lots of infra that makes it easy to run git-based workflows on linux servers, and being able to use XcodeProj and XcodeGen from them would simplify our tools.

### Implementation 👩‍💻👨‍💻

- Removes [XcodeProjCExt](https://github.com/tuist/XcodeProjCExt), which was added in #529. This library was based on CommonCrypto which is macOS-only. I've tried to move things back to Swift while retaining the performance gains of the C library:
  - `CommentedString.validString` is a new implementation that operates in a single pass of the string in the optimal case. In my testing I've been able to get this 30% of the performance of the C implementation. If you'd like, I can go further, but it probably means switching to pointer APIs to avoid ARC overhead.
  - `String.md5` includes the [old implementation](https://github.com/michaeleisel/XcodeProj/blob/dae26db4a0164e2ff7af7be29b14611bbe22105f/Sources/XcodeProj/Extensions/String%2Bmd5.swift#L26), but uses CryptoKit when it's available.

- Changes the sourcery template to generate non-overriding functions. Swift's language model [doesn't support overriding in extensions](https://docs.swift.org/swift-book/LanguageGuide/Extensions.html), only allowing it for backwards compatibility with objc. When compiling on linux, objc support is not present and it fails.
  - Sourcery now generates internally-visible comparison functions, and I've added a boilerplate override to each model object to call the appropriate comparison.

- Small fixups:
  - Bumped AEXML to 4.6.0 to fix an issue with importing Foundation on Linux
  - Glibc glob support
  - Support for running `PBXProjIntegrationTests` on a system with no git author (so that tests pass from a Docker image)

### Verification

You can build and test this branch on a linux system using Swift's official docker images:

```sh
$ docker pull swift:5.3  # grab swift from docker hub
$ docker run -it -v $PWD:/code swift:5.3 swift test --package-path /code --enable-test-discovery
```


pepibumur (approved):
First of all, thank you very much for taking your time for bringing support for Linux to the project @elliottwilliams 🙏 🚀. It's something that has been in our backlog for a long time and it finally happened.
I think the performance trade-offs that you are making make total sense for now. We can certainly go further with optimizations but we can do that in follow-up PRs if it turns out we need them. 
One minor suggestion that I'd make before merging the PR is including a GitHub action that ensures changes compile in Linux. We can also update the README of the project to indicate that the library has support for both, Linux and macOS. Also, would you mind updating the CHANGELOG?
After that, we can merge and I'll release a new major version of the tool. I think this improvement really deserves a major bump :).
Again, thanks a lot for taking your time, I appreciate it a lot and I think the users of the library will do it too.

pepibumur:
@all-contributors add @elliottwilliams for code

elliottwilliams:
@pepibumur thanks so much!! Yeah, totally agree with your suggestions.

Should I do anything about the swiftlint violations? I can try to fix any this PR adds, but it looks like there are a number of failures from existing code.

elliottwilliams:
@pepibumur 

> After that, we can merge and I'll release a new major version of the tool. I think this improvement really deserves a major bump :)

qq: Would you actually be open to making this a minor version? I don't think this introduces any API breaks, and I think it'd be easier to advocate for upgrading downstream projects if there isn't a major version bump :)

elliottwilliams:
> Should I do anything about the swiftlint violations? I can try to fix any this PR adds, but it looks like there are a number of failures from existing code.

I take it back, they were mine! Fixed.

pepibumur:
All yours! https://github.com/tuist/XcodeProj/releases/tag/7.18.0

PR #586, opened 2020-11-07
Update AEXML.

ileitch:
This is necessary so that I can use AEXML directly in my project, and also build it on Linux.


ileitch:
I've no idea why CI doesn't like this change...

pepibumur:
I think we can close this one in favor of [this PR](https://github.com/tuist/XcodeProj/pull/585) where @elliottwilliams is bumping the version of AEXML too.

Issue #592, opened 2020-11-22
JSON decoder not properly decoding `defaultConfigurationIsVisible` in some projects


tjwio:
## Context 🕵️‍♀️
Some of our projects generated via BUCK have a plist encoding so the property looks like:
```
<key>defaultConfigurationIsVisible</key>
<false/>
```
instead of something like:
```
defaultConfigurationIsVisible = 0;
```

This causes the JSON encoder to encode the property as a boolean instead of a string, so the JSON decoder fails to decode that boolean properly since it thinks it's a string

## What 🌱
This error is thrown when trying to decode `defaultConfigurationIsVisible`
```
error: typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "defaultConfigurationIsVisible", intValue: nil)], debugDescription: "Expected to decode String but found a number instead.", underlyingError: nil))
```

## Proposal 🎉
https://github.com/tuist/XcodeProj/pull/593

PR #593, opened 2020-11-22
[Fix] JSON decoder not properly decoding `defaultConfigurationIsVisible` in some projects

tjwio:
Resolves https://github.com/tuist/xcodeproj/issues/592

### Short description 📝
This PR should fix decoding `defaultConfigurationIsVisible` (and other booleans) of multiple types of encoding (ie. plist encoded which encodes to a proper boolean or a normal pbxproj which encodes into a string).

### Solution 📦
Try decoding it as a string like before, then an int, then a boolean and finally return nil if it cannot. Notice each try is marked with `try?` so it doesn't actually throw in case there is a type mismatch.

### Implementation 👩‍💻👨‍💻

- [X] Fix decoder process for booleans
- [X] Test out in both pbxproj formats

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/tadija/AEXML",
         "state": {
           "branch": null,
-          "revision": "502c4d43a6cc9c395d19111e09dc62ad834977b5",
-          "version": "4.6.0"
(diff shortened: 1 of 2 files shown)


pepibumur (approved):

pepibumur:
Thanks for fixing this @tjwio 🙏

PR #595, opened 2020-11-26
iOS Support

azonov:


pepibumur:
Hey @azonov. I don't think XcodeProj should have support for iOS because it's not a dependency that you include in the app. If what you'd like to do is to lint your projects, I'd recommend creating your own Swift CLI script that has XcodeProj as a dependency and that you run from your CI pipeline.

Issue #597, opened 2020-12-15
BuildPhaseTests not handling failure cases properly


tjwio:
## Context 🕵️‍♀️
#596 for context, tip of `main`

## What 🌱
Many of the `_____BuildPhaseTests` aren't properly handling failure cases.
For example many of the tests look similar to this, which ensures that the decoder will throw
```
func test_init_fails_whenFilesIsMissing() {
        var dictionary = testDictionary()
        dictionary.removeValue(forKey: "files")
        let data = try! JSONSerialization.data(withJSONObject: dictionary, options: [])
        let decoder = XcodeprojJSONDecoder()
        do {
            _ = try decoder.decode(PBXCopyFilesBuildPhase.self, from: data)
            XCTAssertTrue(false, "Expected to throw an error but it didn't")
        } catch {}
    }
```
This seems to suggest that removing the key "files" causes the decoder to throw. However, this is not the case -- the testDictionary and JSONDecoder (namely `decodeIntIfPresent` see #596) will always throw no matter what because it encodes `dstSubfolderSpec` as an Int even though `decodeIntIfPresent` expects a string to be encoded.
```
func testDictionary() -> [String: Any] {
        [
            "dstPath": "dstPath",
            "buildActionMask": 0,
            "dstSubfolderSpec": 12,
            "files": ["a", "b"],
            "runOnlyForDeploymentPostprocessing": 0,
            "reference": "reference",
        ]
    }
```

## Proposal 🎉
My proposed solution is two fold:
1. Fix `decodeIntIfPresent` to try to decode an int if possible (ie. if decoding as a string then boolean fails).
2. Fix tests to check correct failure case (ie. in a lot of these it's checking for default or nil values).

PR here: #598

tjwio:
Fixed in #598

PR #598, opened 2020-12-15
[#596][#597] Fix `decodeIntIfPresent` and related tests

### LINKED ISSUE #596: JSON decoder not properly decoding `proxyType` in some projects
## Context 🕵️‍♀️
We have some Xcode projects that are generated by BUCK with `proxyType` as a number instead of a string.

## What 🌱
This means that the `decodeIntIfPresent` will throw when it tries to decode it as a string and then bool. We hit this recently in one of our Xcode projects:
```
Error: typeMismatch(Swift.Bool, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "proxyType", intValue: nil)], debugDescription: "Expected to decode Bool but found a number instead.", underlyingError: nil))
```

## Proposal 🎉
PR here: #598

tjwio:
Resolves #596 and #597 

### Short description 📝
Some of our Xcode projects were failing to be decoded due to `proxyType` being encoded as an Int when `decodeIntIfPresent` only tries to decode as String/Bool.

### Solution 📦
- Fix `decodeIntIfPresent` to try to decode `UInt` if decoding as String/Bool fails.
- Fix `_____BuildPhaseTests` to properly check failure cases now that `decodeIntIfPresent` doesn't wrongly throw when trying to decode encoded ints.

- [X] Fix `decodeIntIfPresent
- [X] Fix tests
- [X] Run tests
- [X] Test in failing project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixed
 
 - JSON decoder not properly decoding `defaultConfigurationIsVisible` in some projects [#593](https://github.com/tuist/XcodeProj/pull/593) by [@tjwio](https://github.com/tjwio)
+- JSON decoder not properly decoding `proxyType` in some projects [#596](https://github.com/tuist/XcodeProj/issues/596) by [@tjwio](https://github.com/tjwio)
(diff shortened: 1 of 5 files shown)


pepibumur (approved):
Thanks for fixing this @tjwio 🙏. Would you mind updating the CHANGELOG? We can merge afterward.

tjwio:
@pepibumur updated 🙂

PR #603, opened 2021-02-02
Add runPostActionsOnFailure to scheme

FranzBusch:
### Short description 📝
Schemes have an additional option to run post actions even on failure modes. This is particularly useful when you want to run a script to track something like the build time after any build concluded.

### Solution 📦
I added a new attribute to the scheme `runPostActionsOnFailure` similar to the already existing `buildImplicitDependencies` attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 🚀 Check out the guidelines [here](https://tuist.io/docs/contribution/changelog-guidelines/)
 
 ## Next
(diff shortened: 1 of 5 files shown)


FranzBusch:
@pepibumur Would it be possible for you to take a look here?

pepibumur (approved):
This looks great @FranzBusch, thanks for adding it. Would you mind updating the changelog? After that we can merge and release a new version of XcodeProj

Issue #602, opened 2021-01-08
Remove end of line comments from xcconfig build settings 


damirdavletov:
## Context 🕵️‍♀️
In our projects we use xcconfig files for project configuration and in some cases we use in-line comments like:
`PRODUCT_NAME = $(TARGET_NAME) // FIXME: ideally, should probably be $(TARGET_NAME:rfc1034identifier), but some of our existing names will change`

## What 🌱
Retrieving the above settings with the code like
`let productName = target.buildConfigurationList?.configuration(name: "Release")?.buildSettings["PRODUCT_NAME"] `
Will produce result:
`"$(TARGET_NAME) // FIXME: ideally, should probably be $(TARGET_NAME:rfc1034identifier), but some of our existing names will change"`
Expected result: 
`"$(TARGET_NAME)"`

## Proposal 🎉
Improve XCConfigParser.settingRegex to cut off end of line comments

PR #604, opened 2021-02-15
Add inline comments support to the `xcconfig` parser

dive:
Resolves https://github.com/tuist/XcodeProj/issues/602

### Short description 📝
Add support for inline comments in `xcconfig` files.

### Solution 📦
Extend the existed regular expression to support `xcconfig` comments and use them as a full-stop condition.

### Implementation 👩‍💻👨‍💻
`XCConfig.swift` regular expression (`settingRegex`) extended with a positive lookahead condition as the last statement. Previously, the expression handles the end of the line ('$') as a final grouping statement and now it parses up until the end of the line or up to the comment indication (`//`) with the `(?=$|\\/\\/)` group.

- [X] Update `XCConfigTests` with failing scenarios
- [X] Adjust the `settingRegex` `NSRegularExpression` to support comments
- [X] Extend `XCConfigTests` with edge-cases

P.S. You can play with the updated regular expression [here](https://regexr.com/5mg3v).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - JSON decoder not properly decoding `defaultConfigurationIsVisible` in some projects [#593](https://github.com/tuist/XcodeProj/pull/593) by [@tjwio](https://github.com/tjwio)
 - JSON decoder not properly decoding `proxyType` in some projects [#596](https://github.com/tuist/XcodeProj/issues/596) by [@tjwio](https://github.com/tjwio)
 - BuildPhaseTests not handling failure cases properly [#597](https://github.com/tuist/XcodeProj/issues/597) by [@tjwio](https://github.com/tjwio)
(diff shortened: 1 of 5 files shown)


adellibovi:
Nice fix @dive!

pepibumur (approved):
Would you mind updating the changelog @dive ? We can merge afterward.

fortmarek (approved):
Nice! This probably will not catch another edge case where the setting is followed by multiline comment `/*` but I am not sure if that's something we want to account for

dive:
> Would you mind updating the changelog @dive ? We can merge afterward.

@pepibumur, added. Thanks!

> Nice! This probably will not catch another edge case where the setting is followed by multiline comment `/*` but I am not sure if that's something we want to account for

@fortmarek, as far as I know, `xcconfig` does not support multiline comments. There is no official specification but Xcode ignores them for sure. The only additional edge-case I am aware of is `NAME="https://example.com"` case but even the current Xcode parser does not allow such entries in `xcconfig`.

fortmarek:
Thanks @dive for the explanation, makes sense 👍

PR #607, opened 2021-04-21
🛑 Prevent overwriting identical workspace data

ferologics:
Partially resolves https://github.com/tuist/tuist/issues/2809 ~for projects using `App.xcodeproj` -- without `App.xcworkspace` at project root. I haven't investigated the issue beyond the scope of `.xcodeproj` generation.~

Edit: The implementation has been updated to prevent overwriting identical workspace data (regardless of whether it's a project or a workspace)

### Short description 📝
Prevent overwriting identical workspace data

### Solution 📦
Check if previous workspace data is equal to newly generated workspace data.

### Implementation 👩‍💻👨‍💻
- [x] Add a check in `XCWorkspace` that prevents overwriting of identical data

How can I test this change @fortmarek? I ran into complications when trying with `1.40.0` tag via `swift build -c release --product tuist` and using that with Xcode `12.4` version:

```
Users/ferologics/Dev/tuist/.build/x86_64-apple-macosx/release/tuist generate --project-only                                     ✔
The 'swiftc' was interrupted with a signal 11 and message:
Stack dump:
0.      Program arguments: /Applications/Xcode-12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift -frontend -interpret /Users/ferologics/Globekeeper/Connect/Tuist/Config.swift -enable-objc-interop -sdk /Applications/Xcode-12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk -I /Users/ferologics/Dev/tuist/.build/x86_64-apple-macosx/release -F /Users/ferologics/Dev/tuist/.build/x86_64-apple-macosx/release -suppress-warnings -target-sdk-version 11.1 -module-name Config -lProjectDescription -framework ProjectDescription -- --tuist-dump
1.      Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
2.      While running user code "/Users/ferologics/Globekeeper/Connect/Tuist/Config.swift"
0  swift                       0x0000000111d7c615 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                       0x0000000111d7b615 llvm::sys::RunSignalHandlers() + 85
2  swift                       0x0000000111d7cbcf SignalHandler(int) + 111
3  libsystem_platform.dylib    0x00007fff206a8d7d _sigtramp + 29
4  libsystem_platform.dylib    0x00007fdedd704ef0 _sigtramp + 18446743935146901904
5  libProjectDescription.dylib 0x0000000117195165 $s18ProjectDescription6ConfigV17GenerationOptionsOSEAASE6encode2toys7Encoder_p_tKFTW + 21
6  libswiftCore.dylib          0x00007fff2ccae3a7 $sSE6encode2toys7Encoder_p_tKFTj + 7
7  libswiftFoundation.dylib    0x00007fff306b17bf $s10Foundation13__JSONEncoder33_12768CA107A31EF2DCE034FD75B541C9LLC4box_ySo8NSObjectCSgSE_pKF + 2591
8  libswiftFoundation.dylib    0x00007fff3076af76 $s10Foundation29_JSONUnkeyedEncodingContainer33_12768CA107A31EF2DCE034FD75B541C9LLV6encodeyyxKSERzlF + 550
9  libswiftFoundation.dylib    0x00007fff3076bf49 $s10Foundation29_JSONUnkeyedEncodingContainer33_12768CA107A31EF2DCE034FD75B541C9LLVs07UnkeyedcD0AAsAEP6encodeyyqd__KSERd__lFTW + 9
10 libswiftCore.dylib          0x00007fff2ca3872b $sSasSERzlE6encode2toys7Encoder_p_tKF + 507
11 libswiftCore.dylib          0x00007fff2ca38828 $sSayxGSEsSERzlSE6encode2toys7Encoder_p_tKFTW + 24
12 libswiftCore.dylib          0x00007fff2ccae3a7 $sSE6encode2toys7Encoder_p_tKFTj + 7
13 libswiftFoundation.dylib    0x00007fff306b17bf $s10Foundation13__JSONEncoder33_12768CA107A31EF2DCE034FD75B541C9LLC4box_ySo8NSObjectCSgSE_pKF + 2591
14 libswiftFoundation.dylib    0x00007fff306b60ae $s10Foundation27_JSONKeyedEncodingContainer33_12768CA107A31EF2DCE034FD75B541C9LLV6encode_6forKeyyqd___xtKSERd__lF + 2734
15 libswiftFoundation.dylib    0x00007fff306b55f5 $s10Foundation27_JSONKeyedEncodingContainer33_12768CA107A31EF2DCE034FD75B541C9LLVyxGs05KeyedcD8ProtocolAAsAFP6encode_6forKeyyqd___0P0QztKSERd__lFTW + 21
16 libswiftCore.dylib          0x00007fff2ca21dce $ss26_KeyedEncodingContainerBoxC6encode_6forKeyyqd___qd_0_tKSERd__s06CodingG0Rd_0_r0_lF + 206
17 libswiftCore.dylib          0x00007fff2ca19c34 $ss22KeyedEncodingContainerV6encode_6forKeyyqd___xtKSERd__lF + 36
18 libProjectDescription.dylib 0x0000000117196fa7 $s18ProjectDescription6ConfigV6encode2toys7Encoder_p_tKF + 247
19 libProjectDescription.dylib 0x0000000117197417 $s18ProjectDescription6ConfigVSEAASE6encode2toys7Encoder_p_tKFTW + 39
20 libswiftCore.dylib          0x00007fff2ccae3a7 $sSE6encode2toys7Encoder_p_tKFTj + 7
21 libswiftFoundation.dylib    0x00007fff306b17bf $s10Foundation13__JSONEncoder33_12768CA107A31EF2DCE034FD75B541C9LLC4box_ySo8NSObjectCSgSE_pKF + 2591
22 libswiftFoundation.dylib    0x00007fff3075f8dc $s10Foundation11JSONEncoderC6encodeyAA4DataVxKSERzlF + 460
23 libswiftFoundation.dylib    0x00007fff3077c84e $s10Foundation11JSONEncoderC6encodeyAA4DataVxKSERzlFTj + 14
24 libProjectDescription.dylib 0x000000011719580a $s18ProjectDescription12dumpIfNeededyyxSERzlFAA6ConfigV_Tg5 + 298
25 libProjectDescription.dylib 0x00000001171955ec $s18ProjectDescription6ConfigV23compatibleXcodeVersions5cloud5cache7plugins17generationOptionsAcA010CompatibleeF0O_AA5CloudVSgAA5CacheVSgSayAA14PluginLocationVGSayAC010GenerationK0OGtcfC + 412
26 libProjectDescription.dylib 0x000000011743b077 $s18ProjectDescription6ConfigV23compatibleXcodeVersions5cloud5cache7plugins17generationOptionsAcA010CompatibleeF0O_AA5CloudVSgAA5CacheVSgSayAA14PluginLocationVGSayAC010GenerationK0OGtcfC + 2776103
27 swift                       0x000000010d937a6f llvm::orc::runAsMain(int (*)(int, char**), llvm::ArrayRef<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, llvm::Optional<llvm::StringRef>) + 2495
28 swift                       0x000000010d914e5b swift::RunImmediately(swift::CompilerInstance&, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > const&, swift::IRGenOptions const&, swift::SILOptions const&, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >&&) + 6747
29 swift                       0x000000010d8eff7a performCompileStepsPostSILGen(swift::CompilerInstance&, swift::CompilerInvocation const&, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::PointerUnion<swift::ModuleDecl*, swift::SourceFile*>, swift::PrimarySpecificPaths const&, int&, swift::FrontendObserver*) + 2426
30 swift                       0x000000010d8dff4e swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 21134
31 swift                       0x000000010d860c27 main + 1255
32 libdyld.dylib               0x00007fff2067ef3d start + 1
33 libdyld.dylib               0x0000000000000015 start + 18446603339972481241
```

### DIFF
--- a/Sources/XcodeProj/Project/XcodeProj.swift
+++ b/Sources/XcodeProj/Project/XcodeProj.swift
@@ -130,6 +130,12 @@ extension XcodeProj: Writable {
     /// - Parameter override: if workspace should be overridden. Default is true.
     ///   If false will throw error if workspace already exists at the given path.
     public func writeWorkspace(path: Path, override: Bool = true) throws {
+
+        // prevent overwriting identical workspace data
+        if let prevData = try? XCWorkspaceData(path: path), prevData == workspace.data {
+            return
+        }
+
         try workspace.write(path: XcodeProj.workspacePath(path), override: override)
     }
 
--- a/Sources/XcodeProj/Project/XcodeProj.swift
+++ b/Sources/XcodeProj/Project/XcodeProj.swift
@@ -130,7 +130,7 @@ extension XcodeProj: Writable {
     /// - Parameter override: if workspace should be overridden. Default is true.
     ///   If false will throw error if workspace already exists at the given path.
     public func writeWorkspace(path: Path, override: Bool = true) throws {
-        guard workspace.data != (try? XCWorkspace(path: path))?.data else { return }
+        guard let p = path.glob("*.xcworkspacedata").first, workspace.data != (try? XCWorkspaceData(path: p)) else { return }
         try workspace.write(path: XcodeProj.workspacePath(path), override: override)
     }
 
--- a/Sources/XcodeProj/Project/XcodeProj.swift
+++ b/Sources/XcodeProj/Project/XcodeProj.swift
@@ -130,7 +130,6 @@ extension XcodeProj: Writable {
     /// - Parameter override: if workspace should be overridden. Default is true.
     ///   If false will throw error if workspace already exists at the given path.
     public func writeWorkspace(path: Path, override: Bool = true) throws {
-        guard let p = path.glob("*.xcworkspacedata").first, workspace.data != (try? XCWorkspaceData(path: p)) else { return }
         try workspace.write(path: XcodeProj.workspacePath(path), override: override)
     }
 
--- a/Sources/XcodeProj/Project/XcodeProj.swift
+++ b/Sources/XcodeProj/Project/XcodeProj.swift
@@ -130,12 +130,7 @@ extension XcodeProj: Writable {
     /// - Parameter override: if workspace should be overridden. Default is true.
     ///   If false will throw error if workspace already exists at the given path.
     public func writeWorkspace(path: Path, override: Bool = true) throws {
-
-        // prevent overwriting identical workspace data
-        if let prevData = try? XCWorkspaceData(path: path), prevData == workspace.data {
-            return
-        }
-
+        guard workspace.data != (try? XCWorkspaceData(path: path)) else { return }
         try workspace.write(path: XcodeProj.workspacePath(path), override: override)
     }
 
--- a/Sources/XcodeProj/Project/XcodeProj.swift
+++ b/Sources/XcodeProj/Project/XcodeProj.swift
@@ -130,7 +130,7 @@ extension XcodeProj: Writable {
     /// - Parameter override: if workspace should be overridden. Default is true.
     ///   If false will throw error if workspace already exists at the given path.
     public func writeWorkspace(path: Path, override: Bool = true) throws {
-        guard workspace.data != (try? XCWorkspaceData(path: path)) else { return }
+        guard workspace.data != (try? XCWorkspace(path: path))?.data else { return }
         try workspace.write(path: XcodeProj.workspacePath(path), override: override)
     }
 
--- a/Sources/XcodeProj/Workspace/XCWorkspace.swift
+++ b/Sources/XcodeProj/Workspace/XCWorkspace.swift
@@ -59,7 +59,7 @@ public final class XCWorkspace: Writable, Equatable {
                 // there's no need to overwrite the contents
                 // this mitigates Xcode forcing users to
                 // close and re-open projects/workspaces
-                // on regneration.
+                // on regeneration.
                 return
             }
             try dataPath.delete()
--- a/Sources/XcodeProj/Workspace/XCWorkspace.swift
+++ b/Sources/XcodeProj/Workspace/XCWorkspace.swift
@@ -53,6 +53,15 @@ public final class XCWorkspace: Writable, Equatable {
     public func write(path: Path, override: Bool = true) throws {
         let dataPath = path + "contents.xcworkspacedata"
         if override, dataPath.exists {
+            if let existingContent: String = try? dataPath.read(),
+               existingContent == data.rawContents() {
+                // Raw data matches what's on disk
+                // there's no need to overwrite the contents
+                // this mitigates Xcode forcing users to
+                // close and re-open projects/workspaces
+                // on regneration.
+                return
+            }
             try dataPath.delete()
         }
         try dataPath.mkpath()
--- a/Sources/XcodeProj/Workspace/XCWorkspaceData.swift
+++ b/Sources/XcodeProj/Workspace/XCWorkspaceData.swift
@@ -36,20 +36,24 @@ extension XCWorkspaceData: Writable {
 
         self.init(children: children)
     }
-
-    // MARK: - <Writable>
-
-    public func write(path: Path, override: Bool = true) throws {
+    
+    func rawContents() -> String {
         let document = AEXMLDocument()
         let workspace = document.addChild(name: "Workspace", value: nil, attributes: ["version": "1.0"])
         _ = children
             .map { $0.xmlElement() }
             .map(workspace.addChild)
+        return document.xmlXcodeFormat
+    }
 
+    // MARK: - <Writable>
+    
+    public func write(path: Path, override: Bool = true) throws {
+        let rawXml = rawContents()
         if override, path.exists {
             try path.delete()
         }
-        try path.write(document.xmlXcodeFormat)
+        try path.write(rawXml)
     }
 }
 


Sources/XcodeProj/Project/XcodeProj.swift
```diff
@@ -130,6 +130,12 @@ extension XcodeProj: Writable {
     /// - Parameter override: if workspace should be overridden. Default is true.
     ///   If false will throw error if workspace already exists at the given path.
     public func writeWorkspace(path: Path, override: Bool = true) throws {
+
+        // prevent overwriting identical workspace data
+        if let prevData = try? XCWorkspaceData(path: path), prevData == workspace.data {
```
fortmarek, line 135: I'd rewrite this into a more readable and concise version (IMHO)
```swift
guard workspace.data != try? XCWorkspaceData(path) else { return } 
```

fortmarek:
@ferologics to resolve the error, could you try running `swift build -c release --product ProjectDescription` first?

Sources/XcodeProj/Project/XcodeProj.swift
ferologics: Good point, that reads so much better! Done ✅

fortmarek: I'd also modify the comment that describes why overwriting the data is bad (something like it forces the Xcode project to be reloaded)

ferologics:
@fortmarek thanks, tho somehow I'm still getting 
image. 

Either I'm testing this wrong, or maybe this check is not enough? The `contents.xcworkspacedata` is still being overwritten.

kwridan (commented):
Thanks @ferologics! This has been one annoying quirk of regenerating projects that would be great to solve.

@yonaskolb would this have any negative side effects for XcodeGen?

From my testing, I noticed if you have a workspace with several projects open, Xcode seems to tolerate regenerating the individual projects, however if you regenerate the entire workspace (including the `.xcworkspace`) that's when we start hitting the issue of needing to  close and re-open Xcode.

Prototyping locally something like this may do the trick:

- extracted a `rawContents()` method in `XCWorkspaceData` which can be used by its write method

```swift
// XCWorkspaceData.swift

   func rawContents() -> String {
        let document = AEXMLDocument()
        let workspace = document.addChild(name: "Workspace", value: nil, attributes: ["version": "1.0"])
        _ = children
            .map { $0.xmlElement() }
            .map(workspace.addChild)
        return document.xmlXcodeFormat
    }

    // MARK: - <Writable>

    public func write(path: Path, override: Bool = true) throws {
        let rawXml = rawContents()
        if override, path.exists {
            try path.delete()
        }
        try path.write(rawXml)
    }
```

- Updated `XCWorkspace.write` to check the raw contents before deleting 

```swift
// XCWorkspace.swift

   public func write(path: Path, override: Bool = true) throws {
        let dataPath = path + "contents.xcworkspacedata"
        if override, dataPath.exists {
            if let existingContent: String = try? dataPath.read(),
               existingContent == data.rawContents() {
                // Raw data matches what's on disk
                // there's no need to overwrite the contents
                // this mitigates Xcode forcing users to
                // close and re-open projects/workspaces
                // on regneration.
                return
            }
            try dataPath.delete()
        }
        try dataPath.mkpath()
        try data.write(path: dataPath)
    }
```

Benchmarking reading the raw contents vs deserialisation they were roughly the same - not sure now which is better 🤷‍♂️

ferologics:
Thanks for giving this a thorough look @kwridan 🙇 I agree with your observations about both the responsibility and the root cause. I adjusted the code according to your suggestion.

ferologics:
After spending more time debugging this with a project I still got the occasional "" upon regenerating the project after some time. A closer look at the `contents.xcworkspacedata` shows that this data is slightly different from the "clean slate" data that we use to verify the equality:

Clean slate:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Workspace
   version = "1.0">
</Workspace>
```

Existing:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Workspace
   version = "1.0">
   <FileRef
      location = "self:">
   </FileRef>
</Workspace>
```

The problematic `<FileRef/>` is added at some point in the lifetime of the project. I am not exactly sure what it's doing there, or what it's purpose is @kwridan @fortmarek, but perhaps it can be added to the "clean slate" xml data?

Sources/XcodeProj/Workspace/XCWorkspace.swift
```diff
@@ -53,6 +53,15 @@ public final class XCWorkspace: Writable, Equatable {
     public func write(path: Path, override: Bool = true) throws {
         let dataPath = path + "contents.xcworkspacedata"
         if override, dataPath.exists {
+            if let existingContent: String = try? dataPath.read(),
+               existingContent == data.rawContents() {
+                // Raw data matches what's on disk
+                // there's no need to overwrite the contents
+                // this mitigates Xcode forcing users to
+                // close and re-open projects/workspaces
+                // on regneration.
```
fortmarek, line 62:
```suggestion
                // on regeneration.
```

Sources/XcodeProj/Workspace/XCWorkspace.swift
```diff
@@ -53,6 +53,15 @@ public final class XCWorkspace: Writable, Equatable {
     public func write(path: Path, override: Bool = true) throws {
         let dataPath = path + "contents.xcworkspacedata"
         if override, dataPath.exists {
+            if let existingContent: String = try? dataPath.read(),
```
fortmarek, line 56: IMHO guard is more explicit if you want to do early exit (as in this case)

fortmarek (commented):
> The problematic <FileRef/> is added at some point in the lifetime of the project. I am not exactly sure what it's doing there, or what it's purpose is

This comes from `XCWorkspaceDataElementLocationType` that has a `case `self`(String) // Single project workspace in xcodeproj directory`. I have not dived deeper into why this one is added as I don't really follow what's the issue with it - the generation should be deterministic, so what exact modification / command creates this change in the data? But I believe we are fine with the fact that this data changes as long as it happens only when the generated `.xcworkspace` is different as well.

Sources/XcodeProj/Workspace/XCWorkspace.swift
ferologics: Yes, but personally I only like to use guards where it reads well.

In this case we're looking for identical data, which suggests `==` and not `!=`.

There's something to be said about the simplicity of an `if`, and not having to backtrack the negative boolean checks that read: "guard if not the case otherwise that" (as opposed to "if case then").

IMHO early escapes should be easy to comprehend, but the `guard` statement complicates them more often than not.

Other than it being a signpost for an early escape, are there other advantages to using a `guard` in this case that I'm unaware of?

If you feel particularly strong on his point I'm fine going with the `guard` asw 🙂

kwridan:
What you're witness I believe an internal Xcode behaviour:

My current understanding of it as follows
- Projects need to always operate with a workspace
- When launching a standalone `.xcodeproj` an implicit workspace is used that is nested within the `.xcodeproj` container
    - `App/App.xcodeproj/project.xcworkspace/contents.xcworkspacedata`
    - The implicit workspace needs to reference the project and as such adds the following on launch

```xml
   <FileRef
      location = "self:">
   </FileRef>
```

- When launching within a dedicated workspace (e.g. when hosting one or more projects), Xcode doesn't use the implicit workspace
    - `App.xcworkspace/contents.xcworkspacedata` 

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Workspace
   version = "1.0">
   <FileRef
      location = "group:App/App.xcodeproj">
   </FileRef>
   <FileRef
      location = "group:Modules/MyFramework.xcodeproj">
   </FileRef>
</Workspace>
```

As such we'd need to figure out which `XCWorkspace` we're dealing with if we were to modify the clean slate default - not sure `XcodeProj` has enough information to decide that at the library level, possibly something we can infer at the `Tuist` level?

kwridan (approved):
Thanks for the updates

pepibumur (approved):

PR #609, opened 2021-05-05
Fix Wrong Product Type String for tvOS App Extensions

rmnblm:
### Short description 📝
I was working on https://github.com/tuist/tuist/pull/2793 and realised that the product type "com.apple.product-type.tv-app-extension" caused the top shelf to crash. There is a solution posted on [StackOverflow]( https://stackoverflow.com/questions/58504864/tvos-13-top-shelf-nsextensionprincipalclass-product-module-name-contentprovid). In addition, I compared the product type to the product type in Apple's [example](https://developer.apple.com/documentation/tvservices/building_a_full_screen_top_shelf_extension). 

### Solution 📦
The product type string of `PBXProductType.tvExtension` was updated to 

``` swift
case tvExtension = "com.apple.product-type.app-extension"
```

### DIFF
--- a/Sources/XcodeProj/Objects/Targets/PBXProductType.swift
+++ b/Sources/XcodeProj/Objects/Targets/PBXProductType.swift
@@ -18,7 +18,7 @@ public enum PBXProductType: String, Decodable {
     case watch2AppContainer = "com.apple.product-type.application.watchapp2-container"
     case watchExtension = "com.apple.product-type.watchkit-extension"
     case watch2Extension = "com.apple.product-type.watchkit2-extension"
-    case tvExtension = "com.apple.product-type.tv-app-extension"
+    case tvExtension = "com.apple.product-type.app-extension"
     case messagesApplication = "com.apple.product-type.application.messages"
     case messagesExtension = "com.apple.product-type.app-extension.messages"
     case stickerPack = "com.apple.product-type.app-extension.messages-sticker-pack"


kwridan (commented):
Thanks for looking into this @rmnblm 

I made a new project with a tvOS app and created all available extension target, and here are all the product types I encountered:

```
productType = "com.apple.product-type.application"; /*TV App */
productType = "com.apple.product-type.app-extension"; /* Top Shelf Extension */
productType = "com.apple.product-type.tv-app-extension"; /* Intents Extension */
productType = "com.apple.product-type.tv-broadcast-extension"; /* Broadcast extension */
productType = "com.apple.product-type.tv-broadcast-extension"; /* Broadcast setup extension */
productType = "com.apple.product-type.tv-broadcast-extension"; /* Broadcast UI extension */
```

It seems both `com.apple.product-type.tv-app-extension` and `com.apple.product-type.app-extension` are correct based on the product type in use. 

Changing the value to `com.apple.product-type.app-extension` will sadly cause a clash with `case appExtension` which shares the same value.

Perhaps dedicated `.topShelfExtension` and `.tvIntentsExtension` product types can be added to Tuist which internally would leverage the appropriate underlying XcodeProj product types?

rmnblm:
Hey @kwridan. Sorry for the delay, it's been a busy 3 weeks. How shall I proceed with your feedback?

kwridan:
>  Sorry for the delay, it's been a busy 3 weeks. How shall I proceed with your feedback?

No worries

I believe there's little we can do on the `XcodeProj` side, what we could look to do is in `tuist` introduce new dedicated product types:

- `tvTopShelfExtension`
- `tvIntentsExtension`

And within [`Product+Core.swift`](https://github.com/tuist/tuist/blob/main/Sources/TuistCore/ValueGraph/ModelExtensions/Product+Core.swift#L5) in tuist, map those to the appropriate underlying `XcodeProj` types _(perhaps with a comment explaining the why they map to different underlying types as I'm sure it will look like an error when we revisit in the future 😅)_:

- `tvTopShelfExtension` > `.appExtension`
- `tvIntentsExtension` > `.tvExtension`

That could be a start, and we can see what else is needed to support the new product type. Does that sound reasonable?

rmnblm:
HI @kwridan 

I added your suggested changes to my [PR](https://github.com/tuist/tuist/pull/2793). 
I think we can close this PR here, yes?

fortmarek:
Let's close this in favor of the PR in tuist

PR #616, opened 2021-05-18
Fix linux build and autogenerate XCScheme equality

elliottwilliams:
### Short description 📝

Swift 5.4 changed something that caused the linux build to (correctly) fail to compile an @objc declaration. This change fixes that and uses the same Sourcery generation for XCScheme objects that we use for PBXObjects.

### Solution 📦

As discussed in #585, the only reason for the @objc declaration was to permit swift to override functions from an extension—objc interop is not actually a concern. This change reimplements the equality logic for XCScheme objects to be like PBXObjects: using memberwise equality functions generated by sourcery, and a dynamic override on each XCScheme class that calls into the memberwise function.

Tested the generated interface with these changes:

    swift build -Xswiftc -emit-module-interface -Xswiftc -swift-version -Xswiftc 5

The only change to the interface is that RemoteRunnable now exposes an `==` method:

```diff
+    public static func == (lhs: XcodeProj.XCScheme.RemoteRunnable, rhs: XcodeProj.XCScheme.RemoteRunnable) -> Swift.Bool
```

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/tadija/AEXML",
         "state": {
           "branch": null,
-          "revision": "502c4d43a6cc9c395d19111e09dc62ad834977b5",
-          "version": "4.6.0"
+          "revision": "8623e73b193386909566a9ca20203e33a09af142",
+          "version": "4.5.0"
         }
       },
       {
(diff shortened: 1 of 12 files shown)


elliottwilliams:
@yonaskolb did this in #615, so I'll close this!

PR #630, opened 2021-08-23
Fix indentation in example

keith:
This also adds a trailing `,` for easier copy pasting to projects with existing dependencies


danyf90 (approved):

pepibumur:
Thanks @keith for pushing that fix upstream.

PR #631, opened 2021-08-23
Remove unnecessary force unwrap

keith:
This validates at least one path exists, and reuses that variable instead


fortmarek (approved):
Makes sense, thanks @keith!

PR #635, opened 2021-09-12
Improve performance of commented string

adellibovi:
### Short description 📝
This PR improves the performance of commented string.

### Solution 📦
Most of the case we do not need to escape, having an early exit saves a lot of copies and allocation, in my case I got the time spent on `validString` from 9s to 1.6s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
-- mprove performance of commented string [#635](https://github.com/tuist/XcodeProj/pull/635) by [@adellibovi](https://github.com/adellibovi)
+- Improve performance of commented string [#635](https://github.com/tuist/XcodeProj/pull/635) by [@adellibovi](https://github.com/adellibovi)
 
 ## 8.0.0 - Amor
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Utils/CommentedString.swift
```diff
@@ -81,19 +40,15 @@ struct CommentedString {
         default: break
         }
 
-        var needsQuoting = false
-        var matchingInvalidPrefix: Trie = self.invalidStrings
+        if string.rangeOfCharacter(from: CommentedString.invalidCharacters) == nil {
+            if string.rangeOfCharacter(from: CommentedString.specialCheckCharacters) == nil {
+                return string
+            } else if !string.contains("//")  && !string.contains("___") {
```
fortmarek, line 46:
```suggestion
            } else if !string.contains("//") && !string.contains("___") {
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Next
 
+### Changed
+
+- mprove performance of commented string [#635](https://github.com/tuist/XcodeProj/pull/635) by [@adellibovi](https://github.com/adellibovi)
```
fortmarek, line 7:
```suggestion
- Improve performance of commented string [#635](https://github.com/tuist/XcodeProj/pull/635) by [@adellibovi](https://github.com/adellibovi)
```

fortmarek (approved):
🚀  Just to double-check - have you made sure there are tests in place that would have caught possible regression?

adellibovi:
> 🚀  Just to double-check - have you made sure there are tests in place that would have caught possible regression?

Yep, there were already a good collection of cases under `CommentedStringTests` ☺️
Also tested the generation with a Tuist integration

Issue #657, opened 2021-12-13
`WorkspaceSettings` can't be instantiated by consumers


jakeatoms:
## Context 🕵️‍♀️
I'm trying to address this Tuist issue: https://github.com/tuist/tuist/issues/3487, but before I can open that PR, I need to address the lack of a public initializer on `WorkspaceSettings`.

## What 🌱
`WorkspaceSettings` initializer is `internal` instead of `public`

## Proposal 🎉
I propose making the initializer public. I also propose making the boolean non-optional to simplify the API and make the behavior explicit for consumers.

I'll have a PR open shortly.

PR #658, opened 2021-12-13
Make `WorkspaceSettings` initializer `public`

jakeatoms:
Resolves https://github.com/tuist/xcodeproj/issues/657

### Short description 📝
This PR exposes the initializer of `WorkspaceSettings`, so we can use it in `Tuist`. ~It also makes the `autoCreateSchemes` a non-optional boolean.~

### Solution 📦
I'm working on this Tuist issue: https://github.com/tuist/tuist/issues/3487
The gist is that when a project specifies its own schemes, Xcode begins to nag when the workspace is opened. I'm working on the related Tuist PR, but before I can proceed, I need to be able to initialize this type.
~I also think that optional booleans are a source of confusion as their behavior is ambiguous. Making it required will provide a clear behavior to the consumer and have a concrete result.~

### Implementation 👩‍💻👨‍💻

- [x] Mad the initializer `public`.
- [x] Tested this change locally using my local changes to Tuist and successfully generated the file.

### DIFF
--- a/Sources/XcodeProj/Project/WorkspaceSettings.swift
+++ b/Sources/XcodeProj/Project/WorkspaceSettings.swift
@@ -37,7 +37,7 @@ public class WorkspaceSettings: Codable, Equatable, Writable {
     public var derivedDataCustomLocation: String?
 
     /// When true, Xcode auto-creates schemes in the project.
-    public var autoCreateSchemes: Bool
+    public var autoCreateSchemes: Bool?
 
     /// Decodable coding keys.
     ///
@@ -59,7 +59,7 @@ public class WorkspaceSettings: Codable, Equatable, Writable {
     public init(buildSystem: BuildSystem = .new,
                 derivedDataLocationStyle: DerivedDataLocationStyle? = nil,
                 derivedDataCustomLocation: String? = nil,
-                autoCreateSchemes: Bool = true)
+                autoCreateSchemes: Bool? = nil)
     {
         self.buildSystem = buildSystem
         self.derivedDataLocationStyle = derivedDataLocationStyle
@@ -88,7 +88,7 @@ public class WorkspaceSettings: Codable, Equatable, Writable {
             derivedDataLocationStyle = .default
         }
         derivedDataCustomLocation = try container.decodeIfPresent(.derivedDataCustomLocation)
-        autoCreateSchemes = try container.decode(.autoCreateSchemes)
+        autoCreateSchemes = try container.decodeIfPresent(.autoCreateSchemes)
     }
 
     /// Encodes the settings into the given encoder.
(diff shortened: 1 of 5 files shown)


Sources/XcodeProj/Project/WorkspaceSettings.swift
```diff
@@ -15,29 +15,29 @@ public class WorkspaceSettings: Codable, Equatable, Writable {
...
     /// Workspace DerivedData directory.
     public var derivedDataLocationStyle: DerivedDataLocationStyle?
-    
+
     /// Path to workspace DerivedData directory.
     public var derivedDataCustomLocation: String?
 
     /// When true, Xcode auto-creates schemes in the project.
-    public var autoCreateSchemes: Bool?
+    public var autoCreateSchemes: Bool
```
kwridan, line 40: This is sadly a breaking change 😞

kwridan (commented):
Thanks for submitting this @jakeatoms !

> This PR exposes the initializer of WorkspaceSettings, so we can use it in Tuist. 

👍 

>  It also makes the autoCreateSchemes a non-optional boolean.

This is sadly a breaking change 😞 - I'm curious as to the motivation for this change?

Within `XcodeProj` `nil` is a valid value and implies not writing the corresponding attributes to the underlying files and letting Xcode pick an appropriate default.

Taking a quick look at the `.xcsettings` files created by Xcode, the `IDEWorkspaceSharedSettings_AutocreateContextsIfNeeded` property is indeed optional, as Xcode will not include this by default. As such `XcodeProj` needs to be able to handle cases where it's missing and also ensure it can omit writing this property too. 

Thinking ahead for the changes in Tuist, we'll need to make sure to only create the `.xcsettings` file and update the corresponding properties when needed _(i.e. when we need to explicitly disable the default behaviour)_, otherwise we'd skip writing this file / or property all together.

jakeatoms:
> I'm curious as to the motivation for this change?

My goal was a better user API, but I was mostly focused on just the property I'm implementing in Tuist. The Descripter would be optional, so if the user didn't set it, it would not create the file. However, your point is well taken because if we implement additional properties in the future, we don't want add this property as well, if the user wants the default behavior. 

That leads me to think that my Tuist implementation of this should be with an enum that defaults to `default` and maps appropriately to a `nil` value here. Thanks for the feedback!

kwridan (approved):
Thanks @jakeatoms

pepicrft:
@all-contributors add @jakeatoms for code

pepicrft:
@jakeatoms I've released a new version of XcodeProj that includes your changes,[ 8.7.1 ](https://github.com/tuist/XcodeProj/releases/tag/8.7.1)

PR #667, opened 2022-02-23
Fix equality checking of dictionaries

brentleyjones:
### Short description 📝

Dictionary equality checks were only wrapping the lefthand side in `NSDictionary`, which was causing equality to return false unnecessarily.

### Solution 📦

Now the righthand side is wrapped in an `NSDictionary` as well.

### Implementation 👩‍💻👨‍💻

- [x] Updated `Equality.stencil`
- [x] Updated `Equality.generated.swift`

### DIFF
--- a/Templates/Equality.stencil
+++ b/Templates/Equality.stencil
@@ -6,7 +6,7 @@ extension {{ type.name }} {
     func isEqual(to rhs: {{ type.name }}) -> Bool {
         {% for variable in type.storedVariables %}
         {% if variable.typeName.dictionary %}
-        if !NSDictionary(dictionary: {{ variable.name}}{% if variable.typeName.isOptional %} ?? [:]{% endif %}).isEqual(to: rhs.{{ variable.name }}{% if variable.typeName.isOptional %} ?? [:]{% endif %}) { return false }
+        if !NSDictionary(dictionary: {{ variable.name}}{% if variable.typeName.isOptional %} ?? [:]{% endif %}).isEqual(NSDictionary(dictionary: rhs.{{ variable.name }}{% if variable.typeName.isOptional %} ?? [:]{% endif %})) { return false }
         {% elif variable|!annotated:"skipEquality" %}
         if {{ variable.name }} != rhs.{{ variable.name }} { return false }
         {% endif %}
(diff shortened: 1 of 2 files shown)


Templates/Equality.stencil
```diff
@@ -6,7 +6,7 @@ extension {{ type.name }} {
     func isEqual(to rhs: {{ type.name }}) -> Bool {
         {% for variable in type.storedVariables %}
         {% if variable.typeName.dictionary %}
-        if !NSDictionary(dictionary: {{ variable.name}}{% if variable.typeName.isOptional %} ?? [:]{% endif %}).isEqual(to: rhs.{{ variable.name }}{% if variable.typeName.isOptional %} ?? [:]{% endif %}) { return false }
+        if !NSDictionary(dictionary: {{ variable.name}}{% if variable.typeName.isOptional %} ?? [:]{% endif %}).isEqual(to: NSDictionary(dictionary: rhs.{{ variable.name }}{% if variable.typeName.isOptional %} ?? [:]{% endif %})) { return false }
```
fortmarek, line 9: I wonder if we have to use `NSDictionary` here 🤔 it's causing issues on Linux

brentleyjones: I dunno the solution then. Without wrapping in an `NSDictionary`, at least on macOS, these checks can return `false` when it should be `true`.

fortmarek: Could you try to drop the `to: ` then? There seems to be function for `NSDictionary`:
```swift
func isEqual(_ object: Any?) -> Bool
```
and
```swift
func isEqual(to otherDictionary: [AnyHashable: Any]) -> [Bool]
```

I think the former could work without compiler errors on both platforms.

brentleyjones: Done. Let's see if it fixes Linux.

brentleyjones: Looks like the tests are passing now!

fortmarek: great to hear, LGTM 🚀 thanks for the PR 👏

fortmarek (approved):

luispadron:
@all-contributors add @brentleyjones for code

Issue #666, opened 2022-02-15
PBXBuildFile objects are not strongly referenced when freshly created


stevelandeyasana:
## Context 🕵️‍♀️

I'm trying to write some code that syncs a subgroup of my project with a directory, and adds new files to the Sources phase of my build.

## What 🌱

I have some code that boils down to this:

```swift
let xcp = try! XcodeProj(path: Path(xcodeprojURL.path))
let project = xcp.pbxproj.projects.first(where: { $0.name == "MyApp" })!

// am_group here is a helper that returns
// `children.first(where: { $0.path == arg }) as! PBXGroup`
let group = project.mainGroup!.am_group("DataLayerFramework").am_group("_ Models").am_group("_ Generated")

let appTarget = xcp.pbxproj.nativeTargets.first(where: { $0.name == "MyApp" })!
let sourcesPhase = appTarget.buildPhases.first(where: { $0.buildPhase == .sources }) as! PBXSourcesBuildPhase

let sourceRoot = Path(xcodeprojURL.deletingLastPathComponent().path)
for filename in directory {
  // It's safe to call this without any guards because XcodeProj dedupes file references
  let file = try! group.addFile(
      at: Path(root + "/" + filename),
      sourceRoot: sourceRoot)
  if sourcesPhase.files?.contains(where: { $0.file?.path == filename }) != true {
      print("Adding \(filename).swift to the build phase")
      let buildFile = PBXBuildFile(file: file, product: nil, settings: [:])
      sourcesPhase.files?.append(buildFile)
  }
}

try! xcp.write(path: Path(xcodeprojPath.path))
```

When I call `xcp.write(...)`, I end up with this in my Sources build phase file list instead of a good references to my file:

```
				"TEMP_D2CF0395-3523-415D-B305-03727A331BCE" /* (null) in Sources */,
```

Based on my reading of PBXObjectReference and ReferenceGenerator, my guess is the file UUIDs aren't being fixed before the build UUIDs are fixed. But I'm not 100% sure of that.

## Proposal 🎉

XcodeProj should fix my references correctly. :-)

stevelandeyasana:
Bumping this, I'd love to fix this on my own but I really don't understand what's going on.

fortmarek:
Hey @stevelandeyasana 👋 

I can try looking into this but would it be possible if you first wrote a unit / integration test for this in the codebase? That would be super helpful 🙏

stevelandeyasana:
[...]

PR #668, opened 2022-03-16
Integration test to demonstrate missing-build-phase-file behavior

stevelandeyasana:

### Short description 📝

Demonstrate the issue I reported in #666

### Solution 📦

You asked for a test demonstrating the issue, so here it is. :-) The "solution" might actually be to update the docs.

### Implementation 👩‍💻👨‍💻

- [x] Replicated buggy code that was previously integrated into another codebase

### DIFF
--- a/Tests/XcodeProjTests/Objects/Project/PBXProjIntegrationTests.swift
+++ b/Tests/XcodeProjTests/Objects/Project/PBXProjIntegrationTests.swift
@@ -14,6 +14,28 @@ final class PBXProjIntegrationTests: XCTestCase {
         }
     }
 
+    func test_uuids_after_adding_to_build_phase() throws {
+        let fixturePath = self.fixturePath().parent().parent()
+        print(fixturePath)
+        let xcp = try XcodeProj(path: self.fixturePath().parent())
+        let project = xcp.pbxproj.projects.first(where: { $0.name == "Project" })!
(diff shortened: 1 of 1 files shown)


Tests/XcodeProjTests/Objects/Project/PBXProjIntegrationTests.swift
```diff
@@ -14,6 +14,28 @@ final class PBXProjIntegrationTests: XCTestCase {
...
+        let iosGroup = project.mainGroup.children.first(where: { $0.path == "iOS" }) as! PBXGroup
+        let appTarget = xcp.pbxproj.nativeTargets.first(where: { $0.name == "iOS" })!
+        let sourcesPhase = appTarget.buildPhases.first(where: { $0.buildPhase == .sources }) as! PBXSourcesBuildPhase
+
+        for filename in ["FileNotInProject.swift"] {
+            let newFilePath = Path(components: fixturePath.components + ["iOS", filename])
+            let file = try iosGroup.addFile(at: newFilePath, sourceRoot: fixturePath)
+            let buildFile = PBXBuildFile(file: file, product: nil, settings: [:])
+            sourcesPhase.files?.append(buildFile)
+        }
```
stevelandeyasana, line 31: When the loop body exits, `buildFile` and `file` are deallocated, so the weak references pointing to them are set to nil. The "fix" would be to add them to a list.

It was not at all clear to me reading the docs that I need to maintain strong references to objects I create before writing output.

Issue #670, opened 2022-03-30
PBXGroup file(named:) API assumes filenames (only) are unique in group


itsthejb:
## Context 🕵️‍♀️

`PBXGroup` contain an array of files. As far as I'm aware, Xcode places very little restriction on what can be placed in a group, and it's up to the user to add files sensibly...

## What 🌱

[`PBXGroup.file(named:)`](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Objects/Files/PBXGroup.swift#L119-L127) allows finding a file within a group, but using only the filename as the search criteria. This means that if, for example, the group contained two files, with the same name, but different locations, then then API would just return the first:

```
group
- 1: fileA (path: `foo/bar/fileA`)
- 2: fileA (path: `bar/fileA`)
```

This API would only ever return `1`, even though it's perfectly possible to have duplicate filenames in a group

## Proposal 🎉

Add an API that searches instead by the path to files. This correctly identifies files uniquely, and would return the correct file:

```
group
- 1: fileA (path: `foo/bar/fileA`)
- 2: fileA (path: `bar/fileA`)
```

- "file with name `fileA` and path `foo/bar/fileA`" returns `1`
- "file with name `fileA` and path `bar/fileA`" returns `2`
- "file with name `fileA` and path `.`" returns `nil`

PR #671, opened 2022-03-30
Add API to search for files in a group with path. Handles non-unique filenames in groups

itsthejb:
Resolves https://github.com/tuist/xcodeproj/issues/670

### Short description 📝

- Add path based accessor for files in a group

### Solution 📦

- Simply generate a full path, using the available source root, so it's possible to filter on a complete abs/relative path, rather than just assuming that the filename is enough

### Implementation 👩‍💻👨‍💻

- [x] Step 1: Add unit test
- [x] Step 2: Implement file find method
- [x] Step 3: Deprecate the name method


Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -120,12 +120,26 @@ public extension PBXGroup {
     ///
     /// - Parameter name: file name.
     /// - Returns: file with the given name contained in the given parent group.
+    @available(*, deprecated, message: "Please use file(with:relativeTo:). This method assumes all names in the group are unique")
```
itsthejb, line 123: Might be going a little far! But arguably the functionality is too broad, and unclear to the user

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -120,12 +120,26 @@ public extension PBXGroup {
...
 
+    /// Returns the file in the group with the given path, relative to a base path.
+    /// Will use the root path to generate normalized absolute paths for the search path
+    /// and also the file paths in the group.
+    ///
+    /// - Parameter path: a file path.
+    /// - Parameter basePath: a base path to search from.
+    /// - Returns: file with the given absolute path contained in the given parent group.
+    func file(with path: Path, relativeTo basePath: Path) -> PBXFileReference? {
+        let normalized = path.normalized(relativeTo: basePath)
```
itsthejb, line 138: 1: Use the `basePath` to generate a complete path to `path`; normalization takes care of various path traversal possibilities

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -120,12 +120,26 @@ public extension PBXGroup {
...
+        return children
+            .first { $0.normalizedPath(relativeTo: basePath) == normalized } as? PBXFileReference
```
itsthejb, line 140: 2: Do likewise for each file in the group; if the normalized result matches, then it _is at the same location on the disk_

Sources/XcodeProj/Objects/Files/PBXGroup.swift
kwridan: Perhaps updating the doc comments here may suffice? It can clarify the usage and point to the other method for cases where searching by path is needed.

Tests/XcodeProjTests/Objects/Files/PBXGroupTests.swift
```diff
@@ -199,6 +199,40 @@ final class PBXGroupTests: XCTestCase {
...
+            try XCTUnwrap(filesToRefs[file1])
+        )
+        XCTAssertEqual(
+            group.file(with: "file", relativeTo: rootPath + "subdir"),
+            try XCTUnwrap(filesToRefs[file2])
+        )
+        XCTAssertEqual(
+            group.file(with: "subdir/file", relativeTo: rootPath),
+            try XCTUnwrap(filesToRefs[file2])
+        )
```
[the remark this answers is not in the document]
kwridan: awesome thanks for adding this!

One aspect to consider is, typically the paths dealt with within project files are relative rather absolute paths either to source root, or to their parent group _(this is so projects are portable / can work on different machines / directories etc...)_

It's worth exploring the behaviour when that is the case.

kwridan (commented):
thanks for taking the time to contribute this @itsthejb

Sources/XcodeProj/Objects/Files/PBXGroup.swift
itsthejb: Sounds good. Let me take care of that right away

Tests/XcodeProjTests/Objects/Files/PBXGroupTests.swift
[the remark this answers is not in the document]
itsthejb: Yes indeed; I verified this in our parent project. The paths just need to share a common root, that doesn't need to be `/`, and often it would be the root of the project (i.e. `.`). I'll adjust this test to include that

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -117,6 +117,8 @@ public extension PBXGroup {
     }
 
     /// Returns the file in the group with the given name.
+    /// - Note: Performs a linear search based on the file's name _only_. To search for a specific file
+    /// based on file-system location, use the `file(with path:)` method.
```
itsthejb, line 121: Note on the limitations, instead of deprecation

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -126,6 +128,27 @@ public extension PBXGroup {
...
+        func absolutePath(for fileRef: PBXFileElement) -> Path? {
+            try? fileRef.fullPath(sourceRoot: ".")
+        }
+        
+        guard let groupAbsPath = absolutePath(for: self) else { return nil }
+        let fileNormalizedAbsolutePath = (groupAbsPath + path).normalize()
+        return children.first {
+            guard let candidateAbsPath = absolutePath(for: $0) else { return false }
+            return candidateAbsPath.normalize() == fileNormalizedAbsolutePath
+        } as? PBXFileReference
```
itsthejb, line 149: The algorithm has now quite importantly changed, and is much improved, but requires some feedback:

- I realised that the functionality already exists (`fullPath(sourceRoot:)`) to generate an absolute path for all files involved. Even passing `.` to represent "from current dir"
- Therefore a relative directory is not required by the caller: simply convert the argument path, and all group paths to normalized absolute paths, and search for that

This way, it's guaranteed that any kind of path (relative, absolute) works as input, since absolute paths are always generated internally as the file system "source of truth"

Tests/XcodeProjTests/Objects/Files/PBXGroupTests.swift
```diff
@@ -199,6 +199,77 @@ final class PBXGroupTests: XCTestCase {
...
+            hasReference: fileReferenceFor[file1Path]
+        )
+        assert(
+            filePath: groupDir + "foo/.." + subDirectoryName + fileName,
+            hasReference: fileReferenceFor[file2Path]
+        )
+        assert(
+            filePath: Path(UUID().uuidString),
+            hasReference: nil
+        )
```
itsthejb, line 270: This got a little involved, but the basic idea is to recreate Xcode-like scenarios:

- You have a `projectDir` which can be located anywhere on disk
- You have a `group`, which has some path relative to parents, or absolute, etc. (as expected in Xcode)
- The `group` contains files that have paths that likewise relate to the group, or absolute etc., might have messy traversals to parent directories
- Specifics of these are irrelevant, since internally we generate those normalized absolute paths

itsthejb:
Ok! This has now actually changed somewhat, and much improved (I hope!). The only real question is whether or not `fullPath(sourceRoot:)` might have some unexpected results in some situations; it seems not, though...

Tests/XcodeProjTests/Objects/Files/PBXGroupTests.swift
```diff
@@ -199,6 +199,77 @@ final class PBXGroupTests: XCTestCase {
...
+        assert(
+            filePath: file2Path,
+            hasReference: fileReferenceFor[file2Path]
+        )
+        assert(
+            filePath: groupDir + subDirectoryName + ".." + fileName,
+            hasReference: fileReferenceFor[file1Path]
+        )
+        assert(
+            filePath: groupDir + "foo/.." + subDirectoryName + fileName,
```
itsthejb, line 264: Testing traversals to parents

Sources/XcodeProj/Objects/Files/PBXGroup.swift
damirdavletov: I think using "." for the sourceRoot is wrong and will only work correctly for cases when the working dir is the same as project location. In other cases it will work in unpredictable way e.g. if you will have absolute file reference in project and a relative one which will resolve in incorrect way they will not be comparable. I'd advise to pass a real sourceRoot to the method.

damirdavletov (changes requested):

Tests/XcodeProjTests/Objects/Files/PBXGroupPathSearchTests.swift
```diff
@@ -0,0 +1,120 @@
+import Foundation
+import PathKit
+import XcodeProj
+import XCTest
+
+final class PBXGroupPathSearchTests: XCTestCase {
```
itsthejb, line 6: Split into own file. Could combine again, but it did get a little messy

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -126,6 +128,28 @@ public extension PBXGroup {
...
+    /// Internally, identity is calculated based on normalized absolute paths for all files involved; i.e. actual
+    /// location in the file system.
+    ///
+    /// - Parameters:
+    ///   - path: a file path to search for.
+    ///   - sourceRoot: path to project's source root.
+    /// - Returns: an existing reference to that file, or `nil` if not found
+    func file(with path: Path, sourceRoot: Path) -> PBXFileReference? {
+        func absolutePath(for fileRef: PBXFileElement) -> Path? {
+            try? fileRef.fullPath(sourceRoot: sourceRoot)
```
itsthejb, line 142: Going with the convention of the other methods, then, and "just pass in `$(SRCROOT)` here and everything should work". Tests seem to show it does

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -187,17 +202,11 @@ public extension PBXGroup {
...
-            let fileRefLastPathComponent = fileRefPath.split(separator: "/").last!
-            if filePath.lastComponent == fileRefLastPathComponent {
-                return try filePath == $0.value.fullPath(sourceRoot: sourceRoot)
-            }
-            return false
-        }) {
+        if override, let existingFileReference = try self.existingFileReference(
+            from: projectObjects,
+            with: filePath,
+            sourceRoot: sourceRoot
```
itsthejb, line 208: Having spent some time going round and round on this, trying to create comprehensive tests, whilst still being slightly confused by how `PBXSourceTree`s actually work, I finally realised that the functionality needed is actually already *here*. It's only necessary to extract is and create a public driver method @kwridan @damirdavletov.

Therefore I decided not to assume too much about my knowledge of this lib, or PBXProj in general, and simply assume that this functionality works and expose it as we require. Seems much simpler. I can't guarantee it previously worked, but it should not break anything either...?

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -233,3 +242,24 @@ public extension PBXGroup {
         return fileReference
     }
 }
+
+private extension PBXGroup {
+
+    func existingFileReference(
+        from objects: PBXObjects? = nil,
```
itsthejb, line 249: Since the original already accessed a `PBXObjects`, I thought to at least keep that as an optional arg. It seems to be just a constant reference, but playing it safe here: https://github.com/tuist/XcodeProj/blob/6943d5c5f9085d070dbde3c669144e30d41ec635/Sources/XcodeProj/Objects/Project/PBXObject.swift#L75-L81

Sources/XcodeProj/Objects/Files/PBXGroup.swift
```diff
@@ -233,3 +242,24 @@ public extension PBXGroup {
...
+            // Optimization: compare lastComponent before fullPath compare
+            guard let fileRefPath = $0.value.path else {
+                return try filePath == $0.value.fullPath(sourceRoot: sourceRoot)
+            }
+            let fileRefLastPathComponent = fileRefPath.split(separator: "/").last!
+            if filePath.lastComponent == fileRefLastPathComponent {
+                return try filePath == $0.value.fullPath(sourceRoot: sourceRoot)
+            }
+            return false
+        })
```
itsthejb, line 263: Moved from https://github.com/tuist/XcodeProj/blob/6943d5c5f9085d070dbde3c669144e30d41ec635/Sources/XcodeProj/Objects/Files/PBXGroup.swift#L190-L200

damirdavletov (approved):

kwridan (commented):
Thanks for the follow ups and updates @itsthejb 👍 

Having dug in a little further the exposed method does introduce some API awkwardness 🤔 - arguably existing in the `addFile` method.

Taking the following simple demo project

Screenshot 2022-03-31 at 20 13 59


```swift
  func testFileWithPath() throws {
    // Given
    let xodeProj = try XcodeProj(path: demoPath)
    let mainGroup = try XCTUnwrap(try xodeProj.pbxproj.rootGroup())
    let sourcesGroup = try XCTUnwrap(mainGroup.children.first(where: { $0.path == "Sources" }) as? PBXGroup)

    // When
    //   Expected 👍
    let readmeReference1 = mainGroup.file(with: "README.md", sourceRoot: ".")
    let helloSwift1 = sourcesGroup.file(with: "Sources/Hello.swift", sourceRoot: ".")
    let helloSwift2 = sourcesGroup.file(with: demoPathSourceRoot + "Sources/Hello.swift", sourceRoot: demoPathSourceRoot)

    //   Possibly expected 🤔
    let helloSwift3 = mainGroup.file(with: "Sources/Hello.swift", sourceRoot: ".")

    //   Not expected ... 🙈
    let readmeReference2 = sourcesGroup.file(with: "README.md", sourceRoot: ".")
    let readmeReference3 = sourcesGroup.file(with: demoPathSourceRoot + "README.md", sourceRoot: demoPathSourceRoot)

    // Then
    XCTAssertNotNil(readmeReference1)
    XCTAssertNotNil(helloSwift1)
    XCTAssertNotNil(helloSwift2)
    XCTAssertNotNil(helloSwift3)
    XCTAssertNil(readmeReference2) // ❌ Fails
    XCTAssertNil(readmeReference3) // ❌ Fails
}
```

The last few cases seem a bit odd, where we are asking the `Sources` group for `README.md` (which shouldn't exist within it) - yet it returns the reference from its parent. Perhaps I am not using the API correctly?

I don't have great any suggestions yet sadly - will have another look through the code to get a better understanding.

cc: @tuist/core incase any of you have further insights.

Thanks again @itsthejb

itsthejb:
@kwridan Thanks for investigating further. To be honest, going deeper into this in order to validate it worked as expected for all cases was beginning to stretch my understanding of `PBXSourceTree`s, which I've only semi-understanded over the years working in Xcode. I was also trying to set up a whole suite of unit tests (which are still in previous commits), but started chasing my tail and getting a headache... 🤯 

So, when I realised there was already an implementation in there, I went with the approach of exposing that, rather than assuming DIY could be any better (probably worse)

Actually our original use case was avoiding duplicate files in groups, and this whole exercise has taught me that `addFile(...)` already has some form of this...

PR #677, opened 2022-04-09
Add `addDependency()` helper method to `PBXAggregateTarget`

brentleyjones:
### Short description 📝

The `addDependency()` helper method is useful for more than just `PBXNativeTarget`, such as for `PBXAggregateTarget`.

### Solution 📦

Add a copy of the extension on `PBXNativeTarget` to `PBXAggregateTarget`.

We could maybe instead move the extension to `PBXTarget`? I'm not sure if it's valid there.

### DIFF
--- a/Tests/XcodeProjTests/Objects/Targets/PBXAggregateTargetTests.swift
+++ b/Tests/XcodeProjTests/Objects/Targets/PBXAggregateTargetTests.swift
@@ -27,4 +27,34 @@ final class PBXAggregateTargetTests: XCTestCase {
             "name": "name",
         ]
     }
+
+    func test_addDependency() throws {
+        let objects = PBXObjects(objects: [])
+
+        let configurationList = XCConfigurationList.fixture()
+        let mainGroup = PBXGroup.fixture()
+        objects.add(object: configurationList)
+        objects.add(object: mainGroup)
+
+        let project = PBXProject(name: "Project",
+                                 buildConfigurationList: configurationList,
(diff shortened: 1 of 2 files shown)


Tests/XcodeProjTests/Objects/Targets/PBXAggregateTargetTests.swift
```diff
@@ -27,4 +27,34 @@ final class PBXAggregateTargetTests: XCTestCase {
...
+        objects.add(object: configurationList)
+        objects.add(object: mainGroup)
+
+        let project = PBXProject(name: "Project",
+                                 buildConfigurationList: configurationList,
+                                 compatibilityVersion: "0",
+                                 mainGroup: mainGroup)
+
+        objects.add(object: project)
+        let target = PBXNativeTarget(name: "Target", buildConfigurationList: nil)
```
kwridan, line 45: Should the test not create an `PBXAggregateTarget` seeing that's a the subject under test? I appreciate this may have been copied from the existing native target test 😅

kwridan (commented):
Thanks @brentleyjones

Tests/XcodeProjTests/Objects/Targets/PBXAggregateTargetTests.swift
brentleyjones: Ahh, yep.

kwridan (approved):
Thanks for the updates 👍

fortmarek (approved):

PR #679, opened 2022-04-27
Fix syntax error in docs

maxwellE:
use forEach over each


kwridan (approved):
Thanks for contributing this @maxwellE

fortmarek (approved):

PR #682, opened 2022-04-29
Allow for initializing a PBXProj via a direct path

maxwellE:
### Short description 📝
Allows for a pbxproj path to be initiated with a direct path, also
moves some related code to the PBXProj class

### Solution 📦
Simply moves around the code that already exists and opens some ACLs to allow callers of the library to use this code

### Implementation 👩‍💻👨‍💻
N/A

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProj.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProj.swift
@@ -52,6 +52,24 @@ public final class PBXProj: Decodable {
         }
     }
 
+    /// Initializes the project with a path to the pbxproj file.
+    ///
+    /// - Parameters:
+    ///   - path: Path to a pbxproj file.
+    public convenience init(path: Path) throws {
+        let pbxproject: PBXProj = try PBXProj.createPBXProj(path: path)
+        var objects: [PBXObject] = []
+        pbxproject.objects.forEach { object in
+            objects.append(object)
+        }
+        self.init(
+            rootObject: pbxproject.rootObject,
+            objectVersion: pbxproject.objectVersion,
+            archiveVersion: pbxproject.archiveVersion,
+            classes: pbxproject.classes,
+            objects: objects)
+    }
+
     // MARK: - Decodable
 
     fileprivate enum CodingKeys: String, CodingKey {
@@ -95,7 +113,7 @@ public final class PBXProj: Decodable {
 
     // MARK: Static Methods
 
-    public static func createPBXProj(path: Path) throws -> PBXProj {
+    private static func createPBXProj(path: Path) throws -> PBXProj {
         let (pbxProjData, pbxProjDictionary) = try PBXProj.readPBXProj(path: path)
         let context = ProjectDecodingContext(
             pbxProjValueReader: { key in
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Objects/Project/PBXProj.swift
```diff
@@ -92,6 +92,35 @@ public final class PBXProj: Decodable {
 
         try rootGroup()?.assignParentToChildren()
     }
+
+    // MARK: Static Methods
+
+    public static func createPBXProj(path: Path) throws -> PBXProj {
```
fortmarek, line 98: this could possibly be another `init` method with different parameters but then you would have bunch of logic in the initializer, so I suppose we can keep it as-is.

fortmarek (approved):
I don't see why not make this reusable but would wait for @kwridan review before merging 👍

Sources/XcodeProj/Objects/Project/PBXProj.swift
maxwellE: I am up for either option, whichever you prefer

Sources/XcodeProj/Objects/Project/PBXProj.swift
```diff
@@ -52,6 +52,24 @@ public final class PBXProj: Decodable {
...
+    /// Initializes the project with a path to the pbxproj file.
+    ///
+    /// - Parameters:
+    ///   - path: Path to a pbxproj file.
+    public convenience init(path: Path) throws {
+        let pbxproject: PBXProj = try PBXProj.createPBXProj(path: path)
+        var objects: [PBXObject] = []
+        pbxproject.objects.forEach { object in
+            objects.append(object)
+        }
```
kwridan, line 64: I wonder if this has any performance impact on large projects 🤔

Looking at the code it seems the array `[PBXObject]` is only transiently needed to pass to the public initialiser, but later gets re-assembled as `PBXObjects` _(which internally loops over the objects to process them)_ - this seems like an unnecessary roundtrip 😅

We may be able to resolve this by:

- making this a regular initialiser and directly assign the variables from loaded `pbxproj`
- creating another private initialiser that can be provided with `PBXObjects` directly to avoid needing an extra hop of making the `[PBXObject]`

Sources/XcodeProj/Objects/Project/PBXProj.swift
```diff
@@ -52,6 +52,24 @@ public final class PBXProj: Decodable {
         }
     }
 
+    /// Initializes the project with a path to the pbxproj file.
+    ///
+    /// - Parameters:
+    ///   - path: Path to a pbxproj file.
+    public convenience init(path: Path) throws {
+        let pbxproject: PBXProj = try PBXProj.createPBXProj(path: path)
```
kwridan, line 60: nit: worth naming this as `pbxproj` to avoid confusion - there is another [`PBXProject`](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Objects/Project/PBXProject.swift) type in the code base which is slightly different 🤐

kwridan (commented):
Thanks for the updates @maxwellE

kwridan (approved):

Issue #683, opened 2022-05-26
invalidateUUIDs() + write() loses references


timothycosta:
## Context 🕵️‍♀️
Trying to make Xcode project deterministic for easy diffing the same as [this issue](https://github.com/tuist/XcodeProj/issues/265).

```
let xcodeproj = try XcodeProj(path: projectPath)
xcodeproj.pbxproj.invalidateUUIDs()
let settings = PBXOutputSettings(projFileListOrder: .byFilename)
try xcodeproj.write(path: projectPath, outputSettings: settings)
```

## What 🌱
If I run the above code on [this project](https://github.com/tuist/XcodeProj/files/8775979/lost.xcodeproj.zip) twice, the first time succeeds. The second time I get an error:
`Error raised at top level: PBXObject with reference "2BAA80704959AADB23CA8DB5F222F967" not found.`

The attached project file has two identical (except for platform) frameworks `Localization` and `Localization Watch`. It seems like XcodeProj thinks they're the same thing, and removes one.

## Proposal 🎉
XcodeProj should be smart enough to realize they're different and not delete the reference.

fortmarek:
Hey @timothycosta 👋 

I agree with you, we'd appreciate any PR for this 🙌

PR #685, opened 2022-06-02
Consider UUIDs when sorting by name

timothycosta:
Resolves https://github.com/tuist/XcodeProj/issues/683

### Short description 📝
Ensure UUIDs are considered during sort by filename to ensure information is not lost when filenames are equal.

### Solution 📦
Consider both the PBXBuildFile UUID and PBXFileElement UUID.

- [X] Consider all UUIDs when filenames are equal
- [X] Update documentation
- [X] Rename sort methods differently so they can be distinguished at the call site


kwridan (commented):
Thanks for contributing this @timothycosta!

Mind extending `PBXOutputSettingsTests` to add a few new cases that help replicate the issue being solved? It will help clarify this scenario and would prevent future regressions in this area.

timothycosta:
@kwridan I've added tests for that, but there are now some other failing tests due to adding files to `iOS/Project.xcodeproj`. They seem "magic numbery". Can you advise?


Screen Shot 2022-06-10 at 11 18 25 AM

Screen Shot 2022-06-10 at 11 20 37 AM

kwridan (commented):
Thanks for the updates @timothycosta 

ouch, you've hit an issue with the other tests after updated an existing fixture they relied on _(e.g. the number of build phases, file references, etc... in said fixture)_. 

Would it be too much trouble to add a separate fixture and test case that replicates this scenario? this should keep it much smaller and hopefully less brittle as it won't impact the other tests.

PR #686, opened 2022-06-02
Ensure correct `LaunchAction`  scheme order

maxwellE:
### Short description 📝

Adds logic and tests to ensure that the XML
ordering for `LaunchAction` schemes matches what
xcode expects


kwridan (approved):
Thanks @maxwellE

maxwellE:
@pepicrft Can I get a review on this? Thanks

danyf90 (approved):

PR #689, opened 2022-06-13
Fix TestAction scheme attribute ordering

maxwellE:
### Short description 📝
`customLLDBInitFile` is not in the correct order for `TestAction` scheme


fortmarek (approved):

kwridan (approved):
Thanks @maxwellE

Issue #687, opened 2022-06-07
Add support for `ExtensionKit` extension product types (new in Xcode 14)


kwridan:
## Context 🕵️‍♀️

Xcode 14 has introduced a few new frameworks and associated extension product types:

- [`AppIntents`](https://developer.apple.com/documentation/AppIntents)
- [`BackgroundAssets`](https://developer.apple.com/documentation/backgroundassets/)

![app-intents](https://user-images.githubusercontent.com/11914919/172392045-56e7577b-eeda-4733-8b77-ebd81a71efca.png)

![background-assets](https://user-images.githubusercontent.com/11914919/173757732-2f3a9482-82bd-4a48-9174-f987f607fcc0.png)

## What 🌱

- The new extensions have a new product type `com.apple.product-type.extensionkit-extension` which not currently supported by XcodeProj

```plist
		FB2930F2284F86D7004FAF29 /* AppIntentsExtension */ = {
			isa = PBXNativeTarget;
			buildConfigurationList = FB2930FB284F86D7004FAF29 /* Build configuration list for PBXNativeTarget "AppIntentsExtension" */;
			buildPhases = (
				FB2930EF284F86D7004FAF29 /* Sources */,
				FB2930F0284F86D7004FAF29 /* Frameworks */,
				FB2930F1284F86D7004FAF29 /* Resources */,
			);
			buildRules = (
			);
			dependencies = (
			);
			name = AppIntentsExtension;
			productName = AppIntentsExtension;
			productReference = FB2930F3284F86D7004FAF29 /* AppIntentsExtension.appex */;
			productType = "com.apple.product-type.extensionkit-extension";
		};
```

```plist
		FBE4DE842859B1EC00C6A189 /* BackgroundExt */ = {
			isa = PBXNativeTarget;
			buildConfigurationList = FBE4DE8E2859B1EC00C6A189 /* Build configuration list for PBXNativeTarget "BackgroundExt" */;
			buildPhases = (
				FBE4DE812859B1EC00C6A189 /* Sources */,
				FBE4DE822859B1EC00C6A189 /* Frameworks */,
				FBE4DE832859B1EC00C6A189 /* Resources */,
			);
			buildRules = (
			);
			dependencies = (
			);
			name = BackgroundExt;
			productName = BackgroundExt;
			productReference = FBE4DE852859B1EC00C6A189 /* BackgroundExt.appex */;
			productType = "com.apple.product-type.extensionkit-extension";
		};
```

[...]

PR #691, opened 2022-06-25
Add `.extensionKitExtension` as the new `PBXProductType`

mtj0928:
Resolves: https://github.com/tuist/XcodeProj/issues/687

### Short description 📝
From Xcode14 beta 1, `com.apple.product-type.extensionkit-extension` was introduced as the new productType.
The new productType is used for a new app extension like AppIntents Extension.

### Solution 📦
Added `.extensionKitExtension` as the new case of `PBXProductType` in this PR, to support the new productType.

### Implementation 👩‍💻👨‍💻
- [x] Added `.extensionKitExtension` as the new case of `PBXProductType`
- [x] Added `fileExtension` of `.extensionKitExtension `
- [x] Added a new UnitTest for `.extensionKitExtension `

### DIFF
--- a/Sources/XcodeProj/Objects/Targets/PBXProductType.swift
+++ b/Sources/XcodeProj/Objects/Targets/PBXProductType.swift
@@ -12,7 +12,7 @@ public enum PBXProductType: String, Decodable {
     case unitTestBundle = "com.apple.product-type.bundle.unit-test"
     case uiTestBundle = "com.apple.product-type.bundle.ui-testing"
     case appExtension = "com.apple.product-type.app-extension"
-    case extensionKit = "com.apple.product-type.extensionkit-extension"
+    case extensionKitExtension = "com.apple.product-type.extensionkit-extension"
     case commandLineTool = "com.apple.product-type.tool"
     case watchApp = "com.apple.product-type.application.watchapp"
     case watch2App = "com.apple.product-type.application.watchapp2"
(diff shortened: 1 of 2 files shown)


kwridan (commented):
Thanks for contributing this @mtj0928 

For consistency other product types (e.g. `appExtension`), I was thinking perhaps `.extensionKitExtension`? I appreciate the repeat of "extension" word is a bit odd 😅

kwridan:
@all-contributors add @mtj0928 for code

mtj0928:
@kwridan 
Thank you for your comment!!

I actually considered the name at first , but I also felt it was a bit odd.
Would you like me to change the name `extensionKitExtension`?

freddi-kit:
I vote to rename as `extensionKitExtension` to make naming consistently to others (appExtension, tvExtension)

kwridan (approved):
Thinking through it some more, the concern I had was around the possibility of there being `extensionKit<SomethingElse>` by the structure of the product current product ID.

That said, we can start out with `.extensionKit` which is less odd, should any future products be added that extend this (e.g. `.extensionKitHost`) we could deprecate `.extensionKit` and re-introduce the specialised `.extensionKitExtension`.

So let's proceed with `.extensionKit`.

freddi-kit (approved):

kwridan:
cc: @tuist/core - any preference to the naming convention - I could see either option working, consistency vs oddness 😅

fortmarek (approved):

kwridan (changes requested):
For visibility marking as request changes. thanks again!

mtj0928:
Thank you for your discussion, I renamed the new case and pushed the changes😄

kwridan (approved):
Thanks @mtj0928

kwridan:
Let's wait till #694 is merged then we can do a release with a few of the Xcode 14 related items within it.

PR #698, opened 2022-07-03
Fix ordering of Scheme toplevel attrs

maxwellE:
Fix for incorrect ordering of `Scheme` attrs when
using `wasCreatedForAppExtension` value

### DIFF
--- a/Tests/XcodeProjTests/Extensions/AEXML+XcodeFormatTests.swift
+++ b/Tests/XcodeProjTests/Extensions/AEXML+XcodeFormatTests.swift
@@ -161,6 +161,8 @@ class AEXML_XcodeFormatTests: XCTestCase {
                 "wasCreatedForAppExtension": "YES",
                 "LastUpgradeVersion": "1320",
                 "version": "1.7"
+            ]
+        )
     }
 
     func test_RemoteRunnable() {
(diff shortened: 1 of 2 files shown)


Tests/XcodeProjTests/Extensions/AEXML+XcodeFormatTests.swift
```diff
@@ -121,6 +131,30 @@ class AEXML_XcodeFormatTests: XCTestCase {
...
+            childName: "Scheme",
+            attributes: [
+                "LastUpgradeVersion": "1320",
+                "wasCreatedForAppExtension": "YES",
+                "version": "1.7"
+            ]
+        )
+    }
+
+    func test_Scheme_attributes_sorted_when_original_unsorted() {
```
kwridan, line 146: minor: this test may be redundant, the order in which dictionary elements are specified to my knowledge is opaque to the underlying types, the dictionaries are equal between this test and `test_Scheme_attributes_sorted_when_original_sorted`, as such only one of those tests will suffice.

An interesting test could be the following (but may be overkill for this)

- Start with an unsorted raw XML (string) scheme
- Deserialize it (read it vis `XCScheme`)
- Serialize it back to XML
- Verify the correct sort order

kwridan (approved):
Thanks @maxwellE

luispadron (approved):

luispadron:
@maxwellE thanks for the PR. I've merged some of them and looks like that created a conflict here. Once that's resolved we can get this merged!

luispadron:
@all

PR #400, opened 2019-04-02
Add support for RemoteRunnable

pepibumur:
Resolves https://github.com/tuist/xcodeproj/issues/276

### Short description 📝
As @brentleyjones pointed out, extensions use a type of runnable that is not supported by xcodeproj, `RemoteRunnable`. This PR adds support for it.

### Implementation 👩‍💻👨‍💻
- [x] Add `Runnable` class that both `RemoteRunnable` and `BuildableProductRunnable` inherit from.
- [x] Update `LaunchAction` to parse the runnable into the right type.
- [x] Add tests.

PR #703, opened 2022-07-07
Use `Runnable` in `ProfileAction`

maxwellE:
This matches the behavior of `LaunchAction` by accepting the more abstract
`Runnable` type

### DIFF
--- a/Sources/XcodeProj/Scheme/XCScheme+ProfileAction.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+ProfileAction.swift
@@ -57,6 +57,41 @@ extension XCScheme {
             super.init(preActions, postActions)
         }
 
+        public convenience init(
+            buildableProductRunnable: Runnable?,
+            buildConfiguration: String,
+            preActions: [ExecutionAction] = [],
+            postActions: [ExecutionAction] = [],
+            macroExpansion: BuildableReference? = nil,
+            shouldUseLaunchSchemeArgsEnv: Bool = true,
+            savedToolIdentifier: String = "",
+            ignoresPersistentStateOnLaunch: Bool = false,
+            useCustomWorkingDirectory: Bool = false,
+            debugDocumentVersioning: Bool = true,
+            askForAppToLaunch: Bool? = nil,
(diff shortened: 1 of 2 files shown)


kwridan (commented):
Thanks @maxwellE ! 

I think this change looks reasonable, but would be helpful to reference the motivation behind this (e.g. was there a case that required this?)

Would we be able to offer an alternate initializer that uses `Runnable` while maintaining `buildableProductRunnable` to avoid making a breaking change?

brentleyjones:
Application extension schemes (like WidgetKit Extensions) need to use a `RemoteRunnable`, similar to what caused `LaunchAction` to change to the `Runnable` based initializer.

luispadron (approved):

maxwellE:
@kwridan [It looks like `LaunchAction` does not use `buildableProductRunnable`](https://github.com/maxwellE/XcodeProj/blob/108962c8055339933d647099b3f25c6f7d792409/Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift#L75) so in a sense we have already engaged a breaking API design by forgetting to update `ProfileAction` when doing that migration

kwridan:
> @kwridan [It looks like LaunchAction does not use buildableProductRunnable](https://github.com/maxwellE/XcodeProj/blob/108962c8055339933d647099b3f25c6f7d792409/Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift#L75) so in a sense we have already engaged a breaking API design by forgetting to update ProfileAction when doing that migration

Thanks @maxwellE, indeed this was a breaking change between and v6.7.0 [v7.0.0](https://github.com/tuist/XcodeProj/blob/main/CHANGELOG.md#700) via https://github.com/tuist/XcodeProj/pull/400

I'm not against this per se, but we would be good to avoid needing a major release just now if we're able to maintain compatibility with the existing method, if that isn't feasible or introduces complexity we can make a major release with this breaking change.

brentleyjones:
I think having a new base `init` for `Runnable` and having the old one call into it is a fine compromise.

kwridan (approved):
Thanks @maxwellE - I pushed a small addition to preserve the public property as well

PR #706, opened 2022-07-10
Add `DEAD_CODE_STRIPPING` default project setting

kwridan:
Relates to: https://github.com/tuist/tuist/discussions/4531

### Short description 📝

- As of Xcode 14, `DEAD_CODE_STRIPPING` is now a recommended setting for macOS or iOS with catalyst projects and targets

### Solution 📦

- Add `DEAD_CODE_STRIPPING` as a default project setting
- iOS already had `DEAD_CODE_STRIPPING` enabled as a default in prior Xcode versions, so explicitly setting shouldn't cause a change in behaviour


luispadron (approved):
Thanks @kwridan

fortmarek (approved):

PR #710, opened 2022-08-02
Fix warning compiling for macOS 13

jpsim:
### Short description 📝

There's a deprecation warning when building with Xcode 14 targeting macOS 13:

```
Sources/XcodeProj/Extensions/String+md5.swift:63:16: warning: 'init(bytesNoCopy:length:encoding:freeWhenDone:)' was deprecated in macOS 13: String does not support no-copy initialization
        return String(bytesNoCopy: ptr, length: hexLen, encoding: .utf8, freeWhenDone: true)!
               ^
```

This is because `String.init(bytesNoCopy:...)` was deprecated in macOS 13: https://developer.apple.com/documentation/swift/string/init(bytesnocopy:length:encoding:freewhendone:)

Further discussion here: https://forums.swift.org/t/does-string-bytesnocopy-copy-bytes/51643/11

### Solution 📦

Use `String.init(cString:)` instead.

### Implementation 👩‍💻👨‍💻

- [x] Use `String.init(cString:)` instead of `String.init(bytesNoCopy:...)`
- [x] Deallocate bytes from pointer after creating the string

### DIFF
--- a/Sources/XcodeProj/Extensions/String+md5.swift
+++ b/Sources/XcodeProj/Extensions/String+md5.swift
@@ -50,6 +50,7 @@ private extension DataProtocol {
     var hexString: String {
         let hexLen = self.count * 2
         let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: hexLen)
+        ptr.assign(repeating: 0, count: hexLen)
         defer { ptr.deallocate() }
         var offset = 0
 
@@ -60,8 +61,6 @@ private extension DataProtocol {
                 offset += 1
             }
         }
-
-        ptr[Int(offset * 2)] = 0 // Null terminator
 
         return String(cString: ptr)
     }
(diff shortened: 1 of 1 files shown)


luispadron (approved):
Thanks!

danyf90 (approved):

jpsim:
I made a few updates after realizing that the initial code wasn't safe. Should be good to go now.

jpsim:
CI didn't run on the latest commits of this PR.

Tests failed on `main`, but they pass for me locally when running `swift test` with Xcode 14b4.

kwridan:
Sorry folks had to revert this 😞 as it was causing some reference generation instability (tests caught this when run repeatedly)

jpsim:
I understand. I don't see how what I wrote is unsafe, but it clearly is.

PR #720, opened 2022-09-02
Sets `customWorkingDirectory` for schemes

maxwellE:
### Short description 📝
This attribute was missed in launch and profile
schemes


Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
```diff
@@ -42,6 +42,7 @@ extension XCScheme {
         public var launchStyle: Style
         public var askForAppToLaunch: Bool?
         public var pathRunnable: PathRunnable?
+        public var customWorkingDirectory: Path?
```
kwridan, line 45: Taking a look at the codebase for any serialised properties `String` has been used in the past whereas `Path` was only convenience within some of the APIs before bridging to `String` 

For example:

https://github.com/tuist/XcodeProj/blob/6943d5c5f9085d070dbde3c669144e30d41ec635/Sources/XcodeProj/Scheme/XCScheme%2BLaunchAction.swift#L70

https://github.com/tuist/XcodeProj/blob/6943d5c5f9085d070dbde3c669144e30d41ec635/Sources/XcodeProj/Objects/Files/PBXFileElement.swift#L12

https://github.com/tuist/XcodeProj/blob/6943d5c5f9085d070dbde3c669144e30d41ec635/Sources/XcodeProj/Objects/BuildPhase/PBXCopyFilesBuildPhase.swift#L22

If I'm not mistaken, Xcode may also accept variables in this field too as such it's not exclusively `Path`s :/ 

Screenshot 2022-09-02 at 07 55 19

Tests/XcodeProjTests/Scheme/XCSchemeTests.swift
```diff
@@ -655,6 +660,7 @@ final class XCSchemeIntegrationTests: XCTestCase {
         XCTAssertEqual(scheme.profileAction?.buildConfiguration, "Release")
         XCTAssertTrue(scheme.profileAction?.shouldUseLaunchSchemeArgsEnv == true)
         XCTAssertEqual(scheme.profileAction?.savedToolIdentifier, "")
+        XCTAssertNil(scheme.profileAction?.customWorkingDirectory)
```
kwridan, line 663: It would be great to update one of the fixtures to include a `customWorkingDirectory` such that one of the tests verifies serialising / deserialising it

kwridan (changes requested):
Thanks for submitting this @maxwellE 

Left a few minor comments

kwridan (approved):
Awesome, thanks @maxwellE

luispadron (approved):

PR #733, opened 2022-11-19
Build Tool Plugin Support

technocidal:
### Short description 📝
I'm an avid user of XcodeGen and would like to have Build Tool Plugin support. It's possible to implement this without having to change this project but it leads to temporary identifiers sticking around in the generated project.

More details can be found in the [issue](https://github.com/yonaskolb/XcodeGen/issues/1290) where I documented my research.

### Solution 📦
Adds a flag to XCSwiftPackageProductDependency to indicate that it should be treated as a build tool plugin.

### Implementation 👩‍💻👨‍💻
- Add flag to XCSwiftPackageProductDependency
- Prefix the product name with "plugin:" as this indicates to Xcode that this should be treated as a plugin
- Check if a targets dependencies includes a package product that sets the flag to true
- Fix the package product reference before continuing

### DIFF
--- a/Sources/XcodeProj/Objects/SwiftPackage/XCSwiftPackageProductDependency.swift
+++ b/Sources/XcodeProj/Objects/SwiftPackage/XCSwiftPackageProductDependency.swift
@@ -24,7 +24,7 @@ public class XCSwiftPackageProductDependency: PBXContainerItem, PlistSerializabl
 
     public init(productName: String,
                 package: XCRemoteSwiftPackageReference? = nil,
-                plugin: Bool = false) {
+                plugin: Bool) {
         self.productName =  "\(plugin ? "plugin:" : "")\(productName)"
         packageReference = package?.reference
         self.plugin = plugin
--- a/Sources/XcodeProj/Objects/SwiftPackage/XCSwiftPackageProductDependency.swift
+++ b/Sources/XcodeProj/Objects/SwiftPackage/XCSwiftPackageProductDependency.swift
@@ -17,13 +17,17 @@ public class XCSwiftPackageProductDependency: PBXContainerItem, PlistSerializabl
             packageReference = newValue?.reference
         }
     }
+    
+    let plugin: Bool
 
     // MARK: - Init
 
     public init(productName: String,
-                package: XCRemoteSwiftPackageReference? = nil) {
-        self.productName = productName
+                package: XCRemoteSwiftPackageReference? = nil,
+                plugin: Bool = false) {
+        self.productName =  "\(plugin ? "plugin:" : "")\(productName)"
         packageReference = package?.reference
+        self.plugin = plugin
         super.init()
     }
 
@@ -37,6 +41,7 @@ public class XCSwiftPackageProductDependency: PBXContainerItem, PlistSerializabl
         } else {
             packageReference = nil
         }
+        plugin = try container.decode(Bool.self, forKey: .plugin)
         try super.init(from: decoder)
     }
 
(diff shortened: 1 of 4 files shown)


technocidal:
@marciniwanicki @adamkhazi Just a quick ping: Is there anything I can do to move this along? Do you've any questions or suggestions regarding this feature? I'd be very grateful for any feedback.

Sources/XcodeProj/Objects/SwiftPackage/XCSwiftPackageProductDependency.swift
```diff
@@ -37,6 +41,11 @@ public class XCSwiftPackageProductDependency: PBXContainerItem, PlistSerializabl
         } else {
             packageReference = nil
         }
+        if let plugin = try container.decodeIfPresent(Bool.self, forKey: .plugin) {
+            self.plugin = plugin
+        } else {
+            plugin = false
+        }
```
kwridan, line 48: if I followed along, there isn't a `plugin` attribute in the pbxproj file, rather the product name is prefixed with `plugin:` - in that case would this not require checking for that instead?

kwridan (commented):
Thanks for contributing this @technocidal  and for the investigation in https://github.com/yonaskolb/XcodeGen/issues/1290 👍 

It would be helpful if you're able to add a new fixture with the pbxproj file that contains a build plugin and have a a few tests read & write to allow verifying the serialisation and deserialisation works as expected.

Some examples of past fixtures https://github.com/tuist/XcodeProj/tree/main/Fixtures and associated tests https://github.com/tuist/XcodeProj/blob/main/Tests/XcodeProjTests/Scheme/XCSchemeTests.swift#L695

kwridan:
@all-contributors add @technocidal  for code

Sources/XcodeProj/Objects/SwiftPackage/XCSwiftPackageProductDependency.swift
technocidal: @kwridan Good point! I just realized that I completely misunderstood what this initializer is for. My bad 🤦‍♂️

kwridan: no worries

iammxrn:
Hey there! Any updates?

kwridan:
I've converted back draft while this is being worked on

technocidal:
I think I dropped the ball on this one. Sorry. I'll try to summarise my thoughts which I should've done way earlier.

I can totally understand that getting this done would be a great thing but I'm a bit worried that this will generate a lot of follow-up maintenance. I've also worked on the Plugin Support for SwiftLint which has turned into a massive headache for everyone involved. There are a lot of edge case related to plugins that are not correctly handled by Xcode. This means that it is entirely possible that adding a plugin to your project breaks it. This happens with SwiftLint and it's dependency on swift-syntax quite a lot. You can find more details in this [discussion](https://github.com/realm/SwiftLint/issues/4641#issuecomment-1386295022).

There are also some other weird linker issues that we've filed a Feedback (FB11877146) for but haven't heard from Apple yet.

Regarding the actual implementation. I think I only need to update my fork and add finish implementing the test suite/fixtures but I'm a bit disheartened due to the problems mentioned above.

kwridan:
No worries at all and no rush. 

Thank you for sharing some of gotchas with plugins - sorry to hear about the hurdles you had to face with it.

clausjoergensen:
@technocidal any updates from Apple?

PR #734, opened 2022-12-01
Allow specifying `platformFilter` in `PBXBuildFile` init

maxwellE:
- Add the ability to specify the platform filter in the init for `PBXBuildFile` for convenience.


brentleyjones (approved):

kwridan (approved):
Thanks @maxwellE 

To clarify, this PR introduces a convenience to specify the `platformFitler` in the initializer, the property itself was used directly on an already created instance - for example https://github.com/tuist/tuist/blob/main/Sources/TuistGenerator/Generator/LinkGenerator.swift#L232

PR #736, opened 2022-12-08
🐛 Fix PBXObject Equatable

vguerci:

### Short description 📝
PBXObject `isEqual`, always returning `true`, doesn't respect Swift Hashable's requirements.
That leads to weird/flaky issues like https://github.com/yonaskolb/XcodeGen/issues/1131

### Solution 📦
Make sure `==` is consistent with `hash`, i.e. same `hashValue` implying `isEqual` being `true`

### Implementation 👩‍💻👨‍💻
- [X] Check how other `isEqual` methods are implemented for consistency across the project
- [X] Implement it according to the `hash()` method, i.e. reference is what needs to be equal


vguerci:
Ah damn, that doesn't pass tests, this needs more work ;p
Closing, will reopen when solved.

PR #737, opened 2022-12-08
Add `platformFilters` attribute

maxwellE:
Adds this attribute to PBXBuildFile and
PBXTargetDependency

### DIFF
--- a/Sources/XcodeProj/Objects/BuildPhase/PBXBuildFile.swift
+++ b/Sources/XcodeProj/Objects/BuildPhase/PBXBuildFile.swift
@@ -37,6 +37,9 @@ public final class PBXBuildFile: PBXObject {
     /// Introduced in: Xcode 11
     public var platformFilter: String?
 
+    /// Platform filters attribute.
+    public var platformFilters: [String]?
+
     /// The cached build phase this build file belongs to
     weak var buildPhase: PBXBuildPhase?
 
@@ -51,11 +54,13 @@ public final class PBXBuildFile: PBXObject {
     public init(file: PBXFileElement? = nil,
                 product: XCSwiftPackageProductDependency? = nil,
                 settings: [String: Any]? = nil,
-                platformFilter: String? = nil) {
+                platformFilter: String? = nil,
+                platformFilters: [String]? = nil) {
         fileReference = file?.reference
         productReference = product?.reference
         self.settings = settings
         self.platformFilter = platformFilter
+        self.platformFilters = platformFilters
         super.init()
     }
 
@@ -66,6 +71,7 @@ public final class PBXBuildFile: PBXObject {
         case settings
         case productRef
         case platformFilter
+        case platformFilters
     }
 
     public required init(from decoder: Decoder) throws {
@@ -80,6 +86,7 @@ public final class PBXBuildFile: PBXObject {
         }
         settings = try container.decodeIfPresent([String: Any].self, forKey: .settings)
         platformFilter = try container.decodeIfPresent(.platformFilter)
+        platformFilters = try container.decodeIfPresent([String].self, forKey: .platformFilters)
         try super.init(from: decoder)
     }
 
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Objects/BuildPhase/PBXBuildFile.swift
```diff
@@ -176,6 +183,9 @@ final class PBXBuildPhaseFile: PlistSerializable, Equatable {
         if let platformFilter = buildFile.platformFilter {
             dictionary["platformFilter"] = .string(.init(platformFilter))
         }
+        if let platformFilters = buildFile.platformFilters {
+            dictionary["platformFilters"] = .string(.init(platformFilters.joined(separator: ",")))
```
maxwellE, line 187: @brentleyjones I have no idea what separator should be used here

brentleyjones (approved):

kwridan (approved):
thanks for adding this.

A small suggestion for future contributions, it would be really helpful to capture where / when some of the attributes appear to aid testing and verifying the behaviours.

Luckily I was able to figure out where to look for this :)

In Xcode the filters option appears for targets and allows selecting all, single or multiple platforms

Filters

It appears selecting an individual platform filter results in the attribute `platformFilter` being used:

single-filter



```xml
3DFCA495EE0D3913DAD9804B /* Framework2.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B7F0EC5C1D588EC72A1C5AEA /* Framework2.framework */; platformFilter = ios; };
```

vs selecting multiple platforms, results in the attribute `platformFilters` (plural) being used:

multiple-filters

```xml
3DFCA495EE0D3913DAD9804B /* Framework2.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B7F0EC5C1D588EC72A1C5AEA /* Framework2.framework */; platformFilters = (ios, macos, tvos, ); };
```

From what I gather those are mutually exclusive. Xcode seems to prioritise `platformFilters` if both are present however upon making any changes would get rid on `platformFilter`. To maintain a backwards compatible API the approach taken here looks reasonable, but perhaps for a future major version we could consider consolidating the properties and internally have `XcodeProj` use the appropriate attribute based on single vs multiple filters.

brentleyjones:
Just a counter-point, Xcode sometimes sets single element versions of the plural value:

```
platformFilters = (tvos, );
```

maxwellE:
> thanks for adding this.
> 
> A small suggestion for future contributions, it would be really helpful to capture where / when some of the attributes appear to aid testing and verifying the behaviours.
> 
> Luckily I was able to figure out where to look for this :)
> 
> In Xcode the filters option appears for targets and allows selecting all, single or multiple platforms
> 
> Filters
> 
> It appears selecting an individual platform filter results in the attribute `platformFilter` being used:
> 
> single-filter
> 
> ```
> 3DFCA495EE0D3913DAD9804B /* Framework2.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B7F0EC5C1D588EC72A1C5AEA /* Framework2.framework */; platformFilter = ios; };
> ```
> 
> vs selecting multiple platforms, results in the attribute `platformFilters` (plural) being used:
> 
> multiple-filters
> 
> ```
> 3DFCA495EE0D3913DAD9804B /* Framework2.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B7F0EC5C1D588EC72A1C5AEA /* Framework2.framework */; platformFilters = (ios, macos, tvos, ); };
> ```
> 
> From what I gather those are mutually exclusive. Xcode seems to prioritise `platformFilters` if both are present however upon making any changes would get rid on `platformFilter`. To maintain a backwards compatible API the approach taken here looks reasonable, but perhaps for a future major version we could consider consolidating the properties and internally have `XcodeProj` use the appropriate attribute based on single vs multiple filters.

I will make sure to follow this workflow next time.

kwridan:
> Just a counter-point, Xcode sometimes sets single element versions of the plural value:

fun times 😅

> I will make sure to follow this workflow next time.

Thank you

Issue #738, opened 2022-12-21
XcodeProj Writable conformance does not include WorkspaceSettings


dayton-bobbitt:
## Context 🕵️‍♀️
I need to prevent Xcode from automatically creating schemes by setting `IDEWorkspaceSharedSettings_AutocreateContextsIfNeeded` to `false` in `WorkspaceSettings.xcsettings`. 

## What 🌱
`WorkspaceSettings` already supports this via `autoCreateSchemes`, so I initialized `XcodeProj` with custom `WorkspaceSettings` in order to disable this feature; however, the `WorkspaceSettings.xcsettings` file that `WorkspaceSettings` represents is never created when the project is written:

```swift
let sharedData = XCSharedData(workspaceSettings: WorkspaceSettings(autoCreateSchemes: false))
let project = XcodeProj(workspace: xcworkspace, pbxproj: pbxproj, sharedData: sharedData)

try project.write(path: path) // WorkspaceSettings's write method is never invoked, so WorkspaceSettings.xcsettings is not written
```

## Proposal 🎉
`XcodeProj`'s `Writable` conformance includes methods for writing other aspects of `XCSharedData`, including `schemes` (XcodeProj.swift:159) and `breakpoints` (XcodeProj.swift:196).

My proposal is to create a new method called `writeWorkspaceSettings` with the following signature:
```swift
public func writeWorkspaceSettings(path: Path, override: Bool = true) throws {
    // Unwrap `sharedData`
    // Invoke `sharedData.workspaceSettings.write(path: workspaceSettingsPath, override: override)`
}
```

This method will be called within `XcodeProj`'s `write` method (XcodeProj.swift:85), resulting in `WorkspaceSettings.xcsettings` being written:
```swift
public func write(path: Path, override: Bool = true, outputSettings: PBXOutputSettings) throws {
    try path.mkpath()
    try writeWorkspace(path: path, override: override)
    try writePBXProj(path: path, override: override, outputSettings: outputSettings)
    try writeSchemes(path: path, override: override)
    try writeBreakPoints(path: path, override: override)
    try writeWorkspaceSettings(path: path, override: override) // Invoke new method here
}
```

---

P.S. `XCSharedData` does not conform to `Writable`, meaning that `XcodeProj.swift` contains the implementation for writing schemes and breakpoints to file, accounting for 77 of the files 207 lines. The addition of a new method for writing `WorkspaceSettings` will only add further bloat.

[...]

PR #743, opened 2023-02-12
Update XCSharedData Writable conformance to include WorkspaceSettings

dayton-bobbitt:
Resolves https://github.com/tuist/XcodeProj/issues/738

- [x] Write a test case in `XcodeProjTests` that fails if `XcodeProj.write` doesn’t write all expected files
- [x] Run the test and see that it fails
- [x] Update `XCSharedData` `Writable` conformance to write `WorkspaceSettings`
- [x] Run `XcodeProjTests` again and see that it passes

### DIFF
--- a/Sources/XcodeProj/Project/WorkspaceSettings.swift
+++ b/Sources/XcodeProj/Project/WorkspaceSettings.swift
@@ -153,3 +153,9 @@ public class WorkspaceSettings: Codable, Equatable, Writable {
         try path.write(data)
     }
 }
+
+extension WorkspaceSettings {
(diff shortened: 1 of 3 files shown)


kwridan (approved):
Thanks @dayton-bobbitt

dayton-bobbitt:
Thanks @kwridan! Is there anyone else in particular we need to review this change before it can be merged?

brentleyjones (approved):

kwridan:
Thanks @brentleyjones for the quick review!

PR #745, opened 2023-02-21
docs: Add try-catch to getting started

AlexKobachiJP:
Not sure if this was left out on purpose to keep the sample as small as possible but I feel that a getting started document should show code that builds without requiring fixes, even if it is as simple as adding a `try` statement, because it otherwise makes the documentation look sloppy.

(Apologies for not creating an issue first.)

### DIFF
--- a/Documentation/getting-started.md
+++ b/Documentation/getting-started.md
@@ -57,7 +57,11 @@ import PathKit
 import XcodeProj
 
 let path = Path("/path/to/my/Project.xcodeproj") // Your project path
-let xcodeproj = XcodeProj(path: path)
+do {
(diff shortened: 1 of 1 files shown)


kwridan (approved):
Thanks for contributing this @AlexKobachiJP 

Agreed, making the docs closer to reality is needed. At the very least `try ...` incase someone does copy and paste the snippet. 👍

kwridan:
https://github.com/all-contributors add @AlexKobachiJP for docs

kwridan:
@all-contributors add @AlexKobachiJP for docs

AlexKobachiJP:
Thanks!

Issue #756, opened 2023-04-22
`XCSchemeManagement` reads from path are unstable


kwridan:
## Issue

It appears creating `XCSchemeManagement` from a path can yield an unstable order for the `schemeUserState` internally which leads to two identical `XCSchemeManagement` not reporting they are equal.

## Details

- `XCSchemeManagement` decodes its values from a plist file
- `schemeUserState` is a dictionary within the plist
- This is then internally converted to an array of `[XCSchemeManagement.UserStateScheme]` without any sorting 
- This results in the order of elements within the final `XCSchemeManagement` being unstable for the same plist file

## Reference

_Originally posted by @JCSooHwanCho in https://github.com/tuist/tuist/issues/5154#issuecomment-1518671235_

JCSooHwanCho:
I made [PR](https://github.com/tuist/XcodeProj/pull/757) about it. Could you check it please?

JCSooHwanCho:
Oh, It's duplicate with your [PR](https://github.com/tuist/XcodeProj/pull/758)

PR #757, opened 2023-04-22
change UserStateScheme property in XCUserData from Array into Set

JCSooHwanCho:
Resolves https://github.com/tuist/xcodeproj/issues/756

### Short description 📝
Although `UserStateScheme` in `XCUserData` has dictionary representation actually, It was converted as Array. This bug is blocking bumping up dependency version in Tuist. 

related PR: [here](https://github.com/tuist/tuist/pull/5154)

### Solution 📦
make that property as Set

- [ ] Change property type from Array into Set
- [ ] Refine Code with strong type safety

### DIFF
--- a/Sources/XcodeProj/Scheme/XCSchemeManagement.swift
+++ b/Sources/XcodeProj/Scheme/XCSchemeManagement.swift
@@ -26,16 +26,7 @@ public struct XCSchemeManagement: Codable, Equatable, Writable {
     }
     
     /// Scheme configuration object.
-    public struct UserStateScheme: Equatable, Codable {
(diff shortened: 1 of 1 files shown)


kwridan:
Thanks @JCSooHwanCho - looks like we both submitted fixes 😅

https://github.com/tuist/XcodeProj/pull/758

We'll likely need to keep the existing array to maintain the current public API without introducing breaking changes. I think this PR could be a good candidate for the next major release.

danyf90:
Replaced by https://github.com/tuist/XcodeProj/pull/756

PR #761, opened 2023-06-16
Add support for Xcode Runtime issue breakpoints

zenangst:
### Short description 📝
This PR adds support for Xcode runtime issues breakpoints.
It simply adds a new case to the `BreakpointExtensionID` called `.runtimeException` which can be used to generate breakpoints.

Solution 📦

Add missing case to `BreakpointExtensionID` for generating runtime issue breakpoints.

### Implementation 👩‍💻👨‍💻

- [x] Adds new `BreakpointExtensionID` case to support runtime issue breakpoints
- [x] Expands the `XCBreakpointListTests` to include looking for the runtime exception
- [x] Update the breakpoints fixture to include a runtime issue breakpoint

### DIFF
--- a/Sources/XcodeProj/Project/XCBreakpointList.swift
+++ b/Sources/XcodeProj/Project/XCBreakpointList.swift
@@ -316,7 +316,7 @@ public final class XCBreakpointList: Equatable, Writable {
             case symbolic = "Xcode.Breakpoint.SymbolicBreakpoint"
             case ideConstraintError = "Xcode.Breakpoint.IDEConstraintErrorBreakpoint"
             case ideTestFailure = "Xcode.Breakpoint.IDETestFailureBreakpoint"
-            case runtimeException = "Xcode.Breakpoint.RuntimeIssueBreakpoint"
+            case runtimeIssue = "Xcode.Breakpoint.RuntimeIssueBreakpoint"
         }
 
         // MARK: - Attributes
--- a/Sources/XcodeProj/Project/XCBreakpointList.swift
+++ b/Sources/XcodeProj/Project/XCBreakpointList.swift
@@ -316,6 +316,7 @@ public final class XCBreakpointList: Equatable, Writable {
             case symbolic = "Xcode.Breakpoint.SymbolicBreakpoint"
             case ideConstraintError = "Xcode.Breakpoint.IDEConstraintErrorBreakpoint"
             case ideTestFailure = "Xcode.Breakpoint.IDETestFailureBreakpoint"
+            case runtimeException = "Xcode.Breakpoint.RuntimeIssueBreakpoint"
         }
 
         // MARK: - Attributes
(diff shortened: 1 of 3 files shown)


zenangst:
Found a nasty little casing issue which caused the tests to not compile. Fixed here https://github.com/tuist/XcodeProj/pull/761/commits/ad3d477b665f47e185417279f6e027f1e0cb26f3

kwridan (approved):
Thanks @zenangst 

This is a good addition 👍 previously any breakpoints list with runtime exceptions within it would completely fail to parse.

One small suggestion / improvement for completeness, runtime exceptions have an associated `type` that is encoded within `BreakpointContent` for example:

```xml
<BreakpointProxy
   BreakpointExtensionID = "Xcode.Breakpoint.RuntimeIssueBreakpoint">
   <BreakpointContent
      uuid = "0E9F84AF-42FF-4236-B75A-97E449E6C66F"
      shouldBeEnabled = "Yes"
      ignoreCount = "0"
      continueAfterRunningActions = "No"
      breakpointStackSelectionBehavior = "1"
      type = "1">
   </BreakpointContent>
</BreakpointProxy>
```

From a quick trial, I've found the following:

- Thread Sanitizer: `type = 1`
- Undefined Behaviour: `type = 2`
- System Frameworks: `type = 8`
- All - `type = 65535`

It appears to be a 16bit mask. It can be modelled as an option set. That said seeing it's only applicable to runtime exceptions it would mean it would be unused / optional for other types. I haven't quite figured out what a good public API for it would be. Perhaps can be done separately with some more thought put into it - mind raising an issue for it and if you have any ideas / suggestions?

Thanks again!

Sources/XcodeProj/Project/XCBreakpointList.swift
```diff
@@ -316,6 +316,7 @@ public final class XCBreakpointList: Equatable, Writable {
             case symbolic = "Xcode.Breakpoint.SymbolicBreakpoint"
             case ideConstraintError = "Xcode.Breakpoint.IDEConstraintErrorBreakpoint"
             case ideTestFailure = "Xcode.Breakpoint.IDETestFailureBreakpoint"
+            case runtimeException = "Xcode.Breakpoint.RuntimeIssueBreakpoint"
```
kwridan, line 319: if it's not too much hassle can we keep the naming close to how Xcode names it `runtimeIssue` ?

zenangst: [5081efd](https://github.com/tuist/XcodeProj/pull/761/commits/5081efd1bf99614754bdab4ef44aa6b3cfd7d74d)

zenangst:
@kwridan I tried something really quickly but I ended up with the clash between using `CaseIterable` and having an associated type on the enum. They don't really work well together.

I think to really nails this, I'll open an issue, so that the rest of the community can also chip in an spitball some ideas, like you suggested.

I pushed a commit to improve on the naming.

zenangst:
Made an issue here - https://github.com/tuist/XcodeProj/issues/763

kwridan:
Awesome thanks @zenangst 🙏

pepicrft:
Thanks for tackling this one @zenangst. I'll go ahead with merging this and I'll tackle your issue separately.

Issue #782, opened 2023-07-08
Better messaging for corrupt `xcworkspace` inside `xcodeproj` files


vijaytholpadi:
## Context 🕵️‍♀️
When trying to migrate the build settings in one of the project files at my company, I noticed that I was shown the following error. It did not provide much information on where I should be look debugging. We could improve the messaging on failures to parse `xcworkspace` files.

## What 🌱
```
We received an error that we couldn't handle:
    - Localized description: The operation couldn't be completed. (AEXML.AEXMLError error 2.)
    - Error: parsingFailed

If you think it's a legit issue, please file an issue including the reproducible steps: https://github.com/tuist/tuist/issues/new/choose
Consider creating an issue using the following link: https://github.com/tuist/tuist/issues/new/choose
```

On deeper inspection through debugging, it seem the project file had an `project.xcworkspace` file inside it whose `contents.xcworkspacedata` file had the following contents

```
<?xml version="1.0" encoding="UTF-8"?>
<Workspace
   version = "1.0">
   <FileRef
      location = "self:">
   </FileRef>
</Workspace>
```

## Proposal 🎉
The above reason was not evident when the `tuist migration list-targets -p <project>.xcodeproj` command was run and required further debugging into `Tuist` and `XcodeProj`.

We should try to add messaging which could help the users identify the issue sooner by going directly to the affected xcworkspace file and fixing it to unblock their migration.

PR #783, opened 2023-07-08
Improves messaging of XCWorkspaceData init with details of the failed workspace

vijaytholpadi:
Resolves https://github.com/tuist/XcodeProj/issues/782

### Short description 📝
Improves messaging of XCWorkspaceData init with details of the failed workspace for cases when the workspace file is embedded within the project file and not obvious.

### Solution 📦
Add messaging for the init of `XCWorkspaceData`

### Implementation 👩‍💻👨‍💻
- [x] Catching AEXMLError.parsingFailed exceptions thrown in `XCWorkspaceData` class
- [x] Report FatalError


Sources/XcodeProj/Workspace/XCWorkspaceData.swift
```diff
@@ -28,13 +28,17 @@ extension XCWorkspaceData: Writable {
...
+        do {
+            let xml = try AEXMLDocument(xml: path.read())
+            let children = try xml
+                .root
+                .children
+                .compactMap(XCWorkspaceDataElement.init(element:))
+
+            self.init(children: children)
+        } catch AEXMLError.parsingFailed {
+            fatalError("We could not parse the file \(path.lastComponent). Make sure it's a valid XML/Plist/JSON file and try again")
```
pepicrft, line 40: I'd refrain from `fatalError` here because the caller might want to catch the error and recover from it. Therefore, I'd suggest to define a `XCWorkspaceDataError.parseError` error type that you can throw from here. That approach also makes testing easier.

Issue #786, opened 2023-07-31
Add an option to get the String representation of PBXProj.


Ibrahimhass:
## Context 🕵️‍♀️
> I want to modify a `pbxproj` in memory to later add it to an in-memory archive. 
The current APIs cannot achieve this without first writing the `pbxproj` it to disk.

## What 🌱
> The project generates a String representation of the `pbxproj` internally before saving to disk. 
Below is the relevant code block inside [PBXProj.swift](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Objects/Project/PBXProj.swift).
```
extension PBXProj: Writable {
    public func write(path: Path, override: Bool) throws {
        try write(path: path, override: override, outputSettings: PBXOutputSettings())
    }

    public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) throws {
        let encoder = PBXProjEncoder(outputSettings: outputSettings)
        let output = try encoder.encode(proj: self)
        if override, path.exists {
            try path.delete()
        }
        try path.write(output)
    }
}
```

## Proposal 🎉
With the following simple refactor, this can be achieved with ease.
```
extension PBXProj: Writable {
    public func write(path: Path, override: Bool) throws {
        try write(path: path, override: override, outputSettings: PBXOutputSettings())
    }

    public func getStringRepresentation(outputSettings: PBXOutputSettings) throws -> String {
        let encoder = PBXProjEncoder(outputSettings: outputSettings)
        return try encoder.encode(proj: self)
    }

    public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) throws {
        let output = try getStringRepresentation(outputSettings: outputSettings)
        if override, path.exists {
            try path.delete()
        }
        try path.write(output)
    }
}
```

PR #787, opened 2023-07-31
Get Data representation for `Writable`

Ibrahimhass:
Resolves https://github.com/tuist/XcodeProj/issues/786

I want to modify a `pbxproj` in memory to add it to an in-memory archive later.
This is needed because we are making a server-side Swift application.
Since it is a server-side application, it would be great to get the modified `pbxproj` without any extra disk I/O operations.
The current APIs can only achieve this by first writing the `pbxproj` to disk.

Also refactored the code to use the above function for better code reuse.

### Implementation 👩‍💻👨‍💻
```
extension PBXProj: Writable {
        try write(path: path, override: override, outputSettings: PBXOutputSettings())
    }

    public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) throws {
        let output = try encodeAsString(outputSettings: outputSettings)
        if override, path.exists {
            try path.delete()
        }
        try path.write(output)
    }
    
    public func encodeAsString(outputSettings: PBXOutputSettings) throws -> String {
        let encoder = PBXProjEncoder(outputSettings: outputSettings)
        return try encoder.encode(proj: self)
    }
}
```

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProj.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProj.swift
@@ -285,7 +285,7 @@ extension PBXProj: Equatable {
 
 extension PBXProj: Writable {
     public func dataRepresentation(outputSettings: PBXOutputSettings) throws -> Data? {
-        let encoder = PBXProjEncoder(outputSettings: PBXOutputSettings())
+        let encoder = PBXProjEncoder(outputSettings: outputSettings)
         return try encoder.encode(proj: self).data(using: .utf8)
     }
 
--- a/Sources/XcodeProj/Objects/Project/PBXProj.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProj.swift
@@ -288,16 +288,16 @@ extension PBXProj: Writable {
         try write(path: path, override: override, outputSettings: PBXOutputSettings())
     }
 
-    public func getStringRepresentation(outputSettings: PBXOutputSettings) throws -> String {
-        let encoder = PBXProjEncoder(outputSettings: outputSettings)
-        return try encoder.encode(proj: self)
-    }
-
     public func write(path: Path, override: Bool, outputSettings: PBXOutputSettings) throws {
-        let output = try getStringRepresentation(outputSettings: outputSettings)
+        let output = try encodeAsString(outputSettings: outputSettings)
         if override, path.exists {
             try path.delete()
         }
         try path.write(output)
+    }
+    
+    public func encodeAsString(outputSettings: PBXOutputSettings) throws -> String {
+        let encoder = PBXProjEncoder(outputSettings: outputSettings)
+        return try encoder.encode(proj: self)
     }
 }
--- a/Sources/XcodeProj/Objects/Project/PBXProj.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProj.swift
@@ -288,7 +288,7 @@ extension PBXProj: Writable {
         try write(path: path, override: override, outputSettings: PBXOutputSettings())
     }
 
-    public func getStringRepresentation(outputSettings: PBXOutputSettings) throws -> String {
+    public func getStringRepresentation(outputSettings: PBXOutputSettings = PBXOutputSettings()) throws -> String {
         let encoder = PBXProjEncoder(outputSettings: outputSettings)
         return try encoder.encode(proj: self)
     }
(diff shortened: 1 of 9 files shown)


luispadron (approved):

kwridan (commented):
Thanks for contributing this @Ibrahimhass and for outlining the motivation behind it.

My initial impression is, while this achieves the desired goal it seems a bit ad-hoc - `PBXProj` is one of many types that get serialised to disk.

Perhaps a more wholistic approach would be to update `Writable` to include `stringRepresentation()` method such that all writable types behave consistently?

Ibrahimhass:
@kwridan Thanks for the suggestion :)
Looking further, I found that the underlying type for Writable [here](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Project/WorkspaceSettings.swift) is `Data`.
So shall we have `stringRepresentation()` as an Optional requirement for Writable?

Something similar to this:
```
/// Protocol that defines how an entity can be written to disk
public protocol Writable {
    /// Writes the object that conforms the protocol.
    ///
    /// - Parameter path: The path to write to
    /// - Parameter override: True if the content should be overridden if it already exists.
    /// - Throws: writing error if something goes wrong.
    func write(path: Path, override: Bool) throws

    /// Writes the object that conforms the protocol.
    ///
    /// - Parameter pathString: The path string to write to
    /// - Parameter override: True if the content should be overridden if it already exists.
    /// - Throws: writing error if something goes wrong.
    func write(pathString: String, override: Bool) throws
    
    /// Gets  the string representation of the object  that conforms the protocol.
    ///
    /// - Throws: error if encoding to String goes wrong.
    func stringRepresentation() throws -> String
}

extension Writable {
    public func write(pathString: String, override: Bool) throws {
        let path = Path(pathString)
        try write(path: path, override: override)
    }
    
    public func stringRepresentation() throws -> String { "" }
}
```

kwridan:
> @kwridan Thanks for the suggestion :)
Looking further, I found that the underlying type for Writable [here](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Project/WorkspaceSettings.swift) is Data.
So shall we have stringRepresentation() as an Optional requirement for Writable?

Thanks for looking further into this would making `dataRepresentation()` be a better fit then?

Ibrahimhass:
@kwridan 
> Thanks for looking further into this. Would making dataRepresentation() be a better fit then?

Thanks for the suggestion. Yes, this could work, we modify Writable to something like:

```
/// Protocol that defines how an entity can be written to disk
public protocol Writable {
    /// Writes the object that conforms the protocol.
    ///
    /// - Parameter path: The path to write to
    /// - Parameter override: True if the content should be overridden if it already exists.
    /// - Throws: writing error if something goes wrong.
    func write(path: Path, override: Bool) throws

    /// Writes the object that conforms the protocol.
    ///
    /// - Parameter pathString: The path string to write to
    /// - Parameter override: True if the content should be overridden if it already exists.
    /// - Throws: writing error if something goes wrong.
    func write(pathString: String, override: Bool) throws
    
    /// Gets the data representation of the object that conforms to the protocol.
    ///
    /// - Throws: error if encoding to Data goes wrong.
    func dataRepresentation() throws -> Data
}
```
The representation can be used to get the string representation of any of the file if required.

Ibrahimhass:
@kwridan Please review the changes.

Sources/XcodeProj/Objects/Project/PBXProj.swift
```diff
@@ -284,13 +284,24 @@ extension PBXProj: Equatable {
 // MARK: - Writable
 
 extension PBXProj: Writable {
+    public func dataRepresentation(outputSettings: PBXOutputSettings) throws -> Data? {
+        let encoder = PBXProjEncoder(outputSettings: PBXOutputSettings())
```
kwridan, line 288:
```suggestion
        let encoder = PBXProjEncoder(outputSettings: outputSettings)
```

kwridan (commented):
Thanks @Ibrahimhass, one small correction needed.

Ibrahimhass:
> Thanks @Ibrahimhass, one small correction needed.

Thanks a lot for pointing this out, I have fixed the code.

Sources/XcodeProj/Objects/Project/PBXProj.swift
Ibrahimhass: @kwridan Thanks for pointing this out, I have fixed this now.

kwridan (approved):
Thanks @Ibrahimhass

kwridan:
@all-contibutors add @Ibrahimhass for code

kwridan:
@all-contributors add @Ibrahimhass for code

Issue #793, opened 2023-09-01
Create PBXProj from Data Representation


Ibrahimhass:
## Context 🕵️‍♀️
> We are working on a Server Side Swift application. We copy an archive containing a  template Xcode project stored on disk to memory using [ZipFoundation](https://github.com/weichsel/ZIPFoundation). We want to be able to re-create a PBXProj object from its data entry in the archive.

## What 🌱
> The project provides the following initializers for PBXProj
```
public init(rootObject: PBXProject? = nil,
    objectVersion: UInt = Xcode.LastKnown.objectVersion,
    archiveVersion: UInt = Xcode.LastKnown.archiveVersion,
    classes: [String: Any] = [:],
    objects: [PBXObject] = []) { ... }

public convenience init(path: Path) throws { ... }
```

## Proposal 🎉
> We can have a simple initializer that can create a PBXProj from the project data.
```
public convenience init(data: Data) throws {
    var propertyListFormat = PropertyListSerialization.PropertyListFormat.xml

    let serialized = try PropertyListSerialization.propertyList(
        from: data,
        options: .mutableContainersAndLeaves,
        format: &propertyListFormat
    )

    // swiftlint:disable:next force_cast
    let pbxProjDictionary = serialized as! [String: Any]
        
    let context = ProjectDecodingContext(
        pbxProjValueReader: { key in
            pbxProjDictionary[key]
        }
    )

    let plistDecoder = XcodeprojPropertyListDecoder(context: context)
    let pbxproj: PBXProj = try plistDecoder.decode(PBXProj.self, from: data)

    self.init(
        rootObject: pbxproj.rootObject,
        objectVersion: pbxproj.objectVersion,
        archiveVersion: pbxproj.archiveVersion,
        classes: pbxproj.classes,
        objects: pbxproj.objects
    )
}
```

PR #798, opened 2023-09-12
Create `PBXProj` class from the Data reprentation of a `pbxproj` file

Ibrahimhass:
https://github.com/tuist/XcodeProj/issues/793

The current APIs can create a PBXProj class when the path to a `pbxproj` file is provided.
I want to initialize the PBXProj file from the data representation of a `pbxproj` stored inside an in-memory archive.
This is needed because we are making a server-side Swift application and want to minimize disk operations.

### Implementation 👩‍💻👨‍💻
```
/// Initializes the project with the data representation of pbxproj file.
///
/// - Parameters:
///   - data: data representation of pbxproj file.
public convenience init(data: Data) throws {
    var propertyListFormat = PropertyListSerialization.PropertyListFormat.xml

    let serialized = try PropertyListSerialization.propertyList(
        from: data,
        options: .mutableContainersAndLeaves,
        format: &propertyListFormat
    )

    // swiftlint:disable:next force_cast
    let pbxProjDictionary = serialized as! [String: Any]
        
    let context = ProjectDecodingContext(
        pbxProjValueReader: { key in
            pbxProjDictionary[key]
        }
    )

    let plistDecoder = XcodeprojPropertyListDecoder(context: context)
    let pbxproj: PBXProj = try plistDecoder.decode(PBXProj.self, from: data)

    self.init(
        rootObject: pbxproj.rootObject,
        objectVersion: pbxproj.objectVersion,
        archiveVersion: pbxproj.archiveVersion,
        classes: pbxproj.classes,
        objects: pbxproj.objects
    )
}
```

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProj.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProj.swift
@@ -80,9 +80,10 @@ public final class PBXProj: Decodable {
             format: &propertyListFormat
         )
 
-        // swiftlint:disable:next force_cast
-        let pbxProjDictionary = serialized as! [String: Any]
-            
+        guard let pbxProjDictionary = serialized as? [String: Any] else {
+            throw PBXProjError.malformed
+        }
+
         let context = ProjectDecodingContext(
             pbxProjValueReader: { key in
                 pbxProjDictionary[key]
@@ -180,8 +181,11 @@ public final class PBXProj: Decodable {
             options: .mutableContainersAndLeaves,
             format: &propertyListFormat
         )
-        // swiftlint:disable:next force_cast
-        let pbxProjDictionary = serialized as! [String: Any]
+
+        guard let pbxProjDictionary = serialized as? [String: Any] else {
+            throw PBXProjError.malformed
+        }
+
         return (plistXML, pbxProjDictionary)
     }
 }
--- a/Sources/XcodeProj/Objects/Project/PBXProj.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProj.swift
@@ -66,6 +66,41 @@ public final class PBXProj: Decodable {
             objects: pbxproj.objects
         )
     }
+    
+    /// Initializes the project with the data representation of pbxproj file.
+    ///
+    /// - Parameters:
+    ///   - data: data representation of pbxproj file.
+    public convenience init(data: Data) throws {
+        var propertyListFormat = PropertyListSerialization.PropertyListFormat.xml
+
+        let serialized = try PropertyListSerialization.propertyList(
+            from: data,
+            options: .mutableContainersAndLeaves,
+            format: &propertyListFormat
+        )
+
+        // swiftlint:disable:next force_cast
+        let pbxProjDictionary = serialized as! [String: Any]
+            
+        let context = ProjectDecodingContext(
+            pbxProjValueReader: { key in
+                pbxProjDictionary[key]
+            }
+        )
+
+        let plistDecoder = XcodeprojPropertyListDecoder(context: context)
+        let pbxproj: PBXProj = try plistDecoder.decode(PBXProj.self, from: data)
+
+        self.init(
+            rootObject: pbxproj.rootObject,
+            objectVersion: pbxproj.objectVersion,
+            archiveVersion: pbxproj.archiveVersion,
+            classes: pbxproj.classes,
+            objects: pbxproj.objects
+        )
+    }
+
 
     private init(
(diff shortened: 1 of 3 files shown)


Sources/XcodeProj/Objects/Project/PBXProj.swift
```diff
@@ -66,6 +66,40 @@ public final class PBXProj: Decodable {
             objects: pbxproj.objects
         )
     }
+    
+    /// Initializes the project with the data representation of pbxproj file.
+    ///
+    /// - Parameters:
+    ///   - data: data representation of pbxproj file.
+    public convenience init(data: Data) throws {
```
maxwellE, line 74: Can we add a simple test to exercise this?

Ibrahimhass: Yes sure, I will add a unit test and write back here.

Ibrahimhass: @maxwellE Thanks for the suggestion I have added a simple test it just tests if PBXProj is properly initialized with the data representation of the pbxproj file.

```
func test_initialize_PBXProj_with_data() {
    let pbxprojPath = iosProjectPath.url.appendingPathComponent("project.pbxproj")
    
    if let pbxprojStringRepresentation = try? String(contentsOf: pbxprojPath) {
        let pbxprojData = Data(pbxprojStringRepresentation.utf8)
        
        XCTAssertNotNil(try? PBXProj(data: pbxprojData))
        return
    }

    XCTFail()
}
```

maxwellE (approved):

Sources/XcodeProj/Objects/Project/PBXProj.swift
maxwellE: Thanks!

Sources/XcodeProj/Objects/Project/PBXProj.swift
```diff
@@ -66,6 +66,40 @@ public final class PBXProj: Decodable {
...
+        var propertyListFormat = PropertyListSerialization.PropertyListFormat.xml
+
+        let serialized = try PropertyListSerialization.propertyList(
+            from: data,
+            options: .mutableContainersAndLeaves,
+            format: &propertyListFormat
+        )
+
+        // swiftlint:disable:next force_cast
+        let pbxProjDictionary = serialized as! [String: Any]
```
luispadron, line 84: nit: Can this just throw instead of `as!`

luispadron (approved):

Sources/XcodeProj/Objects/Project/PBXProj.swift
Ibrahimhass: @luispadron I am now throwing an error for this case and another case where this pattern is used.

Tests/XcodeProjTests/Project/XcodeProjTests.swift
```diff
@@ -25,6 +25,19 @@ final class XcodeProjIntegrationTests: XCTestCase {
...
+        
+        if let pbxprojStringRepresentation = try? String(contentsOf: pbxprojPath) {
+            let pbxprojData = Data(pbxprojStringRepresentation.utf8)
+            
+            XCTAssertNotNil(try? PBXProj(data: pbxprojData))
+            return
+        }
+
+        XCTFail()
+    }
```
kwridan, line 39: For completeness it would be good to check the correctness of initialing from data representation in addition to verifying it's not `nil`

```suggestion
    func test_initialize_PBXProj_with_data() throws {
        // Given
        let pbxprojPath = iosProjectPath + "project.pbxproj"
        let pbxprojFromDisk = try PBXProj(path: pbxprojPath)
        let pbxprojData = try Data(contentsOf: pbxprojPath.url)

        // When
        let pbxprojFromData = try PBXProj(data: pbxprojData)
        try pbxprojFromData.updateProjectName(path: pbxprojPath)

        // Then
        XCTAssertEqual(pbxprojFromData, pbxprojFromDisk)
    }
```

actually doing so reveals initialising from data representation may require also specifying the `name` which currently doesn't get serialised (from a cursory glance, doesn't seem to be part of the project format) .

kwridan (commented):
Thanks @Ibrahimhass

kwridan (approved):
Thanks @Ibrahimhass 

For your use case did you run into issues with `name` being omitted / missing? would that require some additional follow up?

Ibrahimhass:
@kwridan For my use case, `name` does not matter. I am using the forked version for development right now.

PR #799, opened 2023-10-08
Add XCLocalSwiftPackageReference Support

art-divin:
Resolves https://github.com/krzysztofzablocki/Sourcery/issues/1206

### Short description 📝

This PR implements parsing of new element type named `XCLocalSwiftPackageReference` which seems to be introduced with the release of Xcode 15.

### Solution 📦

To support the new element type, I simply copied everything related to `XCRemoteSwiftPackageReference` and removed all code related to VersionMatching and "URL". Instead, only `path` argument is supported for instantiating `XCLocalSwiftPackageReference`.

### Implementation 👩‍💻👨‍💻

- [x] Adopt `XCRemoteSwiftPackageReference` to `XCLocalSwiftPackageReference`
- [x] Add `XCLocalSwiftPackageReference` to `PBXObjectParser` switch statement
- [x] Add caching of `XCLocalSwiftPackageReference` objects in `PBXObjects`
- [x] Add `XCLocalSwiftPackageReference` support in `PBXProject`
- [x] Add integration tests to `PBXProjEncoderTests`


art-divin:
Hello @kwridan, @pepicrft, @yonaskolb 👋🏻 

Firstly, thank you very much for supporting this project. It is used by me in many different Swift scripts I wrote.

Could you please review this PR? It is mentioned in Sourcery (https://github.com/krzysztofzablocki/Sourcery/issues/1206) and [other projects](https://github.com/peripheryapp/periphery/issues/653) which face this issue on Xcode 15.
I am surprised that it was [already implemented in Xcodeproj](https://github.com/CocoaPods/Xcodeproj/pull/911), but not here 🕵🏻 

I hope I did not miss anything, but if I did, could you please be so kind informing me, I'll apply the change as soon as I can.

Thank you 🙏🏻

kwridan (commented):
Thanks for contributing this @art-divin 

I wasn't aware Xcode 15 changed this 😅, local packages were previously supported as file references and Xcode somehow deals with them correctly. It seems this is still the case in Xcode 15 if you add a local package via dragging from Finder to Xcode, however if you use the Package editor UI and select "Add Local" this is where this new format is used.

xc15

From testing locally, it seems we may be missing a few pieces. I created a project with a local package, read it via XcodeProj and wrote it back to disk again - the expectation was that it would result in no diffs, however noticed a few things got omitted 🤔


```diff
-                       packageReferences = (
-                               FBE04FC42AD3E7F70014D6BD /* XCLocalSwiftPackageReference "../LocalPackage" */,
-                       );
```

```diff
 /* End XCConfigurationList section */
 
-/* Begin XCLocalSwiftPackageReference section */
-               FBE04FC42AD3E7F70014D6BD /* XCLocalSwiftPackageReference "../LocalPackage" */ = {
-                       isa = XCLocalSwiftPackageReference;
-                       relativePath = ../LocalPackage;
-               };
-/* End XCLocalSwiftPackageReference section */
-
```

It'll be helpful to include a small fixture of a project with a local package using this new format in `Fixtures` folder and performing a similar integration test (see some of the example test in `XcodeProjIntegrationTests`).

art-divin:
👋🏻 Hello @kwridan ,

thank you very much for your test! 

Apparently, decoding was not set properly. So I have added an integration test to `PBXProjEncoderTests` and the missing implementation to `PBXProjEncoder.swift`.

What I have discovered is that `addSwiftPackage` is missing for "the new way" of adding local references.

I guess it is better to split this feature into multiple PRs, and firstly, implement support to unblock further work.

What do you think?
Any other tests I'd be better adding?

Thank you 🙏🏻

art-divin:
Hey @kwridan ,

I have added one more integration test, and seems that everything[^1] is settled now.

Please let me know if this PR can be merged or still it is missing somethig.

P.S. when run locally, I got failed unit tests related to `git` checks, specifically, locally I have enabled all commits signed by default with gpg. So when run, these unit tests fail with the following error:

```
test_read_write_produces_no_diff(): failed: caught error: "Error Domain=NSPOSIXErrorDomain Code=128 "Unknown error: 128""
```

in file `XCTestCase+Shell.swift` on line 21. Console states the following:
```bash
Test Case '-[XcodeProjTests.XcodeProjIntegrationTests test_read_write_produces_no_diff]' started.
error: cannot run gpg: No such file or directory
error: gpg failed to sign the data
fatal: failed to write commit object
```

[^1]: except for `addSwiftPackage` which lacks functionality to add local package as `XCLocalSwiftPackageReference`.

art-divin:
cc @pepicrft, @yonaskolb

Tests/XcodeProjTests/Objects/Project/PBXProjEncoderTests.swift
```diff
@@ -419,6 +443,10 @@ class PBXProjEncoderTests: XCTestCase {
     private func loadTargetWithCustomBuildRulesProject() {
         proj = try! PBXProj(jsonDictionary: targetWithCustomBuildRulesDictionary().1)
     }
+
+    private func loadProjectWithXCLocalSwiftPackageReference() {
+        proj = try! PBXProj(jsonDictionary: iosProjectWithXCLocalSwiftPackageReference().1)
+    }
```
kwridan, line 449: nit: we can mark the method and test as throwing so the test fails rather than crashes in case of any issues, but I see other existing methods above follow the same pattern so can be a separate tidy up if need be

kwridan (commented):
> I have added one more integration test, and seems that everything[1](https://github.com/tuist/XcodeProj/pull/799/files#user-content-fn-1-afad3b6a1836ecb01abb869af331f109) is settled now.

Thanks for the updates @art-divin and for adding the fixture, note the fixture can be simplified and slimmed down to focus primarily on the local package (i.e. it wouldn't need any of the extra schemes / breakpoint files / remote packages etc...) Locally I created a dummy project with a local package and used that test.

> when run locally, I got failed unit tests related to git checks, specifically, locally I have enabled all commits signed by default with gpg. So when run, these unit tests fail with the following error:

Interesting, indeed looks like it may be related to the signed commits, as locally and on CI those seem to be ok. Perhaps locally only to get it working you can modify the test helper that invokes the git command to include the appropriate extra arguments?

On the topic of this sort of test, I re-ran it against a local project I created with a local swift package and it seems a diff is still produced with the comments:


```diff
@@ -117,7 +117,7 @@
                        );
                        mainGroup = FBF72AFE2AD3E600003BEEEF;
                        packageReferences = (
-                               FBE04FC42AD3E7F70014D6BD /* XCLocalSwiftPackageReference "../LocalPackage" */,
+                               FBE04FC42AD3E7F70014D6BD /* XCLocalSwiftPackageReference "LocalPackage" */,
                        );
                        productRefGroup = FBF72B082AD3E600003BEEEF /* Products */;
                        projectDirPath = "";
@@ -366,7 +366,7 @@
 /* End XCConfigurationList section */
 
 /* Begin XCLocalSwiftPackageReference section */
-               FBE04FC42AD3E7F70014D6BD /* XCLocalSwiftPackageReference "../LocalPackage" */ = {
+               FBE04FC42AD3E7F70014D6BD /* XCLocalSwiftPackageReference "LocalPackage" */ = {
                        isa = XCLocalSwiftPackageReference;
                        relativePath = ../LocalPackage;
                };
```

it might be worth including that sort of test against the fixture?

Tests/XcodeProjTests/Objects/Project/PBXProjEncoderTests.swift
art-divin: Fixed, removed force unwraping 🤝

Sources/XcodeProj/Objects/Project/PBXProject.swift
```diff
@@ -125,13 +125,37 @@ public final class PBXProject: PBXObject {
...
     }
 
-    /// Package references.
+    /// Remote (`XCRemoteSwiftPackageReference`) and Local (`XCLocalSwiftPackageReference`) Package references.
     var packageReferences: [PBXObjectReference]?
 
-    /// Swift packages.
-    public var packages: [XCRemoteSwiftPackageReference] {
+    /// Remote Swift packages.
+    public var remotePackages: [XCRemoteSwiftPackageReference] {
```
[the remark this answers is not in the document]
art-divin: Done, thank you ✅

art-divin:
👋🏻  Hey @kwridan ,

Thank you for your throughout review 🤝 

I have addressed all the points mentioned and hopefully this time PR is ready.

However, I noticed that [some other PRs ](https://github.com/tuist/XcodeProj/pull/773/files)also include changes related to contributors. Should I do it here as well? Anything else?

Thank you

kwridan:
@all-contributors add @art-divin for code

kwridan (approved):
> I have addressed all the points mentioned and hopefully this time PR is ready.

Awesome thank you for the great work on this, testing locally it's all working nicely now 👍 

> I noticed that [some other PRs ](https://github.com/tuist/XcodeProj/pull/773/files)also include changes related to contributors. Should I do it here as well? Anything else?

Done via https://github.com/tuist/XcodeProj/pull/800

luispadron (approved):

luispadron:
One nit would be to squash this PR before merging there are few commits in here

art-divin:
Thank you for your approvals 🤝 

Please merge this MR at your convenience, I do not have enough privileges, thank you 👍🏻

ileitch:
@pepicrft Any chance this change can be released soon? People creating new projects in Xcode 15 that use local package references are currently unable to use XcodeProj (or tools that depend on it) because project parsing fails with an error.

PR #801, opened 2023-11-12
Fix issue that multiple products of the same package cannot be added

kevin-kp:
### Short description 📝
Currently, it does not seem to be possible to add two products from a single package using:
```
public func addSwiftPackage(repositoryURL: String,
                            productName: String,
                            versionRequirement: XCRemoteSwiftPackageReference.VersionRequirement,
                            targetName: String) throws -> XCRemoteSwiftPackageReference
```
This is because the duplicate check only takes the reference into account. It should also take the productName into account to make sure that a duplicate *can* exist of the same reference when the productName is different.

### Solution 📦
Add the productName check to the if-test. No other solutions were considered.

### Implementation 👩‍💻👨‍💻

- [ x] Add duplicate check
- [ x] Fix unit tests

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProject.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProject.swift
@@ -440,7 +440,7 @@ extension PBXProject {
 
         let productDependency: XCSwiftPackageProductDependency
         // Avoid duplication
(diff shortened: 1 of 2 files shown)


luispadron (approved):

fortmarek (approved):
👏

Issue #802, opened 2023-11-13
Cannot remove remotePackage


kevin-kp:
## Context 🕵️‍♀️
We are trying to change the version of a package using XcodeProj. We are doing this by removing the package and readding the swift package.

## What 🌱
On 8.12.0, this works without any issues. However, since https://github.com/tuist/XcodeProj/pull/799 it seems that it isn't possible anymore to remove a swift package.
This is because the setter reads the existing packages when the value is changed. This means that it is only possible to add new packages:
```
var finalReferences: [PBXObjectReference] = packageReferences ?? [] // this contains the removed package
let newReferences = newValue.references()
for reference in newReferences {
    if !finalReferences.contains(reference) {
        finalReferences.append(reference)
    }
}
packageReferences = finalReferences
```

art-divin:
Thank you very much @kevin-kp 👋🏻 

I am sorry I did not foresee this behaviour. 🙇🏻

kevin-kp:
@art-divin No problem 😄 It should be fixed once the PR is merged 😄

PR #803, opened 2023-11-13
Fix issue that it is not possible to remove a remote package

kevin-kp:
Resolves https://github.com/tuist/XcodeProj/issues/802

### Implementation 👩‍💻👨‍💻
It checks that only the references are kept that are not part of the current type. 
This is done by getting the object of the reference and checking the type of this reference.

- [ x] Fix issue
- [ x] Add unit test

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProject.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProject.swift
@@ -154,8 +154,8 @@ public final class PBXProject: PBXObject {
         }
     }
 
-    private func setPackageReferences<T: PBXContainerItem>(_ references: [T]) {
-        let newReferences = references.references()
+    private func setPackageReferences<T: PBXContainerItem>(_ packages: [T]) {
+        let newReferences = packages.references()
         var finalReferences: [PBXObjectReference] = packageReferences?.filter { !($0.getObject() is T) } ?? []
         for reference in newReferences {
             if !finalReferences.contains(reference) {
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Objects/Project/PBXProject.swift
```diff
@@ -154,20 +147,24 @@ public final class PBXProject: PBXObject {
...
-            }
-            packageReferences = finalReferences
+            setPackageReferences(newValue)
         }
         get {
             packageReferences?.objects() ?? []
         }
     }
 
+    private func setPackageReferences<T: PBXContainerItem>(_ references: [T]) {
```
luispadron, line 157:
```suggestion
    private func setPackageReferences<T: PBXContainerItem>(_ packages: [T]) {
```

luispadron: This makes it more clear its a package reference and removes the `references.references()` issue

luispadron (approved):

art-divin:
A bit shame that this PR did not make it into 8.16.0 release 😢

luispadron:
I dont have ability to merge. @pepicrft, @danyf90 @fortmarek could one of you take a look?

kwridan (approved):
Thanks for fixing this @kevin-kp 

small note, to correctly remove a package, there's a few other objects and references that need to be removed _(not just from `packageReferences`,  see https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Objects/Project/PBXProject.swift#L207)_

kevin-kp:
@kwridan 
Thank you, I noticed when there were indeed other things still present in the pbxproj and I indeed used that method for the reverse.

PR #809, opened 2024-01-24
XCScheme string initializer

mattmassicotte:
### Short description 📝
The `XcodeProj` type has a convenience initializer that takes a string. Really is convenient, because it hides the PathKit dependency from callers. But `XCScheme` lacks a similar initializer. I'd love one!

### Implementation 👩‍💻👨‍💻

I just copied in the solution from `XcodeProj` 😅

I had a look, and found no tests that use the `XcodeProj` method, so I went with that too. But happy to do more if needed!


kwridan (approved):
Thanks for contributing this @mattmassicotte

kwridan:
@all-contributors add @mattmassicotte for code

PR #821, opened 2024-05-24
Adopt Strict Concurrency checking

waltflanagan:
- Convert `var` to `let` for structured concurrency
- Add Sendable conformance to types triggering warnings
- Add strict concurrency checking to package
- Annotation `PathKit` imports with `@preconcurrency`
- Convert to a global funciton to avoid shared global state.

- [ ] Step 1
- [ ] Step 2

### DIFF
--- a/Sources/XcodeProj/Errors/Errors.swift
+++ b/Sources/XcodeProj/Errors/Errors.swift
@@ -1,5 +1,5 @@
 import Foundation
-import PathKit
+@preconcurrency import PathKit
 
 // MARK: - Xcodeproj
 
@@ -8,7 +8,7 @@ import PathKit
 /// - notFound: the project cannot be found.
 /// - pbxProjNotFound: the .pbxproj file couldn't be found inside the project folder.
 /// - xcworkspaceNotFound: the workspace cannot be found at the given path.
-public enum XCodeProjError: Error, CustomStringConvertible {
+public enum XCodeProjError: Error, CustomStringConvertible, Sendable {
     case notFound(path: Path)
     case pbxprojNotFound(path: Path)
     case xcworkspaceNotFound(path: Path)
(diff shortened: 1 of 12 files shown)


Sources/XcodeProj/Errors/Errors.swift
```diff
@@ -1,5 +1,5 @@
 import Foundation
-import PathKit
+@preconcurrency import PathKit
```
pepicrft, line 2: I wonder if we should take the opportunity to adopt [Path](https://github.com/tuist/Path/), where we forked `AbsolutePath` and `RelativePath` from swift-tools-support, and use those types instead? The long-term plan is to update Tuist to use that package, so we can start by updating its dependencies.
This is something that we can do in a follow-up PR

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
```diff
@@ -5,20 +5,20 @@ import PathKit
 extension XCScheme {
     // swiftlint:disable:next type_body_length
     public final class LaunchAction: SerialAction {
-        public enum Style: String {
+        public enum Style: String, Sendable {
```
pepicrft, line 8: I'm a bit unfamiliar with the new Swift Concurrency requirements, but why is it required to use `Sendable` here?

pepicrft (approved):

pepicrft:
> In the future we may need to evaluate the public interfaces of XcodeProj as currently many are non-final classes which can not be marked Sendable. This may be a feature we want to better support in the future for consumers of the XcodeProj package as any concurrency management will fall to consumers to do manually.

Should we turn the classes into actors whose property mutations must be awaited? The design of the library relies extensively on in-memory references, so whatever model we adopt would have to respect that if we don't want to have to refactor the entire library.

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
waltflanagan: I believe this is because of `static let defaultLaunchStyle`.  Any static type property must be immutable (`let`) and `Sendable`.

waltflanagan:
I agree that whatever is needed for the future of concurrency here will need to be considered. I think we can get XcodeProj and Tuist independently onto strict concurrency checking and then see where the integration causes challenges.  This should help inform what the public interface of `XcodeProj` should be.

Issue #820, opened 2024-05-05
Incorrect name of the target product returned from `productNameWithExtension()`


mikeger:
## Context 🕵️‍♀️
As an author of [XcodeSelectiveTesting](https://github.com/mikeger/XcodeSelectiveTesting) I've been trying to integrate it for fun with [Wire-iOS](https://github.com/wireapp/wire-ios) project.

This project has several libraries, amongst those you can find `WireCanvas` project. This project has a target called WireCanvas that emits a framework. 

## What 🌱
When parsing `WireCanvas` project, XcodeProj is returning `Canvas.framework` from `productNameWithExtension()`. Correct name of the product is `WireCanvas.framework`

## Proposal 🎉
I tried to fix the issue but was not sure what was causing this behavior. There must be some misconfiguration in the WireCanvas project, but from Xcode side I only see `WireCanvas.framework` in other project linked frameworks phase and as a product name

mikeger:
@kwridan thanks so much for the reply. This sounds reasonable, adding support for `xcconfig` resolution might be too big of a feature for a GitHub issue :) 

Would it make sense to enhance the documentation for `productNameWithExtension()` (any other?) method so people using it would have reasonable expectations? I can propose a small change for this if you think this is appropriate.

kwridan:
Sure that sounds like a reasonable change, this contribution would be welcome! thanks @mikeger

mikeger:
@kwridan I created the PR, thanks for your support.

PR #823, opened 2024-05-28
Update comment to reflect productNameWithExtension() and others

mikeger:
Resolves https://github.com/tuist/xcodeproj/issues/820

### Short description 📝

Based on my experience using XcodeProj, I found it confusing that the value of `productNameWithExtension()` and others could be different from the value displayed in Xcode.

### Solution 📦

Based on the discussion on the issue, we agreed to improve the comment on the given methods.

### Implementation 👩‍💻👨‍💻

I added the comments to the methods that are obviously affected. In case there are other methods, I can update them as well.

### DIFF
--- a/Sources/XcodeProj/Objects/Targets/PBXTarget.swift
+++ b/Sources/XcodeProj/Objects/Targets/PBXTarget.swift
@@ -63,8 +63,6 @@ public class PBXTarget: PBXContainerItem {
     public var productName: String?
 
     /// Target product reference.
-    ///
-    /// This property's value may differ from the value displayed in Xcode if the product name is specified through build settings.
     var productReference: PBXObjectReference?
 
     /// Target product.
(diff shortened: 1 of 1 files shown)


Sources/XcodeProj/Objects/Targets/PBXTarget.swift
```diff
@@ -58,9 +58,13 @@ public class PBXTarget: PBXContainerItem {
     public var name: String
 
     /// Target product name.
+    ///
+    /// This property's value may differ from the value displayed in Xcode if the product name is specified through build settings.
     public var productName: String?
 
     /// Target product reference.
+    ///
+    /// This property's value may differ from the value displayed in Xcode if the product name is specified through build settings.
```
kwridan, line 67: This comment is unnecessary here as it's an internal reference 

```suggestion
```

kwridan (commented):
Thanks for contributing this @mikeger

mikeger:
@kwridan thanks for the comment, I updated the PR

kwridan (approved):
Thanks

pepicrft (approved):

pepicrft:
Thanks a lot for the contribution here @mikeger 🙏🏼

pepicrft:
@all-contributors add @mikeger for code

PR #824, opened 2024-06-08
Add `buildArchitectures` to XCScheme+BuildAction

woin2ee:
This PR is related to https://github.com/tuist/tuist/issues/6003

### Short description 📝
It makes enable to set `Override Architectures` option of build action.

### Solution 📦
I observed the changes in the `xcscheme` file as I changed the `Override Architectures` option. After that, I decided to add the attribute in a similar way to the existing implementation of `buildImplicitDependencies` and `runPostActionsOnFailure`.

### Implementation 👩‍💻👨‍💻
- Add an enum to represent the `Override Architectures` option.
- Add `buildArchitectures` property in `XCScheme.BuildAction`.
- Add and Update tests

### DIFF
--- a/Sources/XcodeProj/Scheme/XCScheme+BuildAction.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+BuildAction.swift
@@ -70,12 +70,45 @@ extension XCScheme {
             }
         }
 
+        public enum Architecture {
+            /// A default value.
+            case matchRunDestination // (Apple recommended: default)
+            case universal
+            case useTargetSettings
+
+            fileprivate var xmlString: String? {
+                switch self {
+                case .matchRunDestination:
+                    "Automatic"
+                case .universal:
+                    "All"
+                case .useTargetSettings:
+                    nil
+                }
+            }
+
+            /// Creates a new instance from the given xml string.
+            ///
+            /// For undefined value, initialized as `matchRunDestination`.
+            fileprivate init(_ xmlString: String) {
+                switch xmlString {
+                case "Automatic":
+                    self = .matchRunDestination
+                case "All":
+                    self = .universal
+                default:
+                    self = .matchRunDestination
+                }
+            }
+        }
+
         // MARK: - Attributes
 
         public var buildActionEntries: [Entry]
         public var parallelizeBuild: Bool
         public var buildImplicitDependencies: Bool
         public var runPostActionsOnFailure: Bool?
+        public var buildArchitectures: Architecture
 
         // MARK: - Init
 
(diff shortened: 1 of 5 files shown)


Sources/XcodeProj/Scheme/XCScheme+BuildAction.swift
```diff
@@ -70,12 +70,44 @@ extension XCScheme {
...
+            ///
+            /// For undefined value, initialized as `matchRunDestination`.
+            fileprivate init(_ xmlString: String) {
+                switch xmlString {
+                case "Automatic":
+                    self = .matchRunDestination
+                case "All":
+                    self = .universal
+                default:
+                    self = .matchRunDestination
```
kwridan, line 99: From testing in a Xcode, unspecified or random values default to `useTargetSettings`

```suggestion
                    self = .useTargetSettings
```

Alternatively can make this an failable init (`init?`)

Sources/XcodeProj/Scheme/XCScheme+BuildAction.swift
```diff
@@ -84,11 +116,13 @@ extension XCScheme {
                     postActions: [ExecutionAction] = [],
                     parallelizeBuild: Bool = false,
                     buildImplicitDependencies: Bool = false,
-                    runPostActionsOnFailure: Bool? = nil) {
+                    runPostActionsOnFailure: Bool? = nil,
+                    buildArchitectures: Architectures = .matchRunDestination) {
```
kwridan, line 120: This is an interesting one

- New schemes created in the Xcode UI do indeed default to `.matchRunDestination`
- The current released version of `XcodeProj` doesn't explicitly write this attribute when creating new schemes (i.e. would be `.useTargetSettings`)

Having `matchRunDestination` as the new default for the `init` here would introduce a change in behaviour for existing clients of the library and all their generated schemes would go from `.useTargetSettings` > `.matchRunDestination` 🙈

```suggestion
                    buildArchitectures: Architectures = .useTargetSettings) {
```

kwridan (commented):
nice work @woin2ee - thanks for contributing this

Sources/XcodeProj/Scheme/XCScheme+BuildAction.swift
woin2ee: The current released version didn't have control over this attribute, so I thought it wouldn't affect existing clients as if it were a new feature.
However, I missed the fact that there may be clients who are using the current version's settings well. Thank you for your review.

Sources/XcodeProj/Scheme/XCScheme+BuildAction.swift
woin2ee: That's right. It’s my mistake. 🙏

kwridan (approved):
Thanks for the updates @woin2ee

kwridan:
@all-contributors add @woin2ee for code

pepicrft (approved):

PR #827, opened 2024-07-20
feat: Introduce the new Xcode 16 models `PBXFileSystemSynchronizedRootGroup` and `PBXFileSystemSynchronizedBuildFileExceptionSet`

pepicrft:
Apple is introducing two new models, `PBXFileSystemSynchronizedRootGroup` and `PBXFileSystemSynchronizedBuildFileExceptionSet`, to reduce Git conflicts. Those types allow targets to reference a folder, whose files get automatically synchronized instead of creating references for every single file.
This PR adds those models along with the serialization logic. Once we use them from Tuist, I'll get ideas for convenient APIs that we might consider pushing down to the XcodeProj layer.


kwridan (commented):
Thanks for adding this @pepicrft 

Worth adding a fixture to test out reading / writing. 

From a quick test it seems there might be a few missing integrations to wire up the new models up as it still doesn't recognise them when attempt to read an Xcode 16 project that uses the new folder references.

```
failed: caught error: "The element PBXFileSystemSynchronizedRootGroup is not supported."
```

pepicrft:
@kwridan I uncovered the missing pieces through tests and fixed some minor issues. I'd appreciate a look when you have some spare time.

Sources/XcodeProj/Objects/Files/PBXFileSystemSynchronizedRootGroup.swift
```diff
@@ -0,0 +1,108 @@
...
+    }
+    
+    // MARK: - PlistSerializable
+    
+    override var multiline: Bool { false }
+
+    override func plistKeyAndValue(proj: PBXProj, reference: String) throws -> (key: CommentedString, value: PlistValue) {
+        var dictionary: [CommentedString: PlistValue] = try super.plistKeyAndValue(proj: proj, reference: reference).value.dictionary ?? [:]
+        dictionary["isa"] = .string(CommentedString(type(of: self).isa))
+        dictionary["exceptions"] = .array(self.exceptionsReferences.map({ exceptionReference in
```
kwridan, line 92: It seems as of Xcode 16 beta 4 the `exceptions` attribute is omitted when empty as opposed to being set as empty. Performing a read / write test I see the following diff (reading from a newly created project in Xcode 16 UI).


```diff
-               FB50EE602C5A922900C43B53 /* FolderRefs */ = {isa = PBXFileSystemSynchronizedRootGroup; explicitFileTypes = {}; explicitFolders = (); path = FolderRefs; sourceTree = "<group>"; };
+               FB50EE602C5A922900C43B53 /* FolderRefs */ = {isa = PBXFileSystemSynchronizedRootGroup; exceptions = (); explicitFileTypes = {}; explicitFolders = (); path = FolderRefs; sourceTree = "<group>"; };
```

Sources/XcodeProj/Objects/Targets/PBXTarget.swift
```diff
@@ -190,14 +210,21 @@ public class PBXTarget: PBXContainerItem {
         }
 
         dictionary["dependencies"] = .array(dependencyReferences.map { .string(CommentedString($0.value, comment: PBXTargetDependency.isa)) })
+        if let fileSystemSynchronizedGroupsReferences {
+            dictionary["fileSystemSynchronizedGroups"] = .array(fileSystemSynchronizedGroupsReferences.map { fileSystemSynchronizedGroupReference in
+                let fileSystemSynchronizedGroup: PBXFileSystemSynchronizedRootGroup? = fileSystemSynchronizedGroupReference.getObject()
+                return .string(CommentedString(fileSystemSynchronizedGroupReference.value, comment: fileSystemSynchronizedGroup?.name))
```
kwridan, line 216: When performing a read / write test of an Xcode 16 project that uses this it ends up omitting the comment string, possibly we're missing reading it?

e.g. 

```diff
                        fileSystemSynchronizedGroups = (
-                               FB50EE602C5A922900C43B53 /* FolderRefs */,
+                               FB50EE602C5A922900C43B53,
                        );
```

kwridan (commented):
thanks for the updates @pepicrft - XcodeProj now successfully reads / write a sample Xcode 16 project 🎉

I spotted a few missing bits from testing that are cause diffs after a read / write.

Sources/XcodeProj/Objects/Targets/PBXTarget.swift
pepicrft: I was accessing the wrong property.

pepicrft:
@kwridan this should be ready for another look.

Sources/XcodeProj/Objects/Targets/PBXTarget.swift
```diff
@@ -76,15 +76,28 @@ public class PBXTarget: PBXContainerItem {
     }
 
     /// Swift package product references.
-    var packageProductDependencyReferences: [PBXObjectReference]
+    var packageProductDependencyReferences: [PBXObjectReference]?
```
pepicrft, line 79: Just noticed when we introduced this attribute we introduced non-determinism. I'm fixing it in this PR.

kwridan (approved):
Awesome thank you @pepicrft 🎉

There are a few additional properties that can be set but can be done as a follow up:

- `compilerFlags`
- `platformFilters`

pepicrft:
> There are a few additional properties that can be set but can be done as a follow up:

I'll add those as part of this PR. Thanks for the review @kwridan

pepicrft:
@kwridan could you remind me of where platform filters are used? I also noticed that there's [another object type](https://github.com/tuist/XcodeProj/issues/838) to define exception for files referenced by copy build phases. I created an issue to tackle that in a follow-up PR.

PR #833, opened 2024-07-22
Update project to Swift 5.10 and start on `Sendable` updates

waltflanagan:


Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
```diff
@@ -259,7 +259,7 @@ extension XCBuildConfiguration {
     /// :nodoc:
     func isEqual(to rhs: XCBuildConfiguration) -> Bool {
         if baseConfigurationReference != rhs.baseConfigurationReference { return false }
-        if !NSDictionary(dictionary: buildSettings).isEqual(NSDictionary(dictionary: rhs.buildSettings)) { return false }
+        if buildSettings != rhs.buildSettings { return false }
```
waltflanagan, line 262: I wasnt able to find info on how to run `Sourcery` to do generate this, but this makes tests pass.  Happy to make sure the generated code is correct but will need an assist on running things.

Sources/XcodeProj/Objects/Project/PBXObjectReference.swift
```diff
@@ -1,72 +1,102 @@
...
+        }
+        set {
+            lock.withLock {
+                _objects = newValue
+            }
+        }
+    }
+    private weak var _objects: PBXObjects?
 
+    // QUESTION: this is exposed to the project but so is `getThrowingObject` and `getObject`.  What access patterns do we want to support?
```
pepicrft, line 34: Good point. I think having three APIs for the same is unnecessary. I believe (it's been a long time) that the motivation for having `getThrowingObject` was to instruct developers at runtime about bad usages of the library. For example adding a reference to an object that's not strongly referenced by the project. However, looking at the `getThrowingObject` implementation, and don't know if the error that we are throwing is more helpful than trying to do a force unwrap of the referenced object. 
I'm between making `getThrowingObject` errors more useful, which would require adding more context to `PBXObjectReference` to ensure it's in sync, which is a bit of a tricky challenge, or removing `getObject` and `getThrowingObject` and having a single accessor. I'd say that we do the latter.

Sources/XcodeProj/Objects/Sourcery/Equality.generated.swift
pepicrft: I think it's ok. Let's make those changes manually.

pepicrft (approved):
Amazing job @waltflanagan. Regarding the APIs to obtain the objects referenced by the PBXObjectReference, feel free to tackle that in a separate PR.

PR #828, opened 2024-07-20
Add Metal Diagnostics Options to LaunchAction

filipracki:
### Short description 📝
This PR adds the configuration for Metal diagnostics, allowing the configuration of these fields using Tuist.

There will be three PRs in this repository: in tuist/tuist and tuist/XcodeGraph.

PR #834, opened 2024-07-30
Fix attributes assignment in Metal Diagnostics Options of LaunchAction

filipracki:
I have made a mistake in https://github.com/tuist/XcodeProj/pull/828

### Short description 📝
In my previous PR, I mistakenly set the attributes "showGraphicsOverview" and "logGraphicsOverview" to the value of .xmlString. This was incorrect, as these two fields, unlike others, should contain values in the form of "Yes" or "No" instead of the values returned from .xmlString, which are "YES" and "NO".

Currently, I have changed it to:
```swift
if showGraphicsOverview {
    attributes["showGraphicsOverview"] = "Yes"
}
if logGraphicsOverview {
    attributes["logGraphicsOverview"] = "Yes"
}
```
Is this approach acceptable, or should we consider creating a new property, such as .xmlStringCapitalized? Alternatively, do you have a better suggestion?

This PR is related to https://github.com/tuist/tuist/pull/6537

### DIFF
--- a/Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
@@ -276,10 +276,10 @@ extension XCScheme {
                 attributes["enableGPUShaderValidationMode"] = gpuShaderValidationModeEnableValue
             }
             if showGraphicsOverview {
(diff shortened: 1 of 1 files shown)


fortmarek (approved):

fortmarek:
I will release a new patch version with this 🤞

PR #836, opened 2024-08-08
chore: set up SwiftLint and SwiftFormat, run them against the project, and run them as part of the CI workflows

pepicrft:
To ensure styling consistency and following good practices with Swift, I'm adding SwiftLint and SwiftFormat to the project.

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,3 @@
+[tools]
+swiftlint = "0.54.0"
+swiftformat = "0.54.3"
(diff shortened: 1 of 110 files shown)


.mise.toml
```diff
@@ -0,0 +1,3 @@
+[tools]
+swiftlint = "0.54.0"
+swiftformat = "0.54.3"
```
danieleformichelli, line 3: Why not latest versions?
```suggestion
swiftlint = "0.56.1"
swiftformat = "0.54.3"
```

danieleformichelli (approved):

.mise.toml
pepicrft: No particular reason :). I'll upgrade.

Issue #837, opened 2024-08-09
error: (DecodingError) keyNotFound(CodingKeys(stringValue: "compatibilityVersion", […]), […]))


F1248:
## Context 🕵️‍♀️
- Xcode version: `16.0 beta 5 (16A5221g)`
- Xcode project format: `Xcode 16.0`
- Periphery version: `2.21.0`

I'm trying to scan my project with [Periphery](https://github.com/peripheryapp/periphery) which uses XcodeProj. When I run `periphery scan` it says the key `compatibilityVersion` was not found, which is correct, my `.xcodeproj` file doesn't contain this key.
It seems that since project format `Xcode 16.0` this key should no longer be included. When I downgrade the project format to `Xcode 15.3`, it adds this key, but as soon as I revert it back to `Xcode 16.0`, it removes it again.

## What 🌱
`error: (DecodingError) keyNotFound(CodingKeys(stringValue: "compatibilityVersion", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"compatibilityVersion\", intValue: nil) (\"compatibilityVersion\").", underlyingError: nil))`

## Proposal 🎉
Don't expect the key `compatibilityVersion` if the `objectVersion` is >= `73`.

pepicrft:
Thanks @F1248 for reporting this. Seems like Xcode 16 uses a slightly different schema when `objectVersion` is above the value that you pointed out. Would you like to take a stab at fixing it? If you add me as a reviewer, I'd be happy to review it and get your changes landed

F1248:
> Thanks @F1248 for reporting this. Seems like Xcode 16 uses a slightly different schema when `objectVersion` is above the value that you pointed out. Would you like to take a stab at fixing it? If you add me as a reviewer, I'd be happy to review it and get your changes landed

@pepicrft, I'm sorry, but I can't fix this myself. I would be delighted if someone else could.

F1248:
@thanhbinh01234, thank you for fixing this!
I'll close this issue as soon as the fix is used in a released version of Periphery, so I can verify that it works for me.

kimdv:
@pepicrft @F1248 I've opened a PR and started the work :)

PR #854, opened 2024-09-17
feat: make `PBXProject.compatibilityVersion` optional and add `PBXProject.preferredProjectObjectVersion` to support Xcode 16

kimdv:
Resolves https://github.com/tuist/XcodeProj/issues/837

### Short description 📝
This PR will make `compatibilityVersion` optional. Since Xcode 16 this field isn't present anymore.
Also I added `preferredProjectObjectVersion` that is also part if Xcode 16 changes.

### Solution 📦
I'm new to this. So I tried to see if there was ny "16.0" in a fresh and newly created Xcode project, so see if it was renamed.
Could not find any. So I made it optional, so we can decode Xcode 16 projects.

### Implementation 👩‍💻👨‍💻
- Made the field optional and added `preferredProjectObjectVersion`
- Did run the branch against a local project and it worked

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProject.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProject.swift
@@ -21,7 +21,7 @@ public final class PBXProject: PBXObject {
     }
 
     /// A string representation of the XcodeCompatibilityVersion.
-    public var compatibilityVersion: String
+    public var compatibilityVersion: String?
 
     /// The region of development.
     public var developmentRegion: String?
@@ -300,7 +300,7 @@ public final class PBXProject: PBXObject {
     ///   - targetAttributes: project target's attributes.
     public init(name: String,
                 buildConfigurationList: XCConfigurationList,
-                compatibilityVersion: String,
+                compatibilityVersion: String?,
                 mainGroup: PBXGroup,
                 developmentRegion: String? = nil,
                 hasScannedForEncodings: Int = 0,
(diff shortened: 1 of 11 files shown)


kimdv:
It would be nice if I could get some help to tests?

kimdv:
I just saw that `preferredProjectObjectVersion` in the Xcode 16 is added. 
But there is no sign of it in XcodeProj. Can I add it in this PR or should it be its own?

Sources/XcodeProj/Objects/Project/PBXProject.swift
```diff
@@ -21,7 +21,10 @@ public final class PBXProject: PBXObject {
     }
 
     /// A string representation of the XcodeCompatibilityVersion.
-    public var compatibilityVersion: String
+    public var compatibilityVersion: String?
+
+    /// An string representation of the PreferredProjectObjectVersion.
+    public var preferredProjectObjectVersion: String?
```
danieleformichelli, line 27: Shouldn't this be an integer? do you have any example?

danieleformichelli (approved):

Sources/XcodeProj/Objects/Project/PBXProject.swift
kimdv: Yes here: preferredProjectObjectVersion = 77;

danieleformichelli (approved):

pepicrft (approved):

pepicrft:
@kimdv there seems to be legit issues in the changes. Would you mind looking into them?

kimdv:
> @kimdv there seems to be legit issues in the changes. Would you mind looking into them?

Should be fixed now. 
Also I have updated the commit messages. Is that right?

kimdv:
@pepicrft I see you changed the PR title? 
Should we shorten it a little if possible?

F1248:
@kimdv, @pepicrft, @danieleformichelli, would you please take a look at #861?

PR #855, opened 2024-09-17
Change order for 

kimdv:
### Short description 📝
In Xcode when writing to a project file, there was a diff in this example because it changed order.

```
/* Begin XCLocalSwiftPackageReference section */
		C39110C82C9A049500FE53C5 /* XCLocalSwiftPackageReference "../packages/Identifier" */ = {
			isa = XCLocalSwiftPackageReference;
			relativePath = "../packages/Identifier";
		};
/* End XCLocalSwiftPackageReference section */

/* Begin XCRemoteSwiftPackageReference section */
		C39110C72C9A047700FE53C5 /* XCRemoteSwiftPackageReference "swift-algorithms" */ = {
			isa = XCRemoteSwiftPackageReference;
			repositoryURL = "https://github.com/apple/swift-algorithms.git";
			requirement = {
				kind = upToNextMajorVersion;
				minimumVersion = 1.2.0;
			};
		};
/* End XCRemoteSwiftPackageReference section */
```

This example above is how it looks when it's made with Xcode.

### Solution 📦
The solution was pretty simple. Just move `XCLocalSwiftPackageReference` and `XCRemoteSwiftPackageReference` around.

### Implementation 👩‍💻👨‍💻
Just switched the code around.


danieleformichelli:
So we are just fixing that we are creating the project the same way you would create if you change it from Xcode UI? Are we fixing an actual problem or it's just to be more consistent with it?

kimdv:
There is no problem or bug. It's just to be consistent with how Xcode makes the project file.

pepicrft (approved):

kimdv:
@pepicrft not sure why it fails on Linux but not macOS?

pepicrft:
> @pepicrft not sure why it fails on Linux but not macOS?

Hey @kimdv. I'm not 100% sure. I'm doing some investigation trying to reproduce the issue locally. I'll report back once I figure out.

pepicrft:
@kimdv I could not reproduce it locally by running the tests through the Swift Docker image for version 5.10.1... So I decided to run the same script that one can run locally from macOS and that works fine.

pepicrft:
@all-contributor add @kimdv for code

Issue #861, opened 2024-09-28
`error: (DecodingError) typeMismatch(Swift.Int, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "preferredProjectObjectVersion", intValue: nil)] […]))`


F1248:
## Context 🕵️‍♀️
- Xcode 16.1 beta 2 (16B5014f)
- Xcode project format: `Xcode 16.0` (`objectVersion = 77`)
- Periphery 2.21.1
- XcodeProj 8.24.1

project.pbxproj:
```
// !$*UTF8*$!
{
	objectVersion = 77;
	objects = {
[…]
/* Begin PBXProject section */
		XXXXXXXXXXXXXXXXXXXXXXXX /* Project object */ = {
			isa = PBXProject;
			[…]
			preferredProjectObjectVersion = 77;
			[…]
		};
/* End PBXProject section */
[…]
	};
	rootObject = XXXXXXXXXXXXXXXXXXXXXXXX /* Project object */;
}
```

## What 🌱
When I try to scan my project with [Periphery](https://github.com/peripheryapp/periphery), that uses XcodeProj, with `periphery scan`, I get this error message:

`error: (DecodingError) typeMismatch(Swift.Int, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "preferredProjectObjectVersion", intValue: nil)], debugDescription: "Expected to decode Int but found a string instead.", underlyingError: nil))`

It seems that for my project, #854 doesn't work as expected.

F1248:
@pepicrft

I created [this project](https://github.com/user-attachments/files/17189695/Test.zip) newly using Xcode 16.1 beta 2 (16B5014f):
`Create New Project…` → `Multipl`→ `App`
```
% cd Test
% periphery scan --setup
Welcome to Periphery!
This guided setup will help you select the appropriate configuration for your project.

* Inspecting project...

error: (DecodingError) typeMismatch(Swift.Int, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "preferredProjectObjectVersion", intValue: nil)], debugDescription: "Expected to decode Int but found a string instead.", underlyingError: nil))
```

aksh1t:
I can reproduce this as well on a fresh iOS project created with Xcode 16 (16A242d).

pepicrft:
I opened [a PR addressing the issue](https://github.com/tuist/XcodeProj/pull/862)

aksh1t:
[...]

PR #862, opened 2024-10-01
fix: issues parsing Xcode 16 projects

pepicrft:
Resolves https://github.com/tuist/XcodeProj/issues/861

This PR contains some Xcode 16-related fixes:
- We were trying to parse `PBXProject.preferredProjectObjectVersion` as an integer but in Xcode 16 is encoded as a string. I'm decoding it as a string and converting it to an integer. If Apple supports strings that don't represent integers down the road we'll have to change the internal type.
- They added the attribute `minimizedProjectReferenceProxies` to the `PBXProject` object.
- `SynchronizedRootGroups` are encoded using multiple-lines. I adjusted the internal model to reflect that and ensure there are no diffs when we write the project back.

### DIFF
--- a/Sources/XcodeProj/Objects/Files/PBXFileSystemSynchronizedRootGroup.swift
+++ b/Sources/XcodeProj/Objects/Files/PBXFileSystemSynchronizedRootGroup.swift
@@ -2,108 +2,113 @@ import Foundation
 import PathKit
 
 public class PBXFileSystemSynchronizedRootGroup: PBXFileElement {
-    /// It maps relative paths inside the synchronized root group to a particular file type.
-    /// If a path doesn't have a particular file type specified, Xcode defaults to the default file type
-    /// based on the extension of the file.
-    public var explicitFileTypes: [String: String]
-
-    /// Returns the references of the exceptions.
-    var exceptionsReferences: [PBXObjectReference]?
-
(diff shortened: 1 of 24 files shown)


Sources/XcodeProj/Objects/Files/PBXFileSystemSynchronizedRootGroup.swift
```diff
@@ -2,108 +2,113 @@ import Foundation
...
+  /// If a path doesn't have a particular file type specified, Xcode defaults to the default file type
+  /// based on the extension of the file.
+  public var explicitFileTypes: [String: String]?
+  
+  /// Returns the references of the exceptions.
+  var exceptionsReferences: [PBXObjectReference]?
+  
+  /// It returns a list of exception objects that override the configuration for some children
+  /// in the synchronized root group.
+  public var exceptions: [PBXFileSystemSynchronizedBuildFileExceptionSet]? {
```
kimdv, line 15: Is seems that the indentation is changed?

pepicrft: Yeah. I wonder I just ran `swiftformat`. I'll check what happened.

pepicrft:
@tuist/community-core this is ready for a review 👀

PR #865, opened 2024-10-09
fix: Inconsistent behaviour with Xcode 16 when `PBXProject.TargetAttributes` is empty 

Speakus:
### Short description 📝

stop generate empty TargetAttributes - because Xcode 16 will always remove such item

### Solution 📦

very simple fix

NB! Better to wait for merging this fix first: https://github.com/fastlane/fastlane/pull/23872

PR #867, opened 2024-10-10
sort project references as xcode16

Speakus:
### Short description 📝

If project saved with XcodeProj then opening it with Xcode 16 generate some changes. So we could respect Xcode 16 sorting by sorting our result with the same way.

### Solution 📦

Do the same sorting as Xcode 16 do

----

NB! This PR based on other fix so please merge this PR first https://github.com/tuist/XcodeProj/pull/865

### DIFF
--- a/Sources/XcodeProj/Objects/Project/PBXProjEncoder.swift
+++ b/Sources/XcodeProj/Objects/Project/PBXProjEncoder.swift
@@ -476,39 +476,16 @@ final class PBXProjEncoder {
         }
 
         for project in projects {
-            do {
-                project.projectReferences = try project.projectReferences.sorted(by: { lhs, rhs in
-                    guard let lProjectRef = lhs["ProjectRef"] else {
-                        throw Errors.unexpectedPbxProj()
-                    }
-                    guard let lFile: PBXFileElement = lProjectRef.getObject() else {
-                        throw Errors.unexpectedPbxProj()
-                    }
-                    guard let rProjectRef = rhs["ProjectRef"] else {
-                        throw Errors.unexpectedPbxProj()
-                    }
-                    guard let rFile: PBXFileElement = rProjectRef.getObject() else {
-                        throw Errors.unexpectedPbxProj()
-                    }
-                    guard let lName = lFile.name else {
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Objects/Project/PBXProjEncoder.swift
```diff
@@ -468,4 +469,23 @@ final class PBXProjEncoder {
...
     }
+
+    private func sortProjectReferences(for projects: [PBXProject], outputSettings: PBXOutputSettings) {
+        guard outputSettings.projReferenceFormat == .xcode else {
+            return
+        }
+
+        for project in projects {
+            /// The project references are sorted alphabetically based on the name of the project it's being referenced.
+            project.projectReferences = project.projectReferences.sorted(by: { lhs, rhs in
```
pepicrft, line 480: @Speakus in which scenario a project contains more than one project reference. Would you mind adding a fixture under `Fixtures` so that we can add some tests?

pepicrft (changes requested):
Let's add some tests to ensure no regressions are introduced in this area.

Sources/XcodeProj/Objects/Project/PBXProjEncoder.swift
mikhailmulyar: yes, will try to provide some example

mikhailmulyar: Here it is, I've added example with wrong ordering and some tests for checking.
https://github.com/mikhailmulyar/XcodeProj/commit/4f282f1d31357f6f3f932f445f61b747022e8e3a

pepicrft: @mikhailmulyar would you mind completing this PR? It'll require closing this one and opening a new one.

mikhailmulyar: @pepicrft Hi. Sorry for delay. Here is update PR with test added https://github.com/tuist/XcodeProj/pull/937

PR #871, opened 2024-10-17
Add handling for Swift Testing Only Parallelization

kelvinharron:
Related to [this Tuist issue](https://github.com/tuist/tuist/issues/6827)

### Short description 📝
This PR introduces XcodeProj Parallelization handling for the new Swift Testing option. This will enable us to expose the option to Tuist `TestingOptions` allowing users to specify the option they need, as the default handling in Xcode 16 for Parallelization is going to benefit the few users who have adopted Swift Testing.

I also removed two dead links from the References section as I was going through the reading material.

### Solution 📦
From changing the values manually in a real project, I gathered:

> Parallelization off in Xcode ->"parallelizable" : false,
Parallelization on in Xcode ->"parallelizable" : true,
Parallelization Swift Testing Only in Xcode -> no value.

With guidance from @fortmarek , I updated the handling of the attributes to cover each individual case. 

I've set it as draft as I need guidance on how to prove this handling outside of unit tests. Please let me know what I could be missing and how to verify the output. Thanks.

### Implementation 👩‍💻👨‍💻

- [x] Introduce enums to cover the three unique cases
- [x] Add handling for each case from.
- [x] Update tests to handle each case.

### DIFF
--- a/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
@@ -6,7 +6,7 @@ public extension XCScheme {
         // MARK: - Attributes
 
         public var skipped: Bool
-        public var parallelizable: Bool
+        public var parallelization: Parallelization
         public var randomExecutionOrdering: Bool
         public var useTestSelectionWhitelist: Bool?
         public var buildableReference: BuildableReference
@@ -17,7 +17,7 @@ public extension XCScheme {
         // MARK: - Init
 
         public init(skipped: Bool,
-                    parallelizable: Bool = false,
+                    parallelization: Parallelization = .none,
                     randomExecutionOrdering: Bool = false,
                     buildableReference: BuildableReference,
                     locationScenarioReference: LocationScenarioReference? = nil,
@@ -25,7 +25,7 @@ public extension XCScheme {
                     selectedTests: [TestItem] = [],
                     useTestSelectionWhitelist: Bool? = nil) {
             self.skipped = skipped
-            self.parallelizable = parallelizable
+            self.parallelization = parallelization
             self.randomExecutionOrdering = randomExecutionOrdering
             self.buildableReference = buildableReference
             self.locationScenarioReference = locationScenarioReference
@@ -36,7 +36,14 @@ public extension XCScheme {
 
         init(element: AEXMLElement) throws {
             skipped = element.attributes["skipped"] == "YES"
-            parallelizable = element.attributes["parallelizable"] == "YES"
+          
+              if let parallelizableValue = element.attributes["parallelizable"] {
+                parallelization = parallelizableValue == "YES" ? .all : .none
+              } else {
+                  parallelization = .swiftTestingOnly
+              }
+          
+          
             useTestSelectionWhitelist = element.attributes["useTestSelectionWhitelist"] == "YES"
             randomExecutionOrdering = element.attributes["testExecutionOrdering"] == "random"
             buildableReference = try BuildableReference(element: element["BuildableReference"])
@@ -63,7 +70,16 @@ public extension XCScheme {
 
         func xmlElement() -> AEXMLElement {
             var attributes: [String: String] = ["skipped": skipped.xmlString]
-            attributes["parallelizable"] = parallelizable ? parallelizable.xmlString : nil
+            
+            switch parallelization {
+            case .all:
+                attributes["parallelizable"] = "YES"
+            case .none:
+                attributes["parallelizable"] = "NO"
+            case .swiftTestingOnly:
+                break // SwiftTesting is inferred by the lack of a value
+            }
+          
             if let useTestSelectionWhitelist {
                 attributes["useTestSelectionWhitelist"] = useTestSelectionWhitelist.xmlString
             }
@@ -99,7 +115,7 @@ public extension XCScheme {
 
         public static func == (lhs: TestableReference, rhs: TestableReference) -> Bool {
             lhs.skipped == rhs.skipped &&
-                lhs.parallelizable == rhs.parallelizable &&
+                lhs.parallelization == rhs.parallelization &&
                 lhs.randomExecutionOrdering == rhs.randomExecutionOrdering &&
                 lhs.buildableReference == rhs.buildableReference &&
                 lhs.locationScenarioReference == rhs.locationScenarioReference &&
--- a/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
@@ -37,12 +37,11 @@ public extension XCScheme {
         init(element: AEXMLElement) throws {
             skipped = element.attributes["skipped"] == "YES"
           
-              if let parallelizableValue = element.attributes["parallelizable"] {
-                parallelization = parallelizableValue == "YES" ? .all : .none
-              } else {
-                  parallelization = .swiftTestingOnly
-              }
-          
+            if let parallelizableValue = element.attributes["parallelizable"] {
+              parallelization = parallelizableValue == "YES" ? .all : .none
+            } else {
+                parallelization = .swiftTestingOnly
+            }
           
             useTestSelectionWhitelist = element.attributes["useTestSelectionWhitelist"] == "YES"
             randomExecutionOrdering = element.attributes["testExecutionOrdering"] == "random"
--- a/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
@@ -38,7 +38,7 @@ public extension XCScheme {
             skipped = element.attributes["skipped"] == "YES"
           
             if let parallelizableValue = element.attributes["parallelizable"] {
-              parallelization = parallelizableValue == "YES" ? .all : .none
+                parallelization = parallelizableValue == "YES" ? .all : .none
             } else {
                 parallelization = .swiftTestingOnly
             }
--- a/Tests/XcodeProjTests/Scheme/XCSchemeTests.swift
+++ b/Tests/XcodeProjTests/Scheme/XCSchemeTests.swift
@@ -68,7 +68,7 @@ final class XCSchemeIntegrationTests: XCTestCase {
     func test_write_testableReferenceDefaultAttributesValuesAreOmitted() {
         let reference = XCScheme.TestableReference(
             skipped: false,
-            parallelizable: false,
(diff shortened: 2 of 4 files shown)


Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -17,15 +17,15 @@ public extension XCScheme {
...
 
         public init(skipped: Bool,
-                    parallelizable: Bool = false,
+                    parallelization: Parallelization = .none,
                     randomExecutionOrdering: Bool = false,
                     buildableReference: BuildableReference,
                     locationScenarioReference: LocationScenarioReference? = nil,
                     skippedTests: [TestItem] = [],
                     selectedTests: [TestItem] = [],
                     useTestSelectionWhitelist: Bool? = nil) {
```
fortmarek, line 26: I believe we'd need to create another version of `init` that accepts `parallelizable` as a boolean, so this doesn't cause a breaking change.

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -6,7 +6,7 @@ public extension XCScheme {
         // MARK: - Attributes
 
         public var skipped: Bool
-        public var parallelizable: Bool
+        public var parallelization: Parallelization
```
fortmarek, line 9: I'd probably keep the name of this property `parallelizable` to keep aligned with the `pbxproj` format.

Tests/XcodeProjTests/Scheme/XCSchemeTests.swift
```diff
@@ -106,12 +106,34 @@ final class XCSchemeIntegrationTests: XCTestCase {
         XCTAssertEqual(subject.attributes["useTestSelectionWhitelist"], "YES")
         XCTAssertEqual(subject.attributes["testExecutionOrdering"], "random")
     }
+  
+  func test_write_testableReferenceAttributesValuesForSwiftTesting() {
```
fortmarek, line 110: Let's also add a test for reading the `.swiftTestingOnly` case (i.e., when `parallelizable` is not specified)?

fortmarek (commented):
thanks for the PR! Some minor comments, but overall looks good 🤞

Tests/XcodeProjTests/Scheme/XCSchemeTests.swift
kelvinharron: Thanks for the prompt on the read tests. I've added one that is covered implicitly by the other reads, as the provided files do not have any.  I've also added cases on the writes for all 3 enum based inits + 2 bool init cases. Would it be worth creating another test file `XCSchemeTests+Parallelizable.swift` to cover all the cases here? I feel like I'm muddying the existing one.

Sources/XcodeProj/Scheme/XCScheme+Parallelizable.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+public enum Parallelizable: String {
```
fortmarek, line 3: I'd declare this in an `XCScheme` extension and add some basic docs for what this `enum` represents/is supposed to be used for.

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -33,10 +33,34 @@ public extension XCScheme {
             self.selectedTests = selectedTests
             self.skippedTests = skippedTests
         }
+      
+        public init(skipped: Bool,
```
fortmarek, line 37: Let's mark this `init` deprecated with a suggestion how to migrate.

Tests/XcodeProjTests/Scheme/XCSchemeTests.swift
```diff
@@ -58,6 +58,12 @@ final class XCSchemeIntegrationTests: XCTestCase {
         }
     }
 
+    func test_read_iosScheme_noParallelizable_setsswiftTestingOnly() throws {
+        let scheme = try XCScheme(path: iosSchemePath)
+
+        XCTAssertEqual(scheme.testAction?.testables.first?.parallelizable, .swiftTestingOnly)
+    }
```
fortmarek, line 65: I haven't realized this is tested [here](https://github.com/tuist/XcodeProj/blob/main/Tests/XcodeProjTests/Scheme/XCSchemeTests.swift#L368) – I think we can remove this test case.

fortmarek (commented):
Small nits, after those are resolved, we can merge this 🤞

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
kelvinharron: Let me know how that reads, as it follows the usages found elsewhere in the codebase.

fortmarek (approved):
Looks good! Before we merge, can you open a draft PR that updates `XcodeGraph` in `tuist/tuist` with this branch, so we can test things E2E there?

fortmarek:
@kelvinharron we can merge this one once the CI is green ✅

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -6,7 +6,7 @@ public extension XCScheme {
         // MARK: - Attributes
 
         public var skipped: Bool
-        public var parallelizable: Bool
+        public var parallelizable: Parallelizable
```
kwridan, line 9: This may constitute as breaking change :/ 

Options:

- Introduce a new type `parallelizableValue: Parallelizable` (or some better name 😅) and keep `parallelizable: Bool` as a Bool + mark deprecated - it can translate the to  and from underlying enum property 
- Ensure a major version is released of XcodeProj

Interesting ideas but may not be sufficient:

- Leverage `ExpressibleByBooleanLiteral` https://developer.apple.com/documentation/swift/expressiblebybooleanliteral - that may mitigate any literal usage but will still break any clients using a boolean variable or reading this boolean variable :/

kwridan (approved):
Thanks @kelvinharron

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
fortmarek: I think it's worth investigating using the `ExpressibleByBooleanLiteral` to make the migration easier. Technically speaking, it won't be sufficient, so let's make a new major version of `XcodeProj`. I think for this library, having major versions more often than let's say for the Tuist CLI is fine.

kwridan: +1 this gets it a bit closer to the nomenclature in the Xcode UI

Screenshot 2024-11-28 at 08 09 46

freak4pc: Hey @kelvinharron - what do you think about this ? 
If you want an extra hand I'm happy to finalize this PR so we can land it into XcodeProj. LMK! 
Thanks :)

kelvinharron: Hey folks, I can apply these changes shortly. We're saying that we'll include the boolean but deprecate the field, would we still intend to keep the deprecated init as well?

kwridan: > Hey folks, I can apply these changes shortly. We're saying that we'll include the boolean but deprecate the field

Thanks @kelvinharron 

> Would we still intend to keep the deprecated init as well?

Ya we'll need to avoid the breaking change seeing the purpose of those additional changes is backwards compatibility

kelvinharron: Thanks for the feedback @kwridan, keep me right on the most recent changes as it feels a little loose with two properties. This PR extends more so than replaces now, so we're handling both mapping functions from attributes and inits as before, with the introduction of TestParallelization option.

kelvinharron: Ace @freak4pc. I've updated and added some tests to reason about the computed var behaviour. Please take a look and let me know if that's expected.

freak4pc (approved):
LGTM! I don't have merge / write access but took a quick peek anyways :)

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
kelvinharron: @kwridan @fortmarek let me know if you have any additional feedback and I can close this out. I think we're encountering some failures on CI with these changes I can look at.

freak4pc:
Anything preventing us from moving forward? Would love to have swift-testing at our disposal :D 
@kwridan @fortmarek

Sources/XcodeProj/Scheme/XCScheme+TestableReference.swift
```diff
@@ -16,6 +22,25 @@ public extension XCScheme {
...
+            self.buildableReference = buildableReference
+            self.locationScenarioReference = locationScenarioReference
+            self.useTestSelectionWhitelist = useTestSelectionWhitelist
+            self.selectedTests = selectedTests
+            self.skippedTests = skippedTests
+        }
+
+        @available(*, deprecated, message: "Use init with Parallelization argument instead")
         public init(skipped: Bool,
                     parallelizable: Bool = false,
```
fortmarek, line 45: Let's remove the default value, so the deprecation warning is triggered only when `parallelizable` is specified.

fortmarek (approved):
This looks good! I'd wait with the merge until the Tuist PR counterpart is approved as well: https://github.com/tuist/tuist/pull/6898

kelvinharron:
> This looks good! I'd wait with the merge until the Tuist PR counterpart is approved as well: [tuist/tuist#6898](https://github.com/tuist/tuist/pull/6898)

Thanks Marek. I still haven't found a way to reliably prove out the tests for that PR meaning its a non starter. @freak4pc care to look at the PR and see if you have any ideas? [Some more context on Slack here.](https://tuist-community.slack.com/archives/CTB0DBVD2/p1731185862534779?thread_ts=1731185721.436619&cid=CTB0DBVD2)

fortmarek:
> Thanks Marek. I still haven't found a way to reliably prove out the tests for that PR meaning its a non starter. @freak4pc care to look at the PR and see if you have any ideas?

I don't have extra ideas apart from what I posted. Feel free to simplify the acceptance test if needed.

However, I think we can merge this one without having a Tuist acceptance test in place once the CI is ✅

pepicrft:
@kelvinharron your changes are now part of [XcodeProj 8.25.0](https://github.com/tuist/XcodeProj/releases/tag/8.25.0)

PR #881, opened 2024-11-12
refactor: align ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES with the Xcode default

yungu0010:
Resolves https://github.com/tuist/tuist/issues/6989

### Short description 📝
I modified BuildSettingsProvider.swift to resolve issue 6989 of tuist/tuist.
Removed the settings for ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES so that tuist always followed the default settings of Xcode.
Please see [this issue](https://github.com/tuist/tuist/issues/6989).

### Solution 📦
tuist/tuist depends on local XcodeProj. Modified local XcodeProj and then test at fixtures/workspace_with_test_locale.

### Implementation 👩‍💻👨‍💻

- [ ] ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES is set to xcode default setting when watchOS application

### DIFF
--- a/Sources/XcodeProj/Utils/BuildSettingsProvider.swift
+++ b/Sources/XcodeProj/Utils/BuildSettingsProvider.swift
@@ -72,10 +72,6 @@ public class BuildSettingsProvider {
             buildSettings.merge(targetSwiftSettings(product: product), uniquingKeysWith: { $1 })
         }
 
-        if let platform, let product, let swift, swift == true {
-            buildSettings.merge(targetSwiftSettings(platform: platform, product: product), uniquingKeysWith: { $1 })
-        }
-
         return buildSettings
     }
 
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Utils/BuildSettingsProvider.swift
```diff
@@ -373,10 +373,6 @@ public class BuildSettingsProvider {
 
     private static func targetSwiftSettings(platform: Platform, product: Product) -> BuildSettings {
```
fortmarek, line 374: I believe the method can be removed completely now

yungu0010: @fortmarek I agree. Removed it!

PR #883, opened 2024-11-25
refactor: Update PathRunnable so that it subclasses Runnable

georgenavarro:
### Short description 📝

PathRunnable currently does not subclass from Runnable like BuildableProductRunnable and RemoteRunnable do. This means that even if you don't care about the specific Runnable implementation, you still have to handle PathRunnable separately from the Runnable subclasses. For example, you would either need two different functions:

```
func doSomething(with runnable: Runnable)
func doSomething(with runnable: PathRunnable)
```

or a function that takes in two different parameters:

```
func doSomething(with runnable: Runnable?, or runnable: PathRunnable?)
```

### Solution 📦

Refactor PathRunnable so that it subclasses from Runnable. This will allow any var or function parameter of type Runnable to hold an instance of PathRunnable. For example the following function would now be able to accept a PathRunnable as an argument in addition to a BuildableProductRunnable or RemoteRunnable:

```
func doSomething(with runnable: Runnable)
```

Since the only changes is to make PathRunnable subclass Runnable, this should be a non breaking change. No properties or functions are being added, deleted.

### Implementation 👩‍💻👨‍💻

- [x] Made the buildableReference parameter of Runnable's init an optional parameter. The buildableReference property of Runnable was already optional so now the types match. Since PathRunnable does not use a  BuildableReference, this change will allow PathRunnable to call Runnable's init with a nil BuildableReference argument.
- [x] Changed PathRunnable to subclass from Runnable like BuildableProductRunnable and RemoteRunnable do.
- [x] Removed PathRunnable's declared conformance to Equatble since it now inherits that from Runnable.
- [x] Removed PathRunnable's runnableDebuggingMode property since it now inherits that from Runnable.
- [x] Refactored the functions that PathRunnable now overrides from Runnable. I used RemoteRunnable as a model for how to accomplish this.
- [x] Removed the pathRunnable property from LaunchAction now that its runnable var can hold an instance of PathRunnable. This avoids the possiblity of having two different runnables, which didn't really make sense. If Xcode detects two runnables, it updates the scheme's launch action so that it only keeps one and deletes the other.
- [x] Updated the serialization test case for PathRunnable.


Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
```diff
@@ -42,7 +42,6 @@ public extension XCScheme {
         public var buildConfiguration: String
         public var launchStyle: Style
         public var askForAppToLaunch: Bool?
-        public var pathRunnable: PathRunnable?
```
kwridan, line 45: we may need a major release of XcodeProj as this (and the initialiser) would be constitute a breaking change

Alternatively 

```swift
public var pathRunnable: PathRunnable? {
   runnable as? PathRunnable
}
```

Sources/XcodeProj/Scheme/XCScheme+PathRunnable.swift
```diff
@@ -3,41 +3,47 @@ import Foundation
...
+            }
+
+            return super.isEqual(other: other) &&
+                filePath == other.filePath
+        }
+
         public static func == (lhs: PathRunnable, rhs: PathRunnable) -> Bool {
-            lhs.runnableDebuggingMode == rhs.runnableDebuggingMode &&
-                lhs.filePath == rhs.filePath
+            lhs.isEqual(other: rhs) && rhs.isEqual(other: lhs)
```
kwridan, line 46: Are both checks needed? Would one suffice here?

```suggestion
            lhs.isEqual(other: rhs) 
```

kwridan (commented):
Thanks for contributing this @georgenavarro 

Looking at the original PR that introduced `PathRunnable` there doesn't seem to be any specific reason why this wouldn't have been a subclass of `Runnable` most likely accidental. As far as I can tell there can only be one `Runnable`, one of which is `PathRunnable` so makes sense to consolidate this.

https://github.com/tuist/XcodeProj/pull/521

That said we'll need a few small tweaks to ensure non breaking changes, perhaps with some deprecation annotations which can be cleaned up in the next major release.

Sources/XcodeProj/Scheme/XCScheme+PathRunnable.swift
```diff
@@ -3,41 +3,47 @@ import Foundation
 import PathKit
 
 public extension XCScheme {
-    class PathRunnable: Equatable {
+    class PathRunnable: Runnable {
```
kwridan, line 6: Making `PathRunnable` subclass `Runnable` means we'll also need to update [ProfileAction](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Scheme/XCScheme%2BProfileAction.swift#L35) to support it

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
georgenavarro: I was so focused on making sure my changes to PathRunnable were non breaking that I forgot to consider my changes to LaunchAction 🤦🏻‍♂️. I'll add this computed var you suggested.

Sources/XcodeProj/Scheme/XCScheme+PathRunnable.swift
georgenavarro: Yes, I believe your change would be sufficient to check for equality. I did it this way because RemoteRunnable had the same implementation so I thought I should follow the pattern. I'll happily make this change as well. 👍🏼

Sources/XcodeProj/Scheme/XCScheme+PathRunnable.swift
georgenavarro: I overlooked this one. I'll try to add this as well.

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
```diff
@@ -161,6 +162,93 @@ public extension XCScheme {
...
+            language: String? = nil,
+            region: String? = nil,
+            showNonLocalizedStrings: Bool = false,
+            launchAutomaticallySubstyle: String? = nil,
+            storeKitConfigurationFileReference: StoreKitConfigurationFileReference? = nil,
+            customLaunchCommand: String? = nil,
+            customLLDBInitFile: String? = nil
+        ) {
+            self.init(
+                runnable: pathRunnable ?? runnable,
```
georgenavarro, line 209: I also realized that I removed PathRunnable as a parameter in LaunchAction's init method which would also be a breaking change.  I created a convenience init method which adds it back for backwards compatibility. I chose to have it call the designated initializer with it's pathRunnable argument nil coalesced to other runnable argument since it is the more specialized one.

kwridan: Thanks @georgenavarro 

Two small suggestions

* Perhaps we can mark it as deprecated and a message to use `init(runnable: pathRunnable)`
* We can make the convenience `init` have a `pathRunnable` without a default to help the compiler disambiguate (and to flag the deprecation warning)

kwridan (commented):
thanks for the updates @georgenavarro

kwridan:
@all-contributors add @georgenavarro for code

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
pepicrft: I pushed a commit addressing this feedback. Thanks @kwridan

pepicrft (approved):

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
```diff
@@ -43,9 +43,15 @@ public extension XCScheme {
...
         public var askForAppToLaunch: Bool?
         public var pathRunnable: PathRunnable? {
-          // For backwards compatibility
-          runnable as? PathRunnable
+            // For backwards compatibility
+            get {
+                runnable as? PathRunnable
+            }
+            set {
+                self.pathRunnable = newValue
```
georgenavarro, line 51: @pepicrft Looks like this was supposed to be `self.runnable = newValue`. Xcode is warning about infinite recursion since it is essentially calling it's own setter.

Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
```diff
@@ -81,6 +87,7 @@ public extension XCScheme {
 
         // MARK: - Init
 
+        @available(*, deprecated, message: "Use the init() that consolidates pathRunnable and runnable into a single parameter.")
```
georgenavarro, line 90: @pepicrft I think this deprecation warning was attached to the wrong init since the other convenience init now only accepts a PathRunnable

georgenavarro:
I've opened a new [PR](https://github.com/tuist/XcodeProj/pull/889) to fix the issues I mentioned above.

pepicrft:
Ouch, thanks for opening a PR for those issues. Let me review it and get it merged.

PR #889, opened 2024-12-08
fix: Infinite recursion and Incorrect deprecation notice in PathRunnable

georgenavarro:
### Short description 📝
A recent [pull request](https://github.com/tuist/XcodeProj/pull/883) that modified `LaunchAction` introduced two issues into the class:
1. The setter for the `pathRunnable` property causes infinite recursion
2. The deprecation notice was attached to the wrong init

### Solution 📦
1. I fixed the infinite recursion in the `pathRunnable` setter by having it modify the backing `runnable` property instead of having it accidentally call itself.
2. I moved the deprecation notice to the init that still accepts a `PathRunnable` and removed it from the init that accepts a generic `Runnable`.

### DIFF
--- a/Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
+++ b/Sources/XcodeProj/Scheme/XCScheme+LaunchAction.swift
@@ -48,7 +48,7 @@ public extension XCScheme {
                 runnable as? PathRunnable
             }
             set {
-                self.pathRunnable = newValue
+                self.runnable = newValue
             }
         }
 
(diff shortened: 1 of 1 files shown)


kwridan (commented):
thanks for catching this @georgenavarro - I missed this during review 🙈

kwridan (approved):
thanks @georgenavarro

pepicrft (approved):

pepicrft:
The installation of Swift in the Linux environment is failing. Since I checked that test and build passes locally, I'll go ahead with the merge and investigate that issue separately.
Thanks a lot @georgenavarro once again for your contributions :)

PR #892, opened 2024-12-20
feat: add path to XcodeProj and XCWorkspace

ajkolean:
Resolves https://github.com/tuist/XcodeGraph/pull/87#discussion_r1892498684

### Short description 📝
This PR introduces an optional `projectPath` property to `XcodeProj`, allowing the project to be initialized and managed without requiring a file system path. It maintains full backward compatibility for existing users while offering more flexibility for in-memory project constructions or integrations with path management logic outside of `XcodeProj`.

### Solution 📦
The `projectPath` property is now optional. When loading from a file system path, `projectPath` is set automatically. If users create `XcodeProj` instances in memory using the new initializer, `projectPath` can remain `nil`. This approach preserves existing behavior while enabling new use cases, such as working with `XcodeProj` in environments where a project might not yet exist on disk.

### Implementation 👩‍💻👨‍💻
- [x] Introduce `projectPath: Path?` as an optional property in `XcodeProj`.
- [x] Update `init(path:)` to set `projectPath` when loading from disk.
- [x] Add a new initializer that allows creating `XcodeProj` instances without a `projectPath`.
- [x] Include `projectPath` in the equality check to distinguish between projects referencing different directories.
- [x] Retain backward compatibility and do not alter existing file loading or writing behavior.

### DIFF
--- a/Sources/XcodeProj/Project/XcodeProj.swift
+++ b/Sources/XcodeProj/Project/XcodeProj.swift
@@ -8,8 +8,8 @@ public final class XcodeProj: Equatable {
     /// Project workspace
     public var workspace: XCWorkspace
 
-    /// The  path to the `.xcodeproj` directory.
-    private let projectPath: Path?
+    /// The path to the `.xcodeproj` directory.
+    public let path: Path?
 
     /// .pbxproj representation
     public var pbxproj: PBXProj
@@ -46,7 +46,7 @@ public final class XcodeProj: Equatable {
             .glob("*.xcuserdatad")
             .compactMap { try? XCUserData(path: $0) }
 
-        projectPath = path
+        self.path = path
         self.pbxproj = pbxproj
         self.workspace = workspace
         self.sharedData = sharedData
@@ -68,12 +68,12 @@ public final class XcodeProj: Equatable {
                 pbxproj: PBXProj,
                 sharedData: XCSharedData? = nil,
                 userData: [XCUserData] = [],
-                projectPath: Path? = nil) {
+                path: Path? = nil) {
         self.workspace = workspace
         self.pbxproj = pbxproj
         self.sharedData = sharedData
         self.userData = userData
-        self.projectPath = projectPath
+        self.path = path
     }
 
     // MARK: - Equatable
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Project/XcodeProj.swift
```diff
@@ -8,6 +8,9 @@ public final class XcodeProj: Equatable {
     /// Project workspace
     public var workspace: XCWorkspace
 
+    /// The  path to the `.xcodeproj` directory.
+    private let projectPath: Path?
```
fortmarek, line 12: Do we want to make it public, so it can be accessed outside of this module?

Sources/XcodeProj/Project/XcodeProj.swift
```diff
@@ -8,6 +8,9 @@ public final class XcodeProj: Equatable {
     /// Project workspace
     public var workspace: XCWorkspace
 
+    /// The  path to the `.xcodeproj` directory.
```
fortmarek, line 11:
```suggestion
    /// The path to the `.xcodeproj` directory.
```

Sources/XcodeProj/Project/XcodeProj.swift
fortmarek, line 12:
```suggestion
    private let path: Path?
```

I'd align with the value of the `init` parameter

fortmarek (commented):
I'd also add `path` to the `XCWorkspace`: https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Workspace/XCWorkspace.swift

I'll leave it up to you if you wanna do it in this PR or in a subsequent one.

Sources/XcodeProj/Project/XcodeProj.swift
ajkolean: 🤦 Yes that was the whole point, earth to Andy lol

fortmarek (approved):

fortmarek (approved):

fortmarek:
We also really need to migrate the paths from `PathKit` to our Path. The needed conversion between the paths is unnecessary pain in the `XcodeGraphMapper`

Sources/XcodeProj/Project/XcodeProj.swift
fortmarek: haha, could happen to anyone 😂

PR #897, opened 2025-01-16
fix: ensure that file references are fixed for filesystem synchronized root groups

bryansum:
### Short description 📝
Noticed that when building projects with the new filesystem synchronized root groups that I wasn't seeing references fixed correctly. This seems like an oversight to #827, which had tests for decoding / encoding and for equality testing but none for this in particular.

### Solution 📦
This adds an additional check to `ReferenceGenerator` for any filesystem synchronized root groups starting from the main group, or recursively any children.

### Implementation 👩‍💻👨‍💻

- [x] Implemented a change to `ReferenceGenerator` fixing any filesystem synchronized root group object references 
- [x] Added a new test to `ReferenceGeneratorTests` to ensure that this is working correctly.


kwridan (approved):
Thanks for spotting and fixing this @bryansum

fortmarek (approved):
Nice catch! Looks good 🙂 

Once the lint checks are resolved, we can merge this ^^

fortmarek:
Thanks again for this PR! 🙌

PR #898, opened 2025-01-20
fix: add missing BuildSettingsProvider for visionOS

alexanderwe:
### Short description 📝

`Tuist` introduced `XcodeGraph` at version 0.19.4 with this PR https://github.com/tuist/tuist/pull/7083. This in fact caused this PR to take effect https://github.com/tuist/XcodeGraph/pull/79. This then ultimately caused this codepath https://github.com/tuist/tuist/blob/main/Sources/TuistGenerator/Settings/DefaultSettingsProvider.swift#L115 to not include `visionOS` related default settings.

### Solution 📦
I added the mapping of `visionOS` to the `BuildSettingsProvider` in order to also correctly set the settings for `visionOS`.

### Implementation 👩‍💻👨‍💻
- [x] Add `visionOS` mappings to `BuildSettingsProvider`
- [x] Update `BuildSettingProviderTests` to test the newly added mappings

### Questions 

- @fortmarek / @pepicrft I was not quite sure if we should also set `CODE_SIGN_IDENTITY` on the `visionOS` platform in order to be aligned with the `iOS` one - what do you think ? 
- Please let me know if the explanation is good enough or if I missed something or have invalid reasoning.

### DIFF
--- a/Sources/XcodeProj/Utils/BuildSettingsProvider.swift
+++ b/Sources/XcodeProj/Utils/BuildSettingsProvider.swift
@@ -247,6 +247,10 @@ public class BuildSettingsProvider {
                 "SKIP_INSTALL": "YES",
                 "LD_RUNPATH_SEARCH_PATHS": ["$(inherited)", "@executable_path/Frameworks"],
             ]
+        case (.visionOS, .application):
+            [
+                "LD_RUNPATH_SEARCH_PATHS": ["$(inherited)", "@executable_path/Frameworks"],
+            ]
         case (.iOS, .framework):
             [
                 "LD_RUNPATH_SEARCH_PATHS": ["$(inherited)", "@executable_path/Frameworks", "@loader_path/Frameworks"],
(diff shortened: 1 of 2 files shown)


Sources/XcodeProj/Utils/BuildSettingsProvider.swift
```diff
@@ -266,6 +270,10 @@ public class BuildSettingsProvider {
                 "APPLICATION_EXTENSION_API_ONLY": "YES",
                 "LD_RUNPATH_SEARCH_PATHS": ["$(inherited)", "@executable_path/Frameworks", "@loader_path/Frameworks"],
             ]
+        case (.visionOS, .framework):
+            [
+                "LD_RUNPATH_SEARCH_PATHS": ["$(inherited)", "@executable_path/Frameworks", "@loader_path/Frameworks"],
+            ]
         case ([.iOS, .tvOS, .watchOS], .staticLibrary):
```
kwridan, line 277: I suspect visionOS is needed in a few other cases such this one (and the app extensions one too)

```suggestion
        case ([.iOS, .tvOS, .watchOS, .visionOS], .staticLibrary):
```

kwridan (approved):
Thanks @alexanderwe

pepicrft:
> @fortmarek / @pepicrft I was not quite sure if we should also set CODE_SIGN_IDENTITY on the visionOS platform in order to be aligned with the iOS one - what do you think ?

I think that makes sense. Thoughts @fortmarek ?

fortmarek:
> I think that makes sense. Thoughts @fortmarek ?

I'm aligned with that as well 👍

pepicrft:
In that case let's propagate your changes as suggested by @kwridan and we can then merge the PR. Thanks a lot @alexanderwe for this one 🚀

fortmarek (approved):
Thanks! Looks good 🙂

fortmarek:
> While doing that I also saw that the BuildSettingProviderTests was also missing tests for watchOS. I have added these tests as well now - if you think these should rather be in another PR to keep this one more atomic just let me know.

These minor fly-by changes are fine – thanks for improving that!

alexanderwe:
@fortmarek Thanks for merging !

PR #903, opened 2025-02-17
refactor!: strongly typed values in `BuildSettings` and `BuildFileSettings`

waltflanagan:
### Short description 📝
Mostly work that was done on the `9.0.0` branch that shouldn't be coupled to Sendability changes.

This introduces the `BuildSetting` type that will allow safer manipulation of build settings and set us with more typesafety when working with build settings that have constraints.

Keys issues are that this introduces source breaking changes with the signature of the `BuildSettings` typealias.


kwridan (approved):
Thanks @waltflanagan this is a much needed change! 👌

Indeed this is source breaking and would likely need a major release

e.g. updates needed for `xcdiff` https://github.com/bloomberg/xcdiff/pull/145

waltflanagan:
@kwridan thats what i was afraid of.  I'll see if i can find some time to put together a 9.0 checklist so we can identify anything else that we may want to break   There are some other `Any` collections but those seem like they could follow a similar approach to these callsites.


Do you think it would be worth a transitional phase that preserves `BuildSettings` as `[String: Any]` but introduces strongly typed backing stores and parsing to ease into a 9.0 later?  It could be something like `typealias TypedBuildSettings = [String: BuildSetting]` and we would have interfaces that expose the legacy `BuildSetting` type but convert to `TypedBuildSettings`.

kwridan:
I wouldn't shy away from a a major release for this per se as there are already a few compatibility APIs lying around that could do with a clean up:

e.g.

https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Scheme/XCScheme%2BTestableReference.swift#L10 

You are right there are other collections (e.g. `attributes`) that could do with a similar update, as such to incrementally tackle them deferring the breaking changes such that they are all done together for the next major release makes sense. 

Another accessor could be exposed with the strong types `typedBuildSettings: TypedBuildSettings` that could then be renamed back to `buildSettings` once the breaking changes are made, that could offer some value till then but at the slightly inconvenience of introducing new public API that is very short lived and would requires clients to migrate of off.

Sources/XcodeProj/Objects/Project/PBXProject.swift
```diff
@@ -562,16 +564,17 @@ extension PBXProject: PlistSerializable {
...
 
-        var plistAttributes: [String: Any] = attributes
+        var plistAttributes: [String: ProjectAttribute] = attributes
 
         // merge target attributes
-        var plistTargetAttributes: [String: Any] = [:]
+        var plistTargetAttributes: [String: [String: ProjectAttribute]] = [:]
         for (reference, value) in targetAttributeReferences {
-            plistTargetAttributes[reference.value] = value.mapValues { value in
-                (value as? PBXObject)?.reference.value ?? value
```
waltflanagan, line 571: This was added as a feature for `XcodeGen` to be able to easily work with these values.  This complicates support for strong types though and the migration due to the removal of this change would be to set the `.string(object.reference.value)` instead of just `object` at the callsites.

Sources/XcodeProj/Objects/Configuration/BuildSettings.swift
```diff
@@ -1,4 +1,69 @@
...
 
 /// Build settings.
-public typealias BuildSettings = [String: Any]
+public typealias BuildSettings = [String: BuildSetting]
+
+public enum BuildSetting: Sendable, Equatable {
+    case string(String)
+    case array([String])
+
+    var valueForWriting: String {
```
fortmarek, line 10: What do you think of this using the `CustomStringConvertible` instead of a completely custom property?

Sources/XcodeProj/Objects/Configuration/BuildSettings.swift
```diff
@@ -1,4 +1,69 @@
...
+
+    public var stringValue: String? {
+        if case let .string(value) = self {
+            value
+        } else {
+            nil
+        }
+    }
+
+    public var arrayValue: [String]? {
```
fortmarek, line 27: would probably also remove this in favor of consumers doing the switch themselves.

Sources/XcodeProj/Objects/Project/PBXProject.swift
```diff
@@ -602,3 +605,64 @@ extension PBXProject: PlistSerializable {
         })
     }
 }
+
+public enum ProjectAttribute: Sendable, Equatable {
```
fortmarek, line 609: nit: would also move this to its own file.

Sources/XcodeProj/Objects/Project/PBXProject.swift
```diff
@@ -602,3 +605,64 @@ extension PBXProject: PlistSerializable {
...
+        }
+    }
+
+    public var arrayValue: [String]? {
+        if case let .array(value) = self {
+            value
+        } else {
+            nil
+        }
+    }
```
fortmarek, line 628: as mentioned above, would consider not including these.

Tests/XcodeProjTests/Utils/BuildSettingsProviderTests.swift
```diff
@@ -411,15 +435,17 @@ class BuildSettingProviderTests: XCTestCase {
...
-                       rhs as NSDictionary,
-                       file: file,
-                       line: line)
-    }
+//    func XCTAssertEqual(_ lhs: BuildSettings, _ rhs: BuildSettings, file: StaticString = #file, line: UInt = #line) {
+//        XCTAssertEqual(lhs as NSDictionary,
+//                       rhs as NSDictionary,
+//                       file: file,
+//                       line: line)
+//    }
```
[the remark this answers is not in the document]
fortmarek: then let's remove it? 😁

fortmarek (commented):
Left some minor comments, but overall, I'm very much in favor of this. Getting rid of `Any` will be quite an improvement when using this library and has been long overdue. Thanks Mike!

Sources/XcodeProj/Objects/Configuration/BuildSettings.swift
waltflanagan: That should work fine 👍

Sources/XcodeProj/Objects/BuildPhase/PBXBuildFile.swift
```diff
@@ -195,3 +213,47 @@ final class PBXBuildPhaseFile: PlistSerializable, Equatable {
...
+        }
+    }
+
+    public var arrayValue: [String]? {
+        if case let .array(value) = self {
+            value
+        } else {
+            nil
+        }
+    }
```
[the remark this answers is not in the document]
waltflanagan: Mostly convenience, the usages of these `[String: Any]` collections generally are casting `as? <Type>` and this felt like a reasonable convenience that would help minimize the disruption to current codebases (`if case let` syntax is not fun to work with).

Tests/XcodeProjTests/Objects/Configuration/BuildFileSettingTests.swift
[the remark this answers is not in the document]
waltflanagan: Sounds good, another thing to learn about.

Sources/XcodeProj/Objects/BuildPhase/PBXBuildFile.swift
```diff
@@ -31,7 +31,25 @@ public final class PBXBuildFile: PBXObject {
...
+
+    /// Potentially present for `PBXSourcesBuildPhase` : https://buck.build/javadoc/com/facebook/buck/apple/xcode/xcodeproj/PBXBuildFile.html
+    public var compilerFlags: String? {
+        if case let .string(compilerFlags) = settings?["COMPILER_FLAGS"] {
+            compilerFlags
+        } else {
+            nil
+        }
+    }
```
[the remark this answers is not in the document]
waltflanagan: `COMPILER_FLAGS` and `ATTRIBUTES` were the only values i could find reference to in the various fixtures and sample projects here and in other both the ruby and python libraries that implement `PBXProj`.  Since it's so much of a pain to know what could possibly be in these collections i figured it would be valuable to directly expose common known keys like this.

Sources/XcodeProj/Objects/BuildPhase/PBXBuildFile.swift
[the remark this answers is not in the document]
fortmarek: not totally opposed, so we can keep these.

waltflanagan:
@kwridan i tackled the other things you mentioned and could use a re-review.  also tried to include some convenience methods to make the impact smaller from a migration perspective (avoiding the need for `if case let`)

Curious generally what else should be done to land this as a 9.0

kwridan (commented):
Thanks for the looking into the other parts @waltflanagan 🙏 

I didn't necessarily mean they had to happen in the build settings PR in one go rather as separate PRs we can incrementally do for the 9.0 release. Is the worry the release process for XcoeProj would automatically ship this PR as an minor release?

The attribute updates may need a closer look to see how to deal with references correctly to ensure their final values are resolved - my thinking is potentially adding another case `.targetReference()` which would be accounted for.

kwridan (changes requested):
marking as requesting changes per above (for the attributes + target reference issues)

thanks!

fortmarek:
> Is the worry the release process for XcoeProj would automatically ship this PR as an minor release?

That would actually happen. Given the amount of changes, I think it's fine to bundle everything here. The alternative would be to create a release branch, post individual PRs there, and then merge the release branch into `main`, so we end up with just one major release. But I'd say it's not worth the extra complexity.

waltflanagan:
@kwridan just added the `targetReference` case and a test that covers calling `fix` after adding to the project.  I think this covers the case you mentioned.  strong types made it a quick change 🎉 

One other risk of this branch is that it would block other fixes from shipping if this was merged but not quite ready to ship. 

lmk next steps on releasing a 9.0.0, i was definitely going to find some time this weekend to check on the XcodeGraph and Tuist impacts of these changes to make sure they fit those needs.  I can also see what kind of impact it would be on XcodeGen.

fortmarek:
> i was definitely going to find some time this weekend to check on the XcodeGraph and Tuist impacts of these changes to make sure they fit those needs. I can also see what kind of impact it would be on XcodeGen.

I think that having a draft PR accommodating the change in `XcodeGraph` and `Tuist` would be a good exercise before we merge this. If you would have time for that, that'd be amazing 🙏

Sources/XcodeProj/Objects/Configuration/BuildSettings.swift
```diff
@@ -1,4 +1,82 @@
...
+            nil
+        }
+    }
+
+    public var boolValue: Bool? {
+        if case let .string(value) = self {
+            switch value {
+            case yes: true
+            case no: false
+            default: nil
```
fortmarek, line 26: Xcode's behavior is actually defaulting to `false` if the value is defined but it's not `YES` or `NO`. But conceptually, this feels better, so I'll let you decide what's better here 😅

fortmarek (approved):
This is good to from my side. But as mentioned, would recommend validating these changes in `XcodeGraph` and `tuist/tuist` before merging.

Sources/XcodeProj/Objects/Project/PBXProject.swift
kwridan: The lates updates now include `.targetReference` 

```swift
pbxProject.setTargetAttributes(
    ["TestTargetID": .targetReference(target)], 
    target: testTarget 
)
```

kwridan (approved):
Thanks for the updates @waltflanagan 

@fortmarek what's the process for marking this as a major release ahead of merging?

fortmarek:
> what's the process for marking this as a major release ahead of merging?

I believe the only thing necessary is the exclamation mark already added by @waltflanagan

Sources/XcodeProj/Objects/Configuration/BuildSettings.swift
fortmarek: Makes sense!

fortmarek:
Thanks for the contribution @waltflanagan! And sorry for taking a bit longer to merge this.

waltflanagan:
No worries! glad its moving along.  I've had some other things going on as well. Thanks for merging!

PR #904, opened 2025-02-17
refactor: strongly type plist values to achieve full sendability

waltflanagan:
This is a dupe of https://github.com/tuist/XcodeProj/pull/835 targetting `main`

This should have not be source breaking and just be faster parsing.


Sources/XcodeProj/Objects/Project/PBXObject.swift
```diff
@@ -45,7 +45,12 @@ public class PBXObject: Hashable, Decodable, Equatable, AutoEquatable {
         let referenceRepository = decoder.context.objectReferenceRepository
         let objects = decoder.context.objects
         let container = try decoder.container(keyedBy: CodingKeys.self)
-        let reference: String = try container.decode(.reference)
+
+        // The reference to an object in a pbxproj is the key of a dictionary, so read the value from the codingPath
```
pepicrft, line 49: 💡Very smart!

Tests/XcodeProjTests/Tests/Fixtures.swift
```diff
@@ -1,42 +1,42 @@
 import Foundation
 import PathKit
-import XcodeProj
+@testable import XcodeProj
 
 func fixturesPath() -> Path {
     Path(#file).parent().parent().parent().parent() + "Fixtures"
 }
 
-func synchronizedRootGroupsFixture() -> (Path, [String: Any]) {
```
pepicrft, line 9: What about renaming all the functions in this fixture to end in `...Data()` instead of `...Dictionary()`?

pepicrft (approved):
Thanks for taking this one one @waltflanagan

PR #911, opened 2025-03-12
chore: auto-release only when SwiftPM dependencies are updated

pepicrft:
I noticed we were creating new releases when dependencies other than SwiftPM's got updated (e.g. Mise dependencies). We should not do this because an update of a Mise dependency has no relation with the XcodeProj package itself.
I'm mimicking the model that we adopted in [Command](https://github.com/tuist/command) where we configure Renovabebot differently to label PRs based on the package manager, and use the convention in the Git configuration to exclude some commits from the detection logic.

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -90,6 +90,22 @@ commit_preprocessors = [
   # remove issue numbers from commits
   { pattern = '\((\w+\s)?#([0-9]+)\)', replace = "" },
 ]
+commit_parsers = [
+    { message = "^feat", group = "Features" },
+    { message = "^fix", group = "Bug Fixes" },
+    { message = "^doc", group = "Documentation" },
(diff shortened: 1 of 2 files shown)


fortmarek (approved):
would love to see added across our repositories. Long-term, we might want to see if we can share bits of `cliff.toml`, but it might be an overkill given we don't change it _that_ often.

pepicrft:
> would love to see added across our repositories. Long-term, we might want to see if we can share bits of cliff.toml, but it might be an overkill given we don't change it that often.

Yeah... with things like this, I don't know what's best, if having a centralized configuration that we can reuse or live with a bit of inconsistency which we fix over time? 

I check git-cliff to see if they support reading the configuration from another repository, and that's not supported.

PR #937, opened 2025-05-16
fix: Align the sorting of project references with Xcode 16

mikhailmulyar:
### Short description 📝
If project saved with XcodeProj then opening it with Xcode 16 generate some changes. So we could respect Xcode 16 sorting by sorting our result with the same way.

### Solution 📦
Do the same sorting as Xcode 16 do

---
NB! This PR is copy of https://github.com/tuist/XcodeProj/pull/867 with proper tests added.

### DIFF
--- a/Fixtures/Xcode16ProjectReferenceOrder/Sources/AppDelegate.swift
+++ b/Fixtures/Xcode16ProjectReferenceOrder/Sources/AppDelegate.swift
@@ -1,0 +1,16 @@
+import Framework1
+import Framework2
+import UIKit
+
+@main
+class AppDelegate: UIResponder, UIApplicationDelegate {
+    var window: UIWindow?
+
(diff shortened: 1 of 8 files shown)


pepicrft:
@mikhailmulyar you'll have to run `mise run lint-fix` to fix the linting issues. After that, we should be good to go once CI is green.

pepicrft:
@all-contributors add @mikhailmulyar for code

mikhailmulyar:
> @mikhailmulyar you'll have to run `mise run lint-fix` to fix the linting issues. After that, we should be good to go once CI is green.

Added fix

pepicrft:
Thanks @mikhailmulyar. That one yields false positives sometimes, so we can ignore and merge. I'll into why we are getting those later.

PR #980, opened 2025-10-06
feat: speed up object parsing

karlpuusepp:
`PBXObject.isa` can be quite expensive in a hot loop, as it internally calls `Mirror` to describe the type. Parsing large projects can be considerably faster by inlining the type names as raw strings.

As an example I measured our own project decoding time go from ~700ms to ~550ms with this change.


fortmarek (approved):
This is great, thanks!

PR #1016, opened 2025-11-25
feat: Compatibility with the iOS platform

### LINKED ISSUE #1015: Compilation errors for iOS platform
Currently, XcodeProj is not compilable on iOS platform. This is an issue when attempting to compile products that have XcodeProj as a (transitive) dependency (ex: CuckooGenerator from https://github.com/Brightify/Cuckoo)

Image

nmcc24:
Resolves https://github.com/tuist/XcodeProj/issues/1015

- [x] Add compiler flags
- [x] Add iOS version checks
- [x] Build XcodeProj target
- [x] Compile XcodeProj tests

### DIFF
--- a/.github/workflows/xcodeproj.yml
+++ b/.github/workflows/xcodeproj.yml
@@ -33,6 +33,14 @@ jobs:
       - run: mise use swift@6.0.2
       - name: Build
         run: swift build --configuration release
+  build-ios:
+    name: Build (iOS)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Build for iOS Simulator
+        run: xcodebuild build -scheme XcodeProj -destination 'platform=iOS Simulator,name=iPhone 16'
+  
   test:
     name: Test (macOS / Xcode)
     runs-on: macos-latest
@@ -41,7 +49,6 @@ jobs:
       - uses: jdx/mise-action@v2
       - name: Run tests
         run: mise run test
-
   test-linux:
     name: Test (Linux)
     runs-on: ubuntu-latest
@@ -55,6 +62,13 @@ jobs:
           git config --global init.defaultBranch main
       - name: Test
         run: swift test
+  test-ios:
+    name: Test (iOS / Xcode)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run tests on iOS Simulator
+        run: xcodebuild test -scheme XcodeProj -destination 'platform=iOS Simulator,name=iPhone 16'
 
   lint:
     name: Lint
--- a/.github/workflows/xcodeproj.yml
+++ b/.github/workflows/xcodeproj.yml
@@ -39,8 +39,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Build for iOS Simulator
-        run: xcodebuild build -scheme XcodeProj -destination 'platform=iOS Simulator,name=iPhone 16'
-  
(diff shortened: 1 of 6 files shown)


pepicrft:
@nmcc24 thanks a lot for the context, I appreciate it.

Wouldn't [`TargetDependencyCondition`](https://docs.swift.org/swiftpm/documentation/packagedescription/targetdependencycondition) be useful in this case, where you can declare a dependency conditionally based on the platform:

```swift
Target(name: "MyTarget", dependencies: [
  .product(name: "XcodeProj, condition: .when([.macOS]))
])
```

nmcc24:
> > Are there any particular areas you'd like me to focus on? For instance, [XCTestCase+Shell.swift‎](https://github.com/tuist/XcodeProj/pull/1016/changes#diff-dae17da4d7d7305d579bcdd3607c9645bb3a46ad5f10fc3c78f637ef97698962) won't be possible to compile on iOS due to the usage of Process unavailable on Mac. I assume there are other cases like that and you may know them already?
> 
> I'd suggest to:
> 
> * Identify which areas can't run on iOS, so that we can document them
> * Make sure our CI ensures the project compiles and its tests run on iOS
> 
> With those two things, I think we should be good to give this one a merge.

XcodeProj is primarily designed for macOS. While the library compiles on iOS, the following areas have limited or no functionality on that platform.

### Source Code

| Area | File | Details |
|------|------|---------|
| **`glob()` file discovery** | `Sources/XcodeProj/Extensions/Path+Extras.swift` | Uses C `glob()`, `glob_t`, `GLOB_BRACE`, `GLOB_TILDE`, and `GLOB_MARK` — none of which are available in the iOS SDK. On iOS, a `FileManager`-based fallback is used instead (no brace expansion support). |
| **`String.md5` availability** | `Sources/XcodeProj/Extensions/String+md5.swift` | CryptoKit availability check specifies `macOS 10.15, iOS 13.0`. On older iOS versions, a slower software MD5 implementation is used. |

The `glob()` function is called from the following entry points, meaning **parsing `.xcodeproj` bundles from disk** is affected:

- `XcodeProj.swift` — discovers `*.pbxproj`, `*.xcworkspace`, and `*.xcuserdatad`
- `XCUserData.swift` — discovers `*.xcscheme`
- `XCSharedData.swift` — discovers `*.xcscheme`
- `XCWorkspace.swift` — discovers `*.xcworkspacedata`

### Tests

| Area | File(s) | Details |
|------|---------|---------|
| **`Process` (subprocess execution)** | `Tests/XcodeProjTests/Extensions/XCTestCase+Shell.swift` | The `Process` class is unavailable on iOS. `checkedOutput()` returns `nil` on iOS. |
| **Integration tests using `git`** | `Tests/XcodeProjTests/Tests/testWrite.swift`, `Tests/XcodeProjTests/Scheme/XCSchemeManagementTests.swift` | These tests invoke `git` via `checkedOutput()` and are skipped on iOS via `XCTSkip`. |
| **Tests gated to macOS/Linux** | `XcodeProjIntegrationTests.swift`, `PBXGroupTests+Path.swift`, `PBXProjIntegrationTests.swift`, `PBXOutputSettingsTests.swift`, `PBXProjEncoderTests.swift` | Already wrapped in `#if os(macOS) \|\| (os(Linux) && compiler(>=6.1))` — excluded from iOS builds entirely. |

nmcc24:
@pepicrft areas documented (https://github.com/tuist/XcodeProj/pull/1016#issuecomment-4260509797) and workflow updated as per your requested. Please let me know your thoughts. Thanks

nmcc24:
@pepicrft would you be able to review this? Thank you

pepicrft:
Thanks @nmcc24

Issue #921, opened 2025-04-01
`error: (DecodingError) typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: …, CodingKeys(stringValue: "shellScript", …)], debugDescription: "Expected to decode String but found an array instead.", …))`


F1248:
## Context 🕵️‍♀️

- Xcode 16.3 (16E140)
- Xcode project format: Xcode 16.3 (`objectVersion = 90`)
- [Periphery](https://github.com/peripheryapp/periphery) 3.0.3
- XcodeProj 8.27.3

project.pbxproj:

```
// !$*UTF8*$!
{
	archiveVersion = 1;
	classes = {
	};
	objectVersion = 90;
	objects = {

…

/* Begin PBXShellScriptBuildPhase section */
		A42660D00000000000000000 /* Test Shell Script */ = {
			isa = PBXShellScriptBuildPhase;
			name = "Test Shell Script";
			shellPath = /bin/zsh;
			shellScript = (
				"echo 'Hello World!'",
				"",
			);
		};
/* End PBXShellScriptBuildPhase section */

…
	};
	rootObject = A42B35300000000000000000 /* Project object */;
}

```

## What 🌱

When I try to scan my project with Periphery (which uses XcodeProj) with `periphery scan`, I get this error message:

`error: (DecodingError) typeMismatch(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "objects", intValue: nil), _DictionaryCodingKey(stringValue: "A42660D12C729D04000E0507", intValue: nil), CodingKeys(stringValue: "shellScript", intValue: nil)], debugDescription: "Expected to decode String but found an array instead.", underlyingError: nil))`

It seems that at least since Xcode project format 16.3 (`objectVersion = 90`), shell scripts (`Targets` → \<target name\> → `Build Phases` → `New Run Script Phase`) are stored in arrays where each element represents a line, whereas before they were stored as a single string.

fceruti:
@dosu Can you flesh out a minimal solution required to solve this that follows this repo's code style, and as you said, handles both new and old format?

fourbytes:
Same issue here on the shellScript key

PR #1018, opened 2025-11-26
fix: support for shellScript in the Xcode 16 format

michaelversus:
Resolves https://github.com/tuist/XcodeProj/issues/921

### Short description 📝
This MR fixes an issue with projects that use the new format, where for build phases, a shell script is an array of strings instead of a string.

### Solution 📦
I am deserializing the script with a fallback approach.
If we fail deserializing using the `String` type we are trying the `[String]` type to ensure that I will not break old xcode version format

### Implementation 👩‍💻👨‍💻

- [x] Step 1 (Introduce extra decoding handling for `shellScript` property of `PBXShellScriptBuildPhase`)
- [x] Step 3 (Removed unused testDictionary function from `PBXShellScriptBuildPhaseTests`)
- [x] Step 3 (Introduce new fixture and update unit tests)


pepicrft:
Thanks a lot @michaelversus 🚀. Release in progress...

PR #1037, opened 2025-12-26
feat: support new Xcode 16 XCBuildConfiguration format

### LINKED ISSUE #852: Xcode 16 • XCBuildConfiguration Changes
## Context 🕵️‍♀️
Related to https://github.com/tuist/XcodeProj/pull/827, when a build configuration uses an xcconfig in the new folders the `baseConfigurationReference` is removed and replaced with `baseConfigurationReferenceAnchor` (`PBXFileSystemSynchronizedRootGroup`) and `baseConfigurationReferenceRelativePath`.

## What 🌱
Xcode 16 introduces changes when you use the new synchronized folders.

## Proposal 🎉
Check for these new values before falling back to the old value for deriving information from the xcconfigs.

johnrbent:
Resolves https://github.com/tuist/xcodeproj/issues/852

### Short description 📝
Adds support for changes to the `XCBuildConfiguration` representation of xcconfig files contained within a file-system synchronized root group.

### Solution 📦
Instead of a single `baseConfigurationReference`, these xcconfig files are represented by a pair of:
  - `baseConfigurationReferenceAnchor`: reference to the synced root group.
  - `baseConfigurationReferenceRelativePath`: relative path to the file from the root group.

*config files within a PBXGroup are still represented by the `baseConfigurationReference`

### Implementation 👩‍💻👨‍💻
- [x] Add the two new optional properties to the `XCBuildConfiguration` type following existing patterns for `baseConfigurationReference`
- [x] Include new unit test on the encoding as well as a fixture project with xcconfig files in a synced root group.
- [ ] I have another PR ready for XcodeGraph to adopt these changes that I'll put up once this is merged. 

(This is my first contribution to the project. I'm happy to make changes if needed)


.editorconfig
```diff
@@ -2,7 +2,7 @@ root = true
 
 [*]
 indent_style = space
-indent_size = 2
+indent_size = 4
```
johnrbent: I wasn't sure about this, but I had to change this to get the correct indentation locally in Xcode. Maybe someone can explain why it should be 2?

mokagio:
Thank you @johnrbent for working on this. It's unfortunate it's been sitting here all this time, but such is the fate with many open source projects. Maintainers are busy. One has to be thankful for the tools to exist in the first place and for the open source community to be so collaborative.

I cherry picked your changes in [a personal fork](https://github.com/mokagio/XcodeProj-Tuist/commits/automattic/fork-patches/) because I run in an issue that your work addresses. Thanks again.

fortmarek:
Hey @johnrbent, thanks a lot for the contribution! I pushed a follow-up commit (145093a8) to address a few items from my review. Happy to revert any of these if you'd prefer to handle them yourself. Summary of what changed and why:

### `.editorconfig`
Reverted `indent_size` from 4 back to 2. The change looked unrelated to the feature and would have shifted indentation behavior repo-wide for every contributor.

### Split `XCBuildConfiguration` init into two dedicated initializers
Replaced the single 4-arg init with:

- `init(name:baseConfiguration:buildSettings:)` — the pre-PR signature, source-compatible for existing callers using xcconfigs in a `PBXGroup`.
- `init(name:baseConfigurationAnchor:baseConfigurationRelativePath:buildSettings:)` — the Xcode 16+ form, where the anchor and relative path are non-optional since they only make sense as a pair.

Each init carries an `- Important:` doc block calling out that the two reference styles are mutually exclusive and pointing readers at the other initializer. This is the documentation-only path you suggested in the original implementation, but expressed at the API level so the type system nudges callers toward the right shape.

### Decoder: reject partial anchor data
Previously, if a pbxproj had `baseConfigurationReferenceAnchor` without `baseConfigurationReferenceRelativePath` (or vice-versa), the decoder would silently fall through to the legacy `baseConfigurationReference` path and drop the partial field. The decoder now throws `DecodingError.dataCorrupted` against the missing key, so corrupt input surfaces immediately rather than silently losing data. Xcode always emits both keys together, so this only fires on genuinely malformed input.

### Doc typos
`"may be reference"` → `"may be referenced"`, and a missing space before a `` ``PBXGroup`` `` link.

### Tests
Added two failure tests covering the new partial-anchor-data decoder paths (anchor-without-relative-path and relative-path-without-anchor), each asserting `DecodingError.dataCorrupted` with the expected coding key. Skipped a success-case decode test because `super.init(from:)` requires a parent coding path (existing limitation of the JSON-decoder test pattern in this file) — the encoder fixture test you added already exercises the full decode→encode round-trip end-to-end.

### Things I deliberately left alone
The `baseConfigurationAnchor` setter still ignores `nil` assignments. This matches the existing `baseConfiguration` setter pattern, and changing it felt out of scope for this PR.

All 355 tests pass locally. Let me know what you think!

fortmarek:
Quick follow-up — pushed bb288525 to drop the partial-anchor-data throwing behaviour and the two tests that asserted it. Decoder is back to the original silent fall-through, so an anchor without its relative-path partner (or vice-versa) is just skipped, same as before.

fortmarek (approved):

fortmarek:
Thanks for the PR! Sorry that this has been left unattended for this long. LGTM once the CI is green.

PR #1044, opened 2026-01-29
feat: add delete support for XCLocalSwiftPackageReference

Kosikowski:
# Add delete support for XCLocalSwiftPackageReference

The `delete` method in `PBXObjects` was missing support for `XCLocalSwiftPackageReference`, causing local package references to remain as orphaned objects in the pbxproj when switching from local back to remote packages.

This adds handling for `localSwiftPackageReferences` in the delete method, matching the existing pattern for `remoteSwiftPackageReferences`.

### Short description 📝

When programmatically managing Swift package dependencies in Xcode projects, switching a local package reference back to a remote package reference leaves orphaned `XCLocalSwiftPackageReference` objects in the project file. This occurs because `PBXObjects.delete(reference:)` handles `remoteSwiftPackageReferences` but not `localSwiftPackageReferences`.

This PR adds the missing delete support for local Swift package references, ensuring clean removal of these objects when they are no longer needed.

### Solution 📦

The solution follows the existing pattern established for other object types in the `delete(reference:)` method. Specifically, it mirrors the implementation for `remoteSwiftPackageReferences`:

```swift
} else if let index = localSwiftPackageReferences.index(forKey: reference) {
    return _localSwiftPackageReferences.remove(at: index).value
}
```

**Alternative approaches considered:**
none

### Implementation 👩‍💻👨‍💻

- [x] Added `localSwiftPackageReferences` handling to `PBXObjects.delete(reference:)` method
- [x] Added unit tests for add/delete operations on `XCLocalSwiftPackageReference`:
  - `test_add_addsObjectToPBXProj` — verifies adding works correctly
  - `test_delete_removesObjectFromPBXProj` — verifies deletion works correctly
  - `test_delete_removesCorrectObject_whenMultipleExist` — verifies only the targeted object is removed
  - `test_delete_doesNothing_whenObjectNotInProj` — verifies safe behaviour for non-existent objects
- [x] Verified all existing tests continue to pass

### Testing 🧪

All tests pass locally:

```
Test Suite 'XCLocalSwiftPackageReferenceTests' passed at 2026-01-29.
    Executed 8 tests, with 0 failures (0 unexpected) in 0.003 seconds
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
     timeout-minutes: 15
     if: "!startsWith(github.event.head_commit.message, '[Release]')"
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 0
       - uses: jdx/mise-action@v3
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,5 +1,5 @@
 [tools]
 swiftformat = "0.54.3"
-tuist = "4.119.1"
-swiftlint = "0.55.1"
+tuist = "4.150.0"
+swiftlint = "0.63.2"
 "git-cliff" = "2.4.0"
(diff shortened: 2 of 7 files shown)


Kosikowski:
No problem at all. All done.

PR #1114, opened 2026-04-19
feat: add `traits` to `XCRemoteSwiftPackageReference` and `XCLocalSwiftPackageReference`

### LINKED ISSUE #1113: Preserve Swift Package trait selections across read/write
## Summary

Xcode 26.4+ writes Swift Package Manager traits ([SE-0450](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md)) into `project.pbxproj` as `traits = ( ... );` entries on `XCRemoteSwiftPackageReference` and `XCLocalSwiftPackageReference`. Stock `XcodeProj` 9.11.0 has no `traits` field on these types, so any tool that loads such a project and calls `XcodeProj.write()` silently discards the user's trait selection. On next open, Xcode sees no `traits` key and falls back to default traits re-enabling features the user opted out of, or pulling back frameworks they excluded (e.g. a `NoUIFramework` selection that kept UIKit/AppKit out would quietly come back on).

## pbxproj format in the wild
Real-world examples:
  - [`getsentry/sentry-cocoa` macOS-CLI-Xcode sample](https://github.com/getsentry/sentry-cocoa/pull/7578): `XCLocalSwiftPackageReference` with `traits = ( NoUIFramework, );`
  - [`ml-explore/mlx-swift-examples`](https://github.com/ml-explore/mlx-swift-examples/blob/main/mlx-swift-examples.xcodeproj/project.pbxproj): `XCRemoteSwiftPackageReference` with `traits = ( );` (empty = explicit default-trait opt-out, semantically distinct from an absent key)

Confirmed introduced in Xcode 26.4 (stable, currently 26.4.1); not emitted by 26.3 or earlier.

## Reproducer

https://github.com/chigichan24/XcodeProj-MinimalProject

```bash
git clone https://github.com/chigichan24/XcodeProj-MinimalProject
cd XcodeProj-MinimalProject
./compare.sh
```

The `Stock/` side (pinned to XcodeProj exact 9.11.0) prints:

```
fixture has 5 trait-related lines ...
after load -> write: 0 trait-related lines
Result: traits preserved in output pbxproj = NO
```

The `Branch/` side (patch applied) preserves them and passes every round-trip check.

## Impact

- Downstream generators built on XcodeProj (notably yonaskolb/XcodeGen#1585) cannot emit traits while this field is absent from the public model. Users of such generators today are either stuck without traits or hand-committing the generated `.xcodeproj` to preserve them
(as the Sentry team does).

## Proposed direction

Add `traits: [String]?` to `XCRemoteSwiftPackageReference` and `XCLocalSwiftPackageReference`. Keep `nil` vs `[]` distinct (SE-0450 treats `traits: []` as "disable default traits"). No public-API change for projects not using traits.

chigichan24:
## What changed

This pull request adds optional `traits: [String]?` support to both `XCRemoteSwiftPackageReference` and `XCLocalSwiftPackageReference`.

XcodeProj now:

- Decodes package trait selections from Xcode project files.
- Writes populated and explicitly empty trait arrays without collapsing their meaning.
- Includes traits in package-reference equality.
- Preserves trait selections through a project read and write cycle.
- Correctly compares local package references by relative path.
- Keeps the optimized string escaping path on supported systems and uses a compatible fallback on older Apple operating system versions.

## Why

Xcode 26.4 writes [Swift Package Manager package traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md) on package-reference objects:

```
traits = (
    NoUIFramework,
);
```

Without model support, reading and writing one of these projects silently removes the trait selection. Xcode then falls back to the package defaults the next time the project opens.

The branch also needed to remain buildable for the package's existing iOS 12 deployment target after being updated to current `main`.

## Root cause

The package-reference models did not decode, store, serialize, or compare the `traits` field introduced by Xcode.

While adding equality coverage, the implementation also exposed an older generated-equality gap for `XCLocalSwiftPackageReference`. Its relative path was not part of equality, so local references with different paths could compare as equal.

Separately, current `main` uses a string initializer introduced in iOS 14 from code compiled for iOS 12. This only surfaced after the pull request workflows moved to the latest Xcode runner.

## Approach

Traits use `[String]?` to preserve the project file exactly:

- `nil` means no explicit selection, so package defaults apply.
- `[]` means default traits are explicitly disabled.
- A populated array contains the explicitly enabled traits.

An array is used instead of a set because Xcode project arrays preserve source-file order.

The selection remains on the local or remote package reference. Product dependency objects are unchanged because Xcode does not write traits there.

A dedicated fixture covers a populated local selection and an explicitly empty remote selection. The integration test verifies that reading and writing the fixture produces no diff.

String escaping keeps the direct-storage implementation where that initializer is available. Older supported systems use the same escaping routine through a byte buffer.

## Impact

Projects using package traits can now be safely read, modified, and written with XcodeProj without losing their trait configuration. Existing callers remain source compatible because the new initializer arguments default to `nil`.

The package also continues to compile for its existing iOS 12 deployment target.

## Validation

- `mise run lint`
- `mise run test`: 381 XCTest tests and 6 Swift Testing tests passed.
- `xcodebuild build -scheme XcodeProj -destination 'generic/platform=iOS Simulator'`
- `xcodebuild test -scheme XcodeProj` on an iPhone 17 Pro simulator running iOS 26.5: 323 XCTest tests and 6 Swift Testing tests passed, with 5 expected skips.
- The package-traits fixture read and write test passed without a diff.


Fixtures/iOS/ProjectWithSwiftPackageTraits.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,601 @@
...
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/ReactiveX/RxSwift";
+			requirement = {
+				kind = upToNextMajorVersion;
+				minimumVersion = 5.0.1;
+			};
+			traits = (
+			);
+		};
+/* End XCRemoteSwiftPackageReference section */
```
chigichan24, line 569: note:  The reason of PR diff size is XL.

`Fixtures/iOS/ProjectWithSwiftPackageTraits.xcodeproj/project.pbxproj` is ~600 lines, which makes this PR look larger than it is. It's a copy of the existing `ProjectWithXCLocalSwiftPackageReference.xcodeproj` fixture with `traits = ( ... );` entries added to the Remote and Local reference sections

The substantive delta from the source fixture is few lines (`diff` output):

```diff
< buildConfigurationList = ... "ProjectWithXCLocalSwiftPackageReference" ...;
> buildConfigurationList = ... "ProjectWithSwiftPackageTraits" ...;
< ... "ProjectWithXCLocalSwiftPackageReference" = { ...
> ... "ProjectWithSwiftPackageTraits" = { ...
+   traits = (
+       NoUIFramework,
+   );
+   traits = (
+   );
```

  A new fixture is used rather than modifying the existing one so that `testReadWriteProducesNoDiff` on `ProjectWithXCLocalSwiftPackageReference.xcodeproj` stays stable — adding `traits = ( ... );` to that existing fixture would have changed what its existing assertion produces and entangled this change with unrelated local-package add/delete tests.

pepicrft (approved):
Validated against current `main`. The implementation preserves absent, empty, and populated package trait selections, includes traits in equality, and keeps project read and write operations stable. The full test suite and repository lint command pass.

pepicrft (approved):
Revalidated at the current head. The package traits model and project read and write coverage are sound. The availability fallback preserves iOS 12 build compatibility while retaining the optimized path on newer systems. All repository workflows pass.

PR #1162, opened 2026-07-30
fix: recognize Icon Composer .icon file type

jsj:
### Short description 📝

This change recognizes Xcode 26 Icon Composer `.icon` bundles during file-reference generation.

### Solution 📦

Xcode maps the `icon` extension to `folder.iconcomposer.icon`.

XcodeProj did not contain this mapping. Thus, Tuist-generated projects created `PBXFileReference` entries without a `lastKnownFileType` value for `AppIcon.icon`.

Local Xcode inferred the file type. Xcode Cloud did not infer the file type. It treated the bundle as a plain folder.

As a result, `actool` received only `Assets.xcassets`. It did not find `AppIcon`.

This change adds the missing mapping to `Xcode.filetype(extension:)`.

### Implementation 👩‍💻👨‍💻

- This change maps `.icon` to `folder.iconcomposer.icon`.
- A regression test covers this mapping.
- `swift test --filter XcodeTests` passes.
- The affected project archives and exports an IPA locally with this mapping.
- Xcode Cloud now includes `AppIcon.icon` in the asset-compiler inputs.
- The project team handles a separate resource-ownership error in that repository.

### DIFF
--- a/Sources/XcodeProj/Project/Xcode.swift
+++ b/Sources/XcodeProj/Project/Xcode.swift
@@ -175,6 +175,7 @@ public enum Xcode {
         "i": "sourcecode.c.c.preprocessed",
         "icns": "image.icns",
         "ico": "image.ico",
+        "icon": "folder.iconcomposer.icon",
         "iconset": "folder.iconset",
         "ii": "sourcecode.cpp.cpp.preprocessed",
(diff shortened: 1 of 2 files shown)


pepicrft:
Thanks @jsj

PR #1179, opened 2026-09-17
fix: resolve base-localized resource paths in JSON projects

ileitch:
Opening a JSON project containing a localized storyboard or XIB can make PBXVariantGroup.fullPath(sourceRoot:) return its containing directory instead of the Base resource. JSON file references omit the language name, while the existing resolver only recognizes a child named Base. Consumers filtering the returned path by resource extension silently skip the file.

For unnamed children, infer the Base variant from an exact Base.lproj directory component. Explicit Base names retain priority, and groups without a Base variant keep their existing directory fallback. The change is confined to path resolution; decoding and encoding are unchanged.

Add an on-disk JSON fixture covering a storyboard whose Base variant is not the first child, a nested XIB, and a group without Base. Tests cover direct loading, writes and reopenings in both formats, preservation of explicit Base names, and conversion of the existing iOS PBX fixture through JSON.

Validation:
- The new fixture tests reproduced five assertion failures before the fix.
- Full macOS suite: 383 XCTest tests and 87 Swift Testing tests passed.
- mise run lint passed (existing warnings remain).
- A separate review used a standalone public-API harness to check ten edge cases, including misleading path components, missing paths, explicit-name precedence, and unchanged child names. No additional issues found.
- The original Periphery JSON/legacy fixture reproduction now resolves the same existing LaunchScreen.storyboard path in both formats.
- Full Linux suite with Swift 6.1.2 (matching upstream CI): 383 XCTest tests and 87 Swift Testing tests passed.


ileitch:
@pepicrft @freddi-kit

freddi-kit (approved):
Thank you for finding this case!

